PR #140, opened 2014-12-18
Added #addin directive for NuGet addins

devlead:
Makes it possible to load and reference addins via NuGet

Sample script
```csharp
#addin "Cake.AliaSql"

AliaSql("Rebuild", new AliaSqlSettings{ToolPath=""});
```

This will give an output  similar to (It outputs an error as I'm sending in bogus settings, but it's from the AddIn code)
```
Installing 'Cake.AliaSql 0.4.1'.
Successfully installed 'Cake.AliaSql 0.4.1'.
Addin: Cake.AliaSql, adding Reference C:/temp/dev/github/cake/tools/Addins/Cake.AliaSql/Cake.AliaSql.dll
Error: Path cannot be empty.
Parameter name: path
```

What it does is basically
  1. Check if AddIns folder exists if not it creates it
  2. Then if checks if the choosen Addin exists and if it contains any assemblies, if not it will try to find nuget.exe and install the package
  3. It will reference any found assemblies

### DIFF
--- a/src/Cake.Core/Scripting/ScriptProcessor.cs
+++ b/src/Cake.Core/Scripting/ScriptProcessor.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using Cake.Core.Diagnostics;
 using Cake.Core.IO;
+using Cake.Core.IO.NuGet;
 using Cake.Core.Scripting.Processors;
 
 namespace Cake.Core.Scripting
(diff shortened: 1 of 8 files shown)


src/Cake.Core/Scripting/ScriptProcessor.cs
```diff
@@ -35,14 +36,31 @@ public ScriptProcessor(IFileSystem fileSystem, ICakeEnvironment environment, ICa
...
+            if (toolResolvers == null)
+            {
+                throw new ArgumentNullException("toolResolvers");
+            }
             _fileSystem = fileSystem;
             _environment = environment;
             _log = log;
 
-            _lineProcessors = new LineProcessor[] {
+            _lineProcessors = new LineProcessor[]
```
patriksvensson: Changed my mind. Those should not be injected.

devlead:
@patriksvensson i asjusted so `NugetToolResolver` now gets injected via `INuGetToolResolver` interface.

PR #152, opened 2015-01-12
Added predictable encoding to AssemblyInfoCreator

devlead:
Currently uses `System.Text.Encoding.Default` which isn't predictable and often doesn't support non ascii characters.


devlead:
@patriksvensson  also now added a small fix `CLSCompliant` attribute had wrong namespace defined, which made projects not compile if `AssemblyInfoSettings.CLSCompliant` was set.

patriksvensson:
I'm having my lunch break now, so I'll merge it as soon as the CI build pass :+1:

PR #154, opened 2015-01-14
Add support for transforming nuspec files

devlead:
Added support for specifying the nuspec files from cake via NuGetPackSettings like below
```csharp
NuGetPack("App.nuspec", new NuGetPackSettings
{
    Files = new[]
    {
        new NuSpecContent {Src = "App.Core.dll", Target = "lib/net45"},
        new NuSpecContent {Src = "App.Core.xml", Target = "lib/net45"},
        new NuSpecContent {Src = "App.Core.pdb", Target = "lib/net45"},
        new NuSpecContent {Src = "LICENSE"}
        new NuSpecContent {Src = "App.Common.*", Target = "lib/net45", Exclude = "*.pdb"},
    }
});
```

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NuGet/Pack/NuGetPackerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NuGet/Pack/NuGetPackerTests.cs
@@ -369,5 +369,83 @@ namespace Cake.Common.Tests.Unit.Tools.NuGet.Pack
                     fixture.FileSystem.GetTextContent("/Working/existing.temp.nuspec").NormalizeLineEndings());
             }
         }
+
+            [Fact]
+            public void Should_Replace_Template_Tokens_In_Nuspec_With_Files()
+            {
+                // Given
+                var fixture = new NuGetFixture();
+                var packer = fixture.CreatePacker();
+
+                // When
+                packer.Pack("./existing.nuspec", new NuGetPackSettings
+                {
+                    Id = "The ID",
+                    Version = "The version",
+                    Title = "The title",
+                    Authors = new[] { "Author #1", "Author #2" },
+                    Owners = new[] { "Owner #1", "Owner #2" },
+                    Description = "The description",
+                    Summary = "The summary",
+                    LicenseUrl = new Uri("https://license.com"),
+                    ProjectUrl = new Uri("https://project.com"),
+                    IconUrl = new Uri("https://icon.com"),
+                    RequireLicenseAcceptance = true,
+                    Copyright = "The copyright",
+                    ReleaseNotes = new[] { "Line #1", "Line #2", "Line #3" },
+                    Tags = new[] { "Tag1", "Tag2", "Tag3" },
+                    Files = new[]
+                    {
+                        new NuSpecContent {Source = "Cake.Core.dll", Target = "lib/net45"},
+                        new NuSpecContent {Source = "Cake.Core.xml", Target = "lib/net45"},
+                        new NuSpecContent {Source = "Cake.Core.pdb", Target = "lib/net45"},
+                        new NuSpecContent {Source = "LICENSE"}
+                    }
+                });
+
+                // Then
+                Assert.Equal(
+                    Resources.Nuspec_Metadata.NormalizeLineEndings(),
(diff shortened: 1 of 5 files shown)


patriksvensson:
This should resolve issue #70 :)
Or am I misinterpreting this PR?

src/Cake.Common/Tools/NuGet/Pack/NuSpecTransformer.cs
```diff
@@ -61,29 +95,40 @@ private static XmlNode FindOrCreateElement(XmlDocument document, XmlNamespaceMan
...
+                string.IsNullOrWhiteSpace(parent.NamespaceURI)
+                    ? document.CreateElement(name)
+                    : document.CreateElement(name, parent.NamespaceURI)
+                );
+        }
+
+        private static void AddAttributeIfSpecified(this XmlNode element, string value, string name)
+        {
+            if (string.IsNullOrWhiteSpace(value) || element.OwnerDocument==null || element.Attributes==null)
+                return;
```
patriksvensson: Curly braces... :smile:

src/Cake.Common/Tools/NuGet/Pack/NuSpecContent.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Represents a NuGet nuspec file
+    /// </summary>
+    public class NuSpecContent
+    {
+        /// <summary>
+        /// The location of the file or files to include. The path is relative to the NuSpec file unless an absolute path is specified. The wildcard character, *, is allowed. Using a double wildcard, **, implies a recursive directory search.
+        /// </summary>
+        public string Src { get; set; }
```
patriksvensson: Should we call this one `Source` instead of `Src`, or should it map 1:1 with the nuspec name?

src/Cake.Common.Tests/Unit/Tools/NuGet/Pack/NuGetPackerTests.cs
```diff
@@ -328,7 +328,14 @@ public void Should_Replace_Template_Tokens_In_Nuspec()
                     RequireLicenseAcceptance = true,
                     Copyright = "The copyright",
                     ReleaseNotes = new[] { "Line #1", "Line #2", "Line #3" },
-                    Tags = new[] { "Tag1", "Tag2", "Tag3" }
```
patriksvensson: Would prefer two extra tests that validates the new `Files` functionality. One with and one without XML namespace (this was a problem that was encountered).

Maybe also test with some other files to make sure that the nuspec transformation isn't being done by the old transformator, where the values were hard coded into the nuspec (functionality we want to preserve).

devlead:
Yes this should indeed solve #70 ,busy now will take a look code / tests.

patriksvensson:
No rush as usual :smile:

src/Cake.Common/Tools/NuGet/Pack/NuSpecTransformer.cs
gep13: If this is something that you want enforced, is there a reason you aren't running StyleCop as part of the build process?  Just curious :smile_cat:

patriksvensson: I want it enforced but I've been to lazy to add StyleCop to be honest :smile:

patriksvensson: Actually, taking the time as I write this to finally add StyleCop. Thanks for the push, I needed it :wink:

gep13: Woo hoo!  Good choice!  Will be interesting to see how you integrate it.  I have it fully integrated for ChocolateyGUI, and it reports StyleCop errors into the AppVeyor Test section and then fails the build:

https://github.com/chocolatey/ChocolateyGUI/blob/develop/BuildScripts/default.ps1#L68

devlead:
Reverted old tests added 2 new test with files with & without namespace
Refactored according to PR feedback

Issue #149, opened 2015-01-07
GitHub gists


patriksvensson:
Would be useful to be able to reference a gist from the cake script.
The source of the gist would be inserted into the build script. Good for sharing utility methods. 

``` csharp
#gist "https://gist.github.com/patriksvensson/872bfcc44b234b12ea8c"
```

RichiCoder1:
Would the idea here be to enable streaming a file into the build script, or specifically allow the loading of full on gists?

patriksvensson:
I'm thinking that it should work the same way as the `#l`-directive, but instead of streaming from disc, stream over the interwebz.

devlead:
Gists after just GIT repositories so perhaps do an generic GIT repo script include?

Something like
`#git https://gist.github.com/872bfcc44b234b12ea8c.git?script=RelativeGlobberPathToFileOrFiles`
Then clone into folder relative to cake (like addins)

Then it could be cached if many scripts were to use the same repo, and just pull if exist to get latest version.

Built in GIT support for clone, pull,  push, add & commit would be really useful in general,  especially as many projects are published via GIT.

patriksvensson:
Hmmm, maybe. Problem is that we probably would need an additional dependency on `libgit2sharp` since it's a pain to locate the git-executable. When I think about it, maybe it's better to do this as an addin when the `#nuget`-support is in place?

devlead:
Meanwhile could just add HTTP support to the `#l` directive, both GITHub repos and gists support raw downloading if files.

patriksvensson:
@devlead I actually like this idea of extending the `#l` directive to support downloading via HTTP.

devlead:
@Redth this basically is what you've started to look at right?

Redth:
Yes. This is what I started doing...

PR #157, opened 2015-01-16
Add capability to use load directive with web urls.

RichiCoder1:
Work in Progress Proof of Concept for how loading scripts from the web could work.

### DIFF
--- a/src/Cake.Core/Scripting/Processors/LoadDirectiveProcessor.cs
+++ b/src/Cake.Core/Scripting/Processors/LoadDirectiveProcessor.cs
@@ -1,6 +1,7 @@
 using System;
-using Cake.Core.IO;
 using System.Globalization;
+using System.Net;
+using Cake.Core.IO;
 
 namespace Cake.Core.Scripting.Processors
 {
@@ -59,20 +60,32 @@ namespace Cake.Core.Scripting.Processors
             }
 
             var directoryPath = GetAbsoluteDirectory(currentScriptPath);
+            ScriptReference scriptReference;
             Uri scriptUri;
             if (!Uri.TryCreate(scriptPathString, UriKind.RelativeOrAbsolute, out scriptUri)
                 || !scriptUri.IsAbsoluteUri || scriptUri.IsFile)
             {
                 var scriptPath = new FilePath(scriptPathString);
-                if (scriptPath.IsRelative)
-                    scriptPath = scriptPath.MakeAbsolute(directoryPath);
-
-                processor.Process(scriptPath, context);
+                scriptReference = new ScriptReference(scriptPath);
             }
             else
             {
-                processor.Process(scriptUri, context);
+                // Download the script to a local directory.
+                FilePath tempFile = System.IO.Path.GetTempFileName();
+                try
+                {
+                    var client = new WebClient();
+                    client.DownloadFile(scriptUri, tempFile.FullPath);
+                }
+                catch (WebException ex)
+                {
+                    throw new Exception("Failed to load file from web. See inner exception.", ex);
+                }
+
+                scriptReference = new ScriptReference(tempFile.FullPath, scriptUri.ToString());
             }
+
+            processor.Process(scriptReference, context);
 
             return true;
         }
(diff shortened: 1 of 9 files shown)


RichiCoder1:
Related to #149

patriksvensson:
Cool! :+1: 

Just one initial thought: Maybe the load directive could download the code and keep the script processor simple?

Since the `Process` method always should process something on disc, maybe `public void Process(FilePath path, ScriptProcessorContext context)` could take something new like a `ScriptReference` that keeps an identifier (could be a path or URI or whatever) and and the path on disc? This way we could make sure that we don't import the same script twice and still be sure that we have something on disc to refer to. Would also be able to write better log messages since the directives that would instantiate those objects could add more contextual information in the future.

RichiCoder1:
Hmm, I think that would be better. Would want to find some way to ensure we don't download the same file twice, as having a unique identifier in Process itself wouldn't help that. Maybe hash the URL and use that as the temp file's name. I just went with the most straight-forward approach first.

RichiCoder1:
Implemented suggestion. Shares the same flaw as previously where two script references to the same script with a slightly different path string with not get rejected. I think only script file checksum will prevent that :P

src/Cake.Core/Scripting/ScriptReference.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Security.Cryptography;
+using System.Text;
+using Cake.Core.IO;
+
+namespace Cake.Core.Scripting
+{
+    /// <summary>
+    /// Unique Script Reference
+    /// </summary>
+    public class ScriptReference
```
patriksvensson: Class should be sealed. Properties should have only getters, and their backing fields should be readonly.

src/Cake.Core/Scripting/ScriptReference.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            Path = localPath;
+
+            if (origin == null)
+            {
+                origin = localPath.ToString();
+            }
+            Origin = origin;
+
+            var hasher = new SHA1CryptoServiceProvider();
```
patriksvensson: is calculating a hash for the path really necessary? Wouldn't the path be enough as an ID?

src/Cake.Core/Scripting/Processors/LoadDirectiveProcessor.cs
```diff
@@ -46,10 +48,44 @@ public override bool Process(IScriptProcessor processor, ScriptProcessorContext
...
+                var scriptPath = new FilePath(scriptPathString);
+                scriptReference = new ScriptReference(scriptPath);
+            }
+            else
+            {
+                // Download the script to a local directory.
+                FilePath tempFile = System.IO.Path.GetTempFileName();
+                try
+                {
+                    var client = new WebClient();
```
patriksvensson: Should make this mockable so we can write tests for this. Maybe a `IWebClient` in `Cake.Core.Net`?

src/Cake.Core/Scripting/ScriptReference.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Gets the local script path.
+        /// </summary>
+        public FilePath Path { get; private set; }
+
+        /// <summary>
+        /// Gets the script's unique token identifier;
+        /// </summary>
+        public string TokenId { get; private set; }
```
patriksvensson: I think the name `Id` would be a better name.

patriksvensson:
Sorry for taking so long to responding to this.

I don't think it's that important to validate checksums. As long as we're not loading the same file (same path or same url) I think it will be fine. If we discover that this is something that is common for some reason, we can take a look at checksum validation.

I think your solution looks good. I've left you some comments of things I think should be addressed before merging.

src/Cake.Core/Scripting/Processors/LoadDirectiveProcessor.cs
RichiCoder1: Hrmm. I think that might be excessive, but it would allow use to unit test this.

RichiCoder1:
Updated with changes except WebClient request.

devlead:
@RichiCoder1 this has unfortunately gone to far out of sync with Cake. Sorry for not getting back to you earlier.
We're closing this for now, please feel free to re-open if you have the opportunity to update.

PR #160, opened 2015-01-20
Fixed globber bug. Closes #159.

### LINKED ISSUE #159: Globber returns no result when given file path
```csharp
var result = GetFiles("./existing_file.dll");
Debug.Assert(result.Count == 1); // Fails
```

The following work though:

```csharp
var result = GetFiles("./**/existing_file.dll");
Debug.Assert(result.Count == 1); // Success

var result2 = GetFiles("./*existing_file.dll");
Debug.Assert(result2.Count == 1); // Success
```

patriksvensson:


devlead:
Looks solid, works for me. Tested a script before & after, after works.

PR #166, opened 2015-02-06
Added filter support to CleanDirectory

devlead:
This commit adds possibility to filter which files / directories get deleted by the `CleanDirectory` alias.
```csharp
    CleanDirectory("C:/temp/deploy/TestDeployRepo", entry=>!entry.Hidden);
```
Will result in something like
```
Deleting contents of C:/temp/deploy/TestDeployRepo
Skipping C:/temp/deploy/TestDeployRepo/.git because specified predicate
Deleting contents of C:/temp/deploy/TestDeployRepo/bin
```

This is done with adding a where predicate to `Cake.Core.IO.IDirectory` `GetDirectories` & `GetFiles`

Also added `Hidden` property to `Directory` & `File` to expose if file/directory is hidden and a common `IFileSystemInfo` do enable a common predicate for `CleanDirectory`

Added new overloads & tests for this, also added option to specify action to get "callback" when files are excluded, use this in the `Cake.Common.IO.DirectoryCleaner::Clean` method to not delete directories when child directories/files still remain after clean and verbose output which files are skipped.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/FileSystemFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/FileSystemFixture.cs
@@ -24,6 +24,7 @@ namespace Cake.Common.Tests.Fixtures
             fileSystem.GetCreatedDirectory("/Temp/Hello/More/Empty");
             fileSystem.GetCreatedDirectory("/Temp/Hello/World");
             fileSystem.GetCreatedDirectory("/Temp/Goodbye");
+            fileSystem.GetCreatedDirectory("/Temp/Hello/Hidden", true);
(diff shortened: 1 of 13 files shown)


RichiCoder1:
This makes me think another nice addition might be Globber support in CleanDirectory as well as "!dir/file.txt" support for exluding files from Globbing.

devlead:
@RichiCoder1 agree, the main work it's done for this to be implemented in more places like the globber.

devlead:
@RichiCoder1 just wanted to keep this commit as small as possible.

RichiCoder1:
Understandably. Was more getting it out of my brain than anything :)

PR #180, opened 2015-02-23
Implemented a file hash calculator.

vktr:
A small but useful alias for calculating file hashes. This can be used when (for example) publishing packages online and one wants to publish the file hash alongside it.

The default hash is SHA256 which is the same as Powershell uses as default hash in the `Get-FileHash` cmdlet.

## Usage example

```csharp
// Calculate the default SHA256 hash.
FileHash sha256Hash = CalculateFileHash("./some-file.txt");
Information("Hash: " + sha256Hash.ToHex());

// Calculate the MD5 hash.
FileHash md5Hash = CalculateFileHash("./some-file.txt", HashAlgorithm.MD5);
Information("Hash: " + md5Hash.ToHex());
```

### DIFF
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -89,6 +89,7 @@
     <Compile Include="Unit\Build\AppVeyor\Data\AppVeyorRepositoryInfoTests.cs" />
     <Compile Include="Unit\Build\BuildSystemAliasesTests.cs" />
     <Compile Include="Unit\Build\BuildSystemTests.cs" />
+    <Compile Include="Unit\Security\FileHashCalculatorTests.cs" />
     <Compile Include="Unit\Solution\Project\Properties\AssemblyInfoCreatorTests.cs" />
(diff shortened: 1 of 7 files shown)


devlead:
@vktr nice, looks like an excellent feature, we will take a look as soon as we can.
Meanwhile it would be excellent if you could rebase into one commit and add tests.

RichiCoder1:
I'd also add the recommendation that you change `Algorithm.Md5` to something mroe specific, and also match the case of the algorithm names. ex:
`HashAlgorithm.MD5`, `FileHashAlgorithm.SHA256`

vktr:
@RichiCoder1 Sure, `HashAlgorithm` is a better name, however the casing goes against the naming guidelines so I didn't want to add exceptions until approved.

RichiCoder1:
I'm not sure what you mean by the casing. Acronyms are a generally accept exception to PascalNaming rules.
https://msdn.microsoft.com/en-us/library/ms229043(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.security.cryptography.md5(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.security.cryptography.sha256(v=vs.110).aspx

PR #183, opened 2015-02-24
Updated version information

devlead:
Updated version design


vktr:
I realize this is tagged as `in progress` so you might already have something, but a `-nologo` option would be nice so you can disable it.

devlead:
@vktr yes only displayed when -version command is specified.

Issue #172, opened 2015-02-11
Query/possible bug with multiple arguments


DavidR91:
If I invoke my build with the following arguments:

`"$(SolutionDir)cake/cake.exe" build.cake -verbosity=diagnostic -root="$(SolutionDir)" -scomDllConfiguration="$(ConfigurationName)"`

I have problems using the argument 'root' since it ends up as a string with the actual value of:

`C:\dev\squaredup-custbuild\v1" -scomDllConfiguration=Release`

Where C:\dev\squaredup-custbuild\v1 is the actual value I passed into 'root' and the remainder is the other argument.

What is the correct way to pass both of these arguments without ending up with an amalgamation of their values?

If the issue here is caused by quotes, what can I do to solve it? (Removing the quotes is not an option: This invocation can use absolute paths which contain spaces) 

**EDIT**:
Interestingly, if I only pass root (-root="$(SolutionDir)") the value is correct, but has an additional quote on the end, leading to 'illegal character in path'. 

If I omit the last quote, everything actually works fine (!!) (-root="$(SolutionDir)). Since I'm invoking cake from VS2013 post-build events I'm not sure if I'm looking at a cake bug here, or a VS bug?

DavidR91:
OK, after testing this out directly from the command line, it's definitely Visual Studio being crap -_-

(http://stackoverflow.com/questions/11587394/how-to-put-double-quotes-in-vs2010-post-build-step#comment36704367_11620677)

You can close this issue - apologies, I had no reason VS was incapable of parsing double quotes properly

patriksvensson:
I see. But more importantly, did you get it to work?
If quotes are a problem, would perhaps wrapping the task execution in a custom MSBuild task solve it?

DavidR91:
Yes, although back to possibly being a cake bug :-1: 

This command works:

`"C:\dev\squaredup-custbuild\v1\cake/cake.exe" -verbosity=diagnostic -squaredUpRoot="C:\dev\squaredup-custbuild\v1" -scomDllConfiguration="Release"`

But this does not:

`"C:\dev\squaredup-custbuild\v1\cake/cake.exe" -verbosity=diagnostic -squaredUpRoot="C:\dev\squaredup-custbuild\v1\" -scomDllConfiguration="Release"`

Note the trailing slash on "C:\dev\squaredup-custbuild\v1\ for the non-working version

Is there something inside cake which is reinterpreting the \" as the escape sequence for double quote? 

[...]

PR #185, opened 2015-03-01
Fixed script path bug.

patriksvensson:
Fixes a bug that might have to do with issue #172.

If a script with the path `test/build.cake` is passed to `Cake.exe`, The script runner would set the `test` directory as the working directory (as expected) but would forget to strip the directory part of the script path which would result in a working directory `./test` and the script path `test/build.cake`, which would evaluate to the absolute script path `./test/test/build.cake`.

Since the convention so far has been to put the build scripts in the root directory, we simply haven't noticed this...


devlead:
@patriksvensson tried 
```powershell
.\src\Cake\bin\Release\Cake.exe .\src\Cake\bin\Debug\XmlTransform.cake
```
And with this patch it finds the script :+1:

devlead:
Merged :+1:

PR #188, opened 2015-03-01
Added XmlTransform Aliases for XSL/XML transformations

devlead:
This PR propsed new aliases for transforming xml data via xml stylesheet.

This is done either via passing xml & xsl as string and getting transformed string back, or passing path to xml, xsl & result.

Example usage with paths:
```csharp
XmlTransform("breakfast_menu.xsl", "breakfast_menu.xml", "breakfast_menu.htm");
```
Example usage with strings:
```csharp
var xls = System.IO.File.ReadAllText("breakfast_menu.xsl", Encoding.UTF8);
var xml = System.IO.File.ReadAllText("breakfast_menu.xml", Encoding.UTF8);
var html = XmlTransform(xls, xml);
Information(html);
```

To both above methods there's an overload where you can pass settings to change generated XML output, example usage:
```csharp
XmlTransform("breakfast_menu.xsl", "breakfast_menu.xml", "breakfast_menu.htm", new XmlTransformationSettings
    {
        OmitXmlDeclaration = false,
        Encoding = Encoding.Unicode
    });
```


patriksvensson:
Looks good to me.

PR #199, opened 2015-03-13
Added support for lazy by verbosity logging

devlead:
Some logging takes time / resources and because of that we might only want to run them if i.e. `verbosity` is set to `Diagnostic`

Currently it will just omit the data when logging higher verbosity then `Cake` is launched with, but the work to evaluate the log message is till done.

That's why a propose that add an option where an function could be passed instead of `format` & `args` to the logging aliases.

A simple example of these new alias would be:
```csharp
Information(
    write => {
        //do some hard word
        write("This will only be called Verbosity>={0}", "Normal");
    });
``` 
The above message would only be evaluated and displayes if cake is launched withe the verbosities Normal, Verbose or Verbose

An more advanced example would be
```csharp
var tests = new KeyValuePair<string, Action<LazyLog>> [] {
                new KeyValuePair<string, Action<LazyLog>>("Error",         Error),
                new KeyValuePair<string, Action<LazyLog>>("Warning",       Warning),
                new KeyValuePair<string, Action<LazyLog>>("Information",   Information),
                new KeyValuePair<string, Action<LazyLog>>("Verbose",       Verbose),
                new KeyValuePair<string, Action<LazyLog>>("Debug",         Debug)
            };

Array.ForEach(
    tests,
    test=>{
        Context.Log.Write(Verbosity.Quiet, LogLevel.Information, "Testing {0}.", test.Key);
        bool funcEvaluated = false;
        LazyLog logFunc = write=> {
            funcEvaluated = true;
            write("Message from logging function ({0:yyyy-MM-dd})", DateTime.Now);
        };
        test.Value(logFunc);
        Context.Log.Write(
            Verbosity.Quiet,
            LogLevel.Information,
            "{0} {1} evaluated.\r\n",
            test.Key, (funcEvaluated) ? "was" : "wasn't");
    });
```

In this case if verbosity=Quiet the output would be
```
Testing Error.
Message from logging function (2015-03-14)
Error was evaluated.

Testing Warning.
Warning wasn't evaluated.

Testing Information.
Information wasn't evaluated.

Testing Verbose.
Verbose wasn't evaluated.

Testing Debug.
Debug wasn't evaluated.
```

And in case of diagnostic
```
Testing Error.
Message from logging function (2015-03-14)
Error was evaluated.

Testing Warning.
Message from logging function (2015-03-14)
Warning was evaluated.

Testing Information.
Message from logging function (2015-03-14)
Information was evaluated.

Testing Verbose.
Message from logging function (2015-03-14)
Verbose was evaluated.

Testing Debug.
Message from logging function (2015-03-14)
Debug was evaluated.
```

### DIFF
--- a/src/Cake.Common/Diagnostics/LoggingAliases.cs
+++ b/src/Cake.Common/Diagnostics/LoggingAliases.cs
@@ -26,6 +26,22 @@ namespace Cake.Common.Diagnostics
             }
             context.Log.Error(format, args);
         }
+        
+        /// <summary>
+        /// Writes an error message to the log using the specified log message action.
+        /// Evaluation message only if verbosity same or more verbose.
+        /// </summary>
+        /// <param name="context">the context.</param>
+        /// <param name="logAction">The function called for message when logging.</param>
+        [CakeMethodAlias]
+        public static void Error(this ICakeContext context, LogAction logAction)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+            context.Log.Error(logAction);
+        }
 
         /// <summary>
         /// Writes a warning message to the log using the specified format information.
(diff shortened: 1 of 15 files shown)


devlead:
Updated PR, refactored back `IVerbosityAwareLog` to Cake, added readonly verbosity property to `ICakeLog` to remove cast.

src/Cake.Common/Diagnostics/LoggingAliases.cs
```diff
@@ -90,5 +90,85 @@ public static void Debug(this ICakeContext context, string format, params object
             }
             context.Log.Debug(format, args);
         }
+
```
patriksvensson: Same here, try grouping the functions together.

patriksvensson:
LGTM :+1:

PR #203, opened 2015-03-15
Added available verbosities to Cake usage

devlead:
Added available verbosities to Cake usage, so don't need to lookup docs for these.

### DIFF
--- a/src/Cake/Commands/HelpCommand.cs
+++ b/src/Cake/Commands/HelpCommand.cs
@@ -1,4 +1,6 @@
-﻿using Cake.Core;
+﻿using System;
+using Cake.Core;
+using Cake.Core.Diagnostics;
 
 namespace Cake.Commands
 {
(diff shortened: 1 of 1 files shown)


src/Cake/Commands/HelpCommand.cs
```diff
@@ -28,6 +30,8 @@ public bool Execute(CakeOptions options)
             _console.WriteLine();
             _console.WriteLine("Options:");
             _console.WriteLine("    -verbosity=value    Specifies the amount of information to be displayed.");
+            _console.WriteLine("                        Verbosities: {0}",
+                string.Join(", ", Enum.GetNames(typeof(Verbosity))));
```
patriksvensson: I would prefer to add the verbosities directly here instead of enumerting them. Makes it easier to see how the formatting will look like as well.

```
-verbosity=value    Specifies the amount of information to be displayed.
                    (diagnostic, verbose, normal, minimal, quiet)
```

PR #214, opened 2015-03-18
Added #tool line directive

devlead:
Adds support for fetching tools via nuget

Example usage
```csharp
#Tool "ILMerge"
var assemblyPaths = GetFiles("./**/Release/Cake.*.dll");
var primaryAssemblyPath = GetFiles("./**/Release/Cake.exe").FirstOrDefault();
ILMerge(
     "./MergedCake.exe",
     primaryAssemblyPath,
     assemblyPaths,
     new ILMergeSettings { Internalize = true });
```
Will output something like
```
Installing 'ilmerge 2.14.1208'.
Successfully installed 'ilmerge 2.14.1208'.
```
and if verbosity is set to diagnostic it will also output any executables found
```
Found tool executable: [fullpath]]/Tools/ILMerge/tools/ILMerge.exe
```

### DIFF
--- a/src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
+++ b/src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
@@ -1,0 +1,191 @@
+﻿using System;
+using System.Linq;
+using Cake.Core.Diagnostics;
+using Cake.Core.IO;
+using Cake.Core.IO.NuGet;
+
+namespace Cake.Core.Scripting.Processors
+{
+    /// <summary>
+    /// Processor for #tool directives.
+    /// </summary>
+    public sealed class ToolDirectiveProcessor : LineProcessor
+    {
+        private static FilePath _nugetPath;
+        private readonly IFileSystem _fileSystem;
+        private readonly ICakeEnvironment _environment;
+        private readonly ICakeLog _log;
+        private readonly IToolResolver _nugetToolResolver;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ToolDirectiveProcessor" /> class.
+        /// </summary>
+        /// <param name="fileSystem">The file system.</param>
+        /// <param name="environment">The environment.</param>
+        /// <param name="log">The log.</param>
+        /// <param name="nugetToolResolver">The NuGet tool resolver.</param>
+        public ToolDirectiveProcessor(IFileSystem fileSystem, ICakeEnvironment environment, ICakeLog log, INuGetToolResolver nugetToolResolver)
+            : base(environment)
+        {
+            if (fileSystem == null)
+            {
+                throw new ArgumentNullException("fileSystem");
+            }
+            if (environment == null)
+            {
+                throw new ArgumentNullException("environment");
+            }
+            if (log == null)
+            {
+                throw new ArgumentNullException("log");
+            }
+            if (nugetToolResolver == null)
+            {
+                throw new ArgumentNullException("nugetToolResolver");
+            }
+
+            _fileSystem = fileSystem;
+            _environment = environment;
+            _log = log;
+            _nugetToolResolver = nugetToolResolver;
+        }
+
+        /// <summary>
+        /// Processes the specified line.
+        /// </summary>
+        /// <param name="processor">The script processor.</param>
+        /// <param name="context">The script processor context.</param>
+        /// <param name="currentScriptPath">The current script path.</param>
+        /// <param name="line">The line to process.</param>
+        /// <returns>
+        ///   <c>true</c> if the processor handled the line; otherwise <c>false</c>.
+        /// </returns>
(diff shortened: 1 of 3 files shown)


src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+
+            if (!directive.Equals("#tool", StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            // Fetch the tool NuGet ID.
+            var toolId = tokens
+                .Select(value => value.UnQuote())
+                .Skip(1).FirstOrDefault();
```
patriksvensson: What happens if someone write `#tool" with nothing else on the line? Will the `Skip` throw? If so, there should probably be a split and a count here to give the user a good error message.

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+
+            if (string.IsNullOrWhiteSpace(toolId))
+            {
+                return false;
+            }
+
+            // Fetch optional NuGet source.
+            var source = tokens
+                .Skip(2)
```
patriksvensson: Same here as the previous comment.

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+                .Select(value => value.UnQuote())
+                .FirstOrDefault();
+
+            // Get the directory path to Cake.
+            var applicationRoot = _environment.GetApplicationRoot();
+
+            // Get the tool directory.
+            var toolsRootDirectoryPath = applicationRoot
+                .Combine("..\\Tools")
```
patriksvensson: What happens if the user using the tool does not follow  the same conventions as we do, i.e. keeps `Cake.exe` in the root folder or at a location that's not expected. Will `../tools` still be a good path?

patriksvensson: The path `Tools` is written as `tools` in the rest of the source code.

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+            if (toolExecutables.Length == 0)
+            {
+                InstallTool(toolId, toolsRootDirectory, source);
+                toolExecutables = GetToolExecutables(toolDirectoryPath);
+            }
+
+            // Validate found assemblies.
+            if (toolExecutables.Length == 0)
+            {
+                throw new CakeException("Failed to find Tool executables");
```
patriksvensson: Missing period. `Tool` should be `tool`.

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+            });
+            process.WaitForExit();
+        }
+
+        private FilePath GetNuGetPath()
+        {
+            var nugetPath = _nugetPath ?? (_nugetPath = _nugetToolResolver.ResolveToolPath());
+            if (nugetPath == null)
+            {
+                throw new CakeException("Failed to find NuGet");
```
patriksvensson: Missing period.

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                return false;
+            }
+
+            // Fetch the tool NuGet ID.
+            var toolId = tokens
+                .Select(value => value.UnQuote())
+                .Skip(1).FirstOrDefault();
+
+            if (string.IsNullOrWhiteSpace(toolId))
```
devlead: @patriksvensson it will skip if tool id is null

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
devlead: source is optional

src/Cake.Core/Scripting/Processors/ToolDirectiveProcessor.cs
devlead: Perhaps use same folder as addins initially?

devlead:
Will take a look at the feedback probably tonight, agree with most;) and then update PR.

devlead:
Have updated and rebased against latest develop.

patriksvensson:
Sweet. Will merge ASAP :smile:

patriksvensson:
Merged

PR #224, opened 2015-03-31
Add TeamCity BuildSystem.

RichiCoder1:
PoC for TeamCity (and any other buildsystem) integration. Planning to use as basis for MyGet provider.

### DIFF
--- a/src/Cake.Common/Build/BuildSystem.cs
+++ b/src/Cake.Common/Build/BuildSystem.cs
@@ -1,4 +1,7 @@
-﻿using Cake.Common.Build.AppVeyor;
+﻿using System;
+using Cake.Common.Build.AppVeyor;
+using Cake.Common.Build.TeamCity;
+using Cake.Core;
 
 namespace Cake.Common.Build
 {
@@ -9,14 +12,25 @@ namespace Cake.Common.Build
     public sealed class BuildSystem
     {
         private readonly IAppVeyorProvider _appVeyorProvider;
+        private readonly ITeamCityProvider _teamCityProvider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="BuildSystem"/> class.
         /// </summary>
-        /// <param name="appVeyorProvider">The AppVeyor service.</param>
-        public BuildSystem(IAppVeyorProvider appVeyorProvider)
+        /// <param name="appVeyorProvider">The AppVeyor Provider.</param>
+        /// <param name="teamCityProvider">The TeamCity Provider.</param>
+        public BuildSystem(IAppVeyorProvider appVeyorProvider, ITeamCityProvider teamCityProvider)
         {
+            if (appVeyorProvider == null)
+            {
+                throw new ArgumentNullException("appVeyorProvider");
+            }
+            if (teamCityProvider == null)
+            {
+                throw new ArgumentNullException("teamCityProvider");
+            }
             _appVeyorProvider = appVeyorProvider;
+            _teamCityProvider = teamCityProvider;
         }
 
         /// <summary>
@@ -40,6 +54,60 @@ namespace Cake.Common.Build
         }
 
         /// <summary>
+        /// Gets the AppVeyor Provider.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// if(BuildSystem.IsRunningOnAppVeyor)
+        /// {
+        ///     // Upload artifact to AppVeyor.
+        ///     BuildSystem.AppVeyor.UploadArtifact("./build/release_x86.zip");
+        /// }
+        /// </code>
+        /// </example>
+        public IAppVeyorProvider AppVeyor
+        {
+            get { return _appVeyorProvider; }
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the current build is running on TeamCity.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// if(BuildSystem.IsRunningOnTeamCity)
+        /// {
+        ///     TeamCity.ProgressMessage("Doing an action...");
+        ///     // Do action...
+        /// }
+        /// </code>
+        /// </example>
+        /// <value>
+        /// <c>true</c> if the build currently is running on TeamCity; otherwise, <c>false</c>.
+        /// </value>
+        public bool IsRunningOnTeamCity
+        {
+            get { return _teamCityProvider.IsRunningOnTeamCity; }
+        }
+
+        /// <summary>
+        /// Gets the AppVeyor Provider.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// if(BuildSystem.IsRunningOnAppVeyor)
+        /// {
+        ///     // Upload artifact to AppVeyor.
+        ///     BuildSystem.AppVeyor.UploadArtifact("./build/release_x86.zip");
+        /// }
+        /// </code>
+        /// </example>
+        public ITeamCityProvider TeamCity
+        {
+            get { return _teamCityProvider; }
+        }
+
+        /// <summary>
         /// Gets a value indicating whether the current build is local build.
         /// </summary>
         /// <example>
@@ -60,7 +128,7 @@ namespace Cake.Common.Build
         /// </value>
         public bool IsLocalBuild
         {
-            get { return !IsRunningOnAppVeyor; }
+            get { return !(IsRunningOnAppVeyor || IsRunningOnTeamCity); }
         }
     }
 }
--- a/src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
+++ b/src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
@@ -1,0 +1,215 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Linq;
+using Cake.Core;
+using Cake.Core.IO;
+
+namespace Cake.Common.Build.TeamCity
+{
+    /// <summary>
+    /// Responsible for communicating with TeamCity.
+    /// </summary>
+    public sealed class TeamCityProvider : ITeamCityProvider
+    {
+        private const string MessagePrefix = "##teamcity[";
+        private const string MessagePostfix = "]";
+        private static readonly Dictionary<string, string> SanitizationTokens =
+            new Dictionary<string, string>
+            {
+                { "|", "||" },
+                { "\'", "|\'" },
+                { "\n", "|n" },
+                { "\r", "|r" },
+                { "[", "|['" },
+                { "]", "|]" }
+            };
+
+        private readonly ICakeEnvironment _environment;
+
+        /// <summary>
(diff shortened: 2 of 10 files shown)


src/Cake.Common/Build/BuildSystem.cs
```diff
@@ -8,15 +11,19 @@ namespace Cake.Common.Build
...
+        private readonly Lazy<AppVeyorProvider> _appVeyorProvider;
+        private readonly Lazy<TeamCityProvider> _teamCityProvider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="BuildSystem"/> class.
         /// </summary>
-        /// <param name="appVeyorProvider">The AppVeyor service.</param>
-        public BuildSystem(IAppVeyorProvider appVeyorProvider)
+        /// <param name="context">The Cake context.</param>
+        public BuildSystem(ICakeContext context)
```
patriksvensson: I really think the providers should be injected here.

src/Cake.Common/Build/BuildSystem.cs
```diff
@@ -8,15 +11,19 @@ namespace Cake.Common.Build
     /// </summary>
     public sealed class BuildSystem
     {
-        private readonly IAppVeyorProvider _appVeyorProvider;
+        private readonly Lazy<bool> _isRunningOnBuildSystem;
```
patriksvensson: Would prefer to not have these evaluated lazily. No real performance overhead since the build system normally caches them. Also increases complexity.

src/Cake.Common/Build/BuildSystem.cs
```diff
@@ -60,7 +121,12 @@ public bool IsRunningOnAppVeyor
         /// </value>
         public bool IsLocalBuild
         {
-            get { return !IsRunningOnAppVeyor; }
+            get { return !_isRunningOnBuildSystem.Value; }
+        }
+
+        private bool IsRunningOnBuildSystem()
```
patriksvensson: This method could be removed.

src/Cake.Common/Build/AppVeyor/IAppVeyorProvider.cs
```diff
@@ -1,33 +0,0 @@
-﻿using Cake.Common.Build.AppVeyor.Data;
-using Cake.Core.IO;
-
-namespace Cake.Common.Build.AppVeyor
-{
-    /// <summary>
-    /// Represents a service that communicates with AppVeyor.
-    /// </summary>
-    public interface IAppVeyorProvider
```
patriksvensson: Would actually want to keep the interface for the AppVeyor provider since it allows mocking. Should be injected into to BuildSystem. Would also be nice if there was an ITeamCityProvider.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+        public static void ProgressMessage(string message)
+        {
+            WriteServiceMessage("progressMessage", message);
+        }
+
+        /// <summary>
+        /// Write a progressStart message to the TeamCity build log.
+        /// </summary>
+        /// <param name="message">Build log message.</param>
+        public static void StartProgress(string message)
```
patriksvensson: Would be nice to return an IDisposable for the different Start blocks that would call the corresponding End method on Dispose. This would allow for code like this:

using(var scope = TeamCity.WriteProgress("Hello World"))
{
    // Do stuff.
}

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+        public static void EndProgress(string message)
+        {
+            WriteServiceMessage("progressFinish", "message", message);
+        }
+
+        /// <summary>
+        /// Write the start of a message block to the TeamCity build log.
+        /// </summary>
+        /// <param name="blockName">Block name.</param>
+        public static void StartBlock(string blockName)
```
patriksvensson: I think these methods should be called something like `WriteStartBlock`.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+                            }
+                            return string.Format("{0}='{1}'", keypair.Key, Sanitize(keypair.Value));
+                        })
+                        .ToArray());
+            Console.WriteLine("{0}{1} {2}{3}", MessagePrefix, messageName, valueString, MessagePostfix);
+        }
+
+        private static string Sanitize(string source)
+        {
+            var toReplace = new Dictionary<string, string>
```
patriksvensson: Could be a static readonly field.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            WriteServiceMessage(messageName, new Dictionary<string, string> { { " ", attributeValue } });
+        }
+
+        private static void WriteServiceMessage(string messageName, string attributeName, string attributeValue)
+        {
+            WriteServiceMessage(messageName, new Dictionary<string, string> { { attributeName, attributeValue } });
+        }
+
+        private const string MessagePrefix = "##teamcity[";
```
patriksvensson: Move to top (under private fields).

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+namespace Cake.Common.Build.TeamCity
+{
+    /// <summary>
+    /// Responsible for communicating with TeamCity.
+    /// </summary>
+    public sealed class TeamCityProvider
+    {
+        private readonly ICakeEnvironment _environment;
+
+        internal TeamCityProvider(ICakeEnvironment environment)
```
patriksvensson: Make constructor public.

patriksvensson:
Awesome work! Left some comments in the code.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
RichiCoder1: That's a cool idea. I'll look into it.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
RichiCoder1: Good idea. Consistency.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
RichiCoder1: Will do.

src/Cake.Common/Build/TeamCity/TeamCityProvider.cs
RichiCoder1: Indeed.

RichiCoder1:
Addressed code review feedback. Let me know what you think!

src/Cake.Common/Build/BuildSystem.cs
```diff
@@ -9,14 +12,17 @@ namespace Cake.Common.Build
...
         /// <summary>
         /// Initializes a new instance of the <see cref="BuildSystem"/> class.
         /// </summary>
-        /// <param name="appVeyorProvider">The AppVeyor service.</param>
-        public BuildSystem(IAppVeyorProvider appVeyorProvider)
+        /// <param name="appVeyorProvider">The AppVeyor Provider.</param>
+        /// <param name="teamCityProvider">The TeamCity Provider.</param>
+        public BuildSystem(IAppVeyorProvider appVeyorProvider, ITeamCityProvider teamCityProvider)
         {
             _appVeyorProvider = appVeyorProvider;
```
patriksvensson: Should throw if either of the providers are `null`.

src/Cake.Common/Build/TeamCity/TeamCityDisposableHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Cake.Common.Build.TeamCity
+{
+    /// <summary>
+    /// Disposable helper for writing TeamCity message blocks.
+    /// </summary>
+    public sealed class TeamCityBlock : IDisposable
```
patriksvensson: I really like this :smile:

1. Maybe the `StartBlock` method could be removed and done explicitly in the method that creates the block. This way we keep all the logic in one place.
2. Maybe we could make one generic block class that that takes an `Action<ITeamCityProvider>` that you pass to the constructor (for writing the end block).
3. Maybe an explicit implemented `IDisposable` interface that calls an EndBlock method or something so you also have the option to not use the using pattern.

RichiCoder1: In regards to 1 & 2: Thinking about it now, I think I may just do what I've seen down elsewhere, and make a `DisposableHelper` of some sort, that wraps an action in an a Disposable wrapper. Could be used in other places then. That would move startblock out and remove the need for both these wrappers.

Regarding 3: I purposely kept `WriteEndBlock` available on the TC provider for that reason. If they're doing something that deviates from the normal `using`, they should just use the `Start` and `End` methods.

src/Cake.Common/Build/BuildSystem.cs
RichiCoder1: :+1:

src/Cake.Common/Build/TeamCity/TeamCityDisposableHelpers.cs
patriksvensson: 1. I would actually want to avoid adding functionality for the purpose of being reused. I would prefer that we implement stuff without usability in mind, and do this as a refactoring part later on i.e. if we see that a lot of stuff sprinkled across the project/solution uses the same mecanism.

2. I think we should choose whether or not to use the disposable pattern or Start/End. If we want to give the option to the user to choose, I would rather see Start/End implemented on the provider, and add specialized extensionmethods on top of it that extends the ITeamCityProvider with the IDisposable stuff (using the Start/End methods).

RichiCoder1:
Moved Block and Build block into extensions. Addressed other feedback.

patriksvensson:
Looks good to me :+1:
Just need to fix the StyleCop errors and it's ready to be merged.

The `Debug` configuration runs no StyleCop/FxCop-rules, only the `Release` configuration does that. You could also run the `build.ps1` before pushing to make sure that there are no errors. I should probably put this information in the documentation when I think about it.

patriksvensson:
@RichiCoder1 Just saw your post on Gitter, so you can ignore my previous comment about `build.ps1` :smile:

devlead:
@RichiCoder1 Got a question about TeamCity support, this PR doesn't build anymore, when you have time could you take look at rebasing with latest Cake version and fixing the warnings/feedback?

RichiCoder1:
Sqaushed and rebased.

patriksvensson:
Merged! :+1:

RichiCoder1:
Nice!

PR #232, opened 2015-04-15
Added Windows  8.1 & 10 OS Support

devlead:
Since Windows 8.1 Microsoft changed Win32 GetVersionEx that .Net uses for compatibility reasons report latest version app is compiled for.

This PR adds manifest aware of 8.1 & 10 to the Cake.exe, so `System.Environment.OSVersion` reports correct version up to Windows 10.

Example cake
```csharp
Information("OSVersion version: {0}", System.Environment.OSVersion);
```

Before fix it returns `6.2` on 8.1 & 10, after fix it returns `6.3` and `10.0`.

Example output on Windows 10 build 10049 before fix
```
OSVersion version: Microsoft Windows NT 6.2.9200.0
```
Example output Windows 10 build 10049 after fix
```
OSVersion version: Microsoft Windows NT 10.0.10049.0
```

### DIFF
--- a/src/Cake/Cake.csproj
+++ b/src/Cake/Cake.csproj
@@ -57,6 +57,9 @@
   <PropertyGroup>
     <StartupObject>Cake.Program</StartupObject>
   </PropertyGroup>
+  <PropertyGroup>
+    <ApplicationManifest>app.manifest</ApplicationManifest>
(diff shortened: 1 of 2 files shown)


patriksvensson:
Merged! :+1:

PR #233, opened 2015-04-17
Add NUnit options for redirecting output

RossHammer:
Nunit has options to redirect test stdout and stderr to a files. I made the options available in the NUnitSettings.

### DIFF
--- a/src/Cake.Common/Tools/NUnit/NUnitRunner.cs
+++ b/src/Cake.Common/Tools/NUnit/NUnitRunner.cs
@@ -124,6 +124,16 @@ namespace Cake.Common.Tools.NUnit
                 builder.Append("/trace:" + settings.Trace);
             }
 
+            if (settings.Output != null)
+            {
+                builder.AppendQuoted(string.Format(CultureInfo.InvariantCulture, "/output:{0}", settings.Output.MakeAbsolute(_environment).FullPath));
+            }
+
+            if (settings.ErrorOutput != null)
+            {
+                builder.AppendQuoted(string.Format(CultureInfo.InvariantCulture, "/err:{0}", settings.ErrorOutput.MakeAbsolute(_environment).FullPath));
+            }
+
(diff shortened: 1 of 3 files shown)


src/Cake.Common/Tools/NUnit/NUnitRunner.cs
```diff
@@ -124,6 +124,16 @@ private ProcessArgumentBuilder GetArguments(IEnumerable<FilePath> assemblyPaths,
                 builder.Append("/trace:" + settings.Trace);
             }
 
+            if (settings.OutputFile != null)
+            {
+                builder.AppendQuoted(string.Format(CultureInfo.InvariantCulture, "/output:{0}", settings.OutputFile.MakeAbsolute(_environment).FullPath));
```
patriksvensson: According to the [NUnit-Console Command Line Options](http://www.nunit.org/index.php?p=consoleCommandLine&r=2.6.3) page the parameter name should be `/out:`.

Is `/output` an alias?

patriksvensson:
@RossHammer: Thanks! :+1:
The code per se looks good to me, except one thing which is minor. I left a comment in the code.

RossHammer:
Output is just the long form and out is the short form. It doesn't matter what one is used so I just picked the long form version.

From nunit's help string
> /output=STR             File to receive test output (Short format: /out=STR)

patriksvensson:
@RossHammer I understand! In that case, a suggestion would be to change `err` to `error` (if there's a long form for that as well). This would make it consistent with the other flags.

What do you think?

RossHammer:
It's stupid but there is only one form for the argument otherwise I would have done that.
>/err=STR                File to receive test error output
>From nunit's help string

patriksvensson:
@RossHammer Then it all looks good to me! :smile:

patriksvensson:
Merged! :+1:
Thank you!

PR #237, opened 2015-04-27
Added aliases for making paths absolute.

patriksvensson:
Added two aliases for making paths absolute.

```csharp
// Get absolute file path.
var file = GetAbsoluteFilePath(File("./foo.txt"));

// Get absoltue directory path.
var dir = GetAbsoluteDirectoryPath(Directory("./foo"));
```

Not sure of the name of the aliases are good or consistent.
Any thoughts about this?


patriksvensson:
Maybe call the methods `GetAbsoluteFile` and `GetAbsoluteDirectory` instead since we never expose `IDirectory` or `IFile` directly in the DSL?

Issue #240, opened 2015-05-04
Allow support for the InternalsVisibleTo attribute.


cgourlay:
Can the AssemblyInfoCreator support the InternalsVisibleTo attribute? This is something I need to allow internals to be testable.

I am happy to make a pull request for this - is it right to assume that the change(s) are made in the Cake.Common.Solution.Project.Properties namespace?

patriksvensson:
Hello @cgourlay,

You're more than welcome to give it a try :smile: 

I would suggest adding a `ICollection<string>` to `AssemblyInfoSettings.cs` so you can provide multiple assemblies to be ignored.

cgourlay:
Thanks for the suggestion @patriksvensson - I'll look at this today.

cgourlay:
Hi @patriksvensson 

I've an initial implementation in my own fork here: b2486ff874d0177180a3d2e652b5ebe3a9dbdf46

I've looked at the sample 'build.cake' file and I'm not really following why there is a need to ignore multiple assemblies. Would you be able to expand on what you are looking for?

devlead:
Well you can expose your internals to multiple assemblies:
```csharp
[assembly:InternalsVisibleTo("Friend1a")]
[assembly:InternalsVisibleTo("Friend1b")]
```
https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.internalsvisibletoattribute(v=vs.110).aspx

cgourlay:
Of course! I've latched on to the suggestion of assemblies being ignored. 

Are we saying that it's enough to alter my implementation to handle an `ICollection<string>` of assembly names?

devlead:
Correct, `ICollection` is great as many inherit from it I.e. List, Array etc.

cgourlay:
Nice one! Thanks for clarifying. I'll get this sorted in a moment and submit a pull request.

cgourlay:
Mmmm. The `AssemblyInfoRegistration.cs` is using a Dictionary and preventing multiple entries from being added and I can see why this is desired behaviour for all the other attributes.

`private void Add(string name, string @namespace, string value)
        {
            if (Attributes.ContainsKey(name))
            {
                Attributes[name] = value;
            } ...`

Not being overly familiar with the codebase I'd welcome views on how best to proceed - rather than just jump in.

patriksvensson:
Hello @cgourlay,

Maybe something like:

```csharp
[...]

PR #243, opened 2015-05-06
[WIP] Added support for InternalsVisibleTo attribute as per issue #240

cgourlay:
The existing implementation of the `GetRegistrations` method uses a Dictionary to manage the attributes to be added to the AssInfo.  I ran into problems where the InternalsVisibleTo attribute needs to be applied n times. Worked round this by using a `List<KeyValuePair<string,string>>`. However, this loses a couple of things: prevention of adding duplicate attributes and StringComparer.OrdinalIgnoreCase.

Would appreciate feedback.

Cheers


src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreator.cs
```diff
@@ -109,9 +110,13 @@ private static AssemblyInfoRegistration GetRegistrations(AssemblyInfoSettings se
             registration.AddString("AssemblyCopyright", "System.Reflection", settings.Copyright);
             registration.AddString("AssemblyTrademark", "System.Reflection", settings.Trademark);
             registration.AddString("Guid", "System.Runtime.InteropServices", settings.Guid);
+            foreach (var assemblyName in settings.InternalsVisibleTo ?? Enumerable.Empty<string>())
```
patriksvensson: I think it would be a good idea to not treat the InternalsVisibleTo as a registration.

Maybe something like this in the `Create` method:

```csharp
using (var stream = _fileSystem.GetFile(absoluteOutputPath).OpenWrite())
using (var writer = new StreamWriter(stream, System.Text.Encoding.UTF8))
{
    // Write header.
    writer.WriteLine("//------------------------------------------------------------------------------");
    writer.WriteLine("// <auto-generated>");
    writer.WriteLine("//     This code was generated by Cake.");
    writer.WriteLine("// </auto-generated>");
    writer.WriteLine("//------------------------------------------------------------------------------");

    if (registration.Attributes.Count > 0)
    {
        // Write namespaces.
        var namespaces = registration.Namespaces.Select(n => string.Concat("using ", n, ";"));
        if(settings.InternalsVisibleTo.Count > 0)
        {
            // Add namespace for InternalsVisibleTo to the set.
            registration.Namespaces.Add("System.Runtime.CompilerServices");
        }
        foreach (var @namespace in namespaces)
        {
            writer.WriteLine(@namespace);
        }

        writer.WriteLine();

        // Write attributes.
        foreach (var attribute in registration.Attributes)
        {
            writer.WriteLine(string.Concat("[assembly: ", attribute.Key, "(", attribute.Value, ")]"));
        }

        // Write the internals visible to attributes.
        if(settings.InternalsVisibleTo.Count > 0)
        {
            writer.WriteLine();

            // Write attributes.
            foreach (var attribute in settings.InternalsVisibleTo)
            {
                writer.WriteLine(string.Concat("[assembly: InternalsVisibleTo(", attribute, ")]"));
            }
        }        
    }
}
```

cgourlay:
Hi @patriksvensson 

Maybe I am just not getting what a registration is. Would you be able to clarify what a a registration is and why InternalsVisibleTo differs from the other attributes?

Happy to change accordingly but if it's not a registration then the unit tests fail due to the loc: `if (registration.Attributes.Count > 0)`

patriksvensson:
Hi!

I'm at work at the moment, but I'll try to answer tonight when I get home.

patriksvensson:
Hello @cgourlay,

Sorry for the late reply...

The AssemblyInfoRegistration is just a container that formats values and makes sure that namespaces only are imported once, for all the attributes the user added via the `AssemblyInfoSettings` class.

I think your solution works fine for now. Just fix the errors/warnings and I'll merge it. You can test the build by changing to the `Release` configuration in Visual Studio or running `build.ps1` in the root.

/ Patrik

devlead:
Replaced by #265

Issue #108, opened 2014-09-05
Octopus Deploy: Deploy Release


patriksvensson:

devlead:
Solved by #244

PR #244, opened 2015-05-08
OctopusDeploy Create Release

matdube:
Implementation of the Create Release command for OctopusDeploy (https://github.com/cake-build/cake/issues/108)

### DIFF
--- a/src/Cake.Common/Tools/OctopusDeploy/CreateReleaseArgumentBuilder.cs
+++ b/src/Cake.Common/Tools/OctopusDeploy/CreateReleaseArgumentBuilder.cs
@@ -6,18 +6,14 @@ namespace Cake.Common.Tools.OctopusDeploy
     internal class CreateReleaseArgumentBuilder
     {
         private readonly string _projectName;
-        private readonly string _server;
-        private readonly string _apiKey;
         private readonly CreateReleaseSettings _settings;
         private readonly ICakeEnvironment _environment;
 
         private readonly ProcessArgumentBuilder _builder;
 
-        public CreateReleaseArgumentBuilder(string projectName, string server, string apiKey, CreateReleaseSettings settings, ICakeEnvironment environment)
+        public CreateReleaseArgumentBuilder(string projectName, CreateReleaseSettings settings, ICakeEnvironment environment)
         {
             _projectName = projectName;
-            _server = server;
-            _apiKey = apiKey;
             _settings = settings;
             _environment = environment;
             _builder = new ProcessArgumentBuilder();
@@ -50,17 +46,17 @@ namespace Cake.Common.Tools.OctopusDeploy
             _builder.AppendQuoted(_projectName);
 
             _builder.Append("--server");
-            _builder.Append(_server);
+            _builder.Append(_settings.Server);
 
             _builder.Append("--apiKey");
-            _builder.AppendSecret(_apiKey);
+            _builder.AppendSecret(_settings.ApiKey);
 
             AppendArgumentIfNotNull("username", _settings.Username);
 
             if (_settings.Password != null)
             {
-                _builder.Append("--" + "password");
-                _builder.AppendQuoted(_settings.Password);
+                _builder.Append("--password");
+                _builder.AppendQuotedSecret(_settings.Password);
             }
 
             AppendArgumentIfNotNull("configFile", _settings.ConfigurationFile);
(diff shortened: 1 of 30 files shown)


src/Cake.Common/Tools/OctopusDeploy/OctopusDeployRunner.cs
```diff
@@ -0,0 +1,81 @@
...
+using Cake.Core;
+using Cake.Core.IO;
+using Cake.Core.Utilities;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    /// <summary>
+    /// The OctopusDeploy runner
+    /// </summary>
+    public sealed class OctopusDeployRunner : Tool<CreateReleaseSettings>
```
RichiCoder1: Shouldn't this runner and it's argument builder names reflect that they're for *releases* specifically?

matdube: I thought it would be logicial to group both CreateRelease and Deploy in the same runner since they invoke the same exe.  Deploy is another issue, which I plan on doing next.  Would it make more sense if I included both in the same pull request ?

RichiCoder1: With the way Runners are structured, they're intended to be tightly coupled to a single set of settings. So you're have a create release runner, and a deploy runner, all which are bootstrapped by aliases. Correct me if I'm wrong @patriksvensson .

patriksvensson: Epic contribution! Thank you!

What @RichiCoder1 is saying is correct. A runner should be tightly coupled with specific settings. The thought behind this is to keep the runners specialized, single responsibility and easy to test. I know that there might be different opinions about this (there always is), but I think it's important to be consistent with the other tools like the [NuGet](https://github.com/cake-build/cake/tree/develop/src/Cake.Common/Tools/NuGet) tools.

I think a good idea would be to rename the `OctopusDeployRunner` to something like `OctopusDeployRelaseCreator` or similar that reflect the responsibilities of the tool.

@matdube, @RichiCoder1, @devlead: Any naming suggestions for this?

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployAliases.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.Annotations;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    /// <summary>
+    /// Contains functionality for working with Octopus Deploy.
+    /// </summary>
+    [CakeAliasCategory("Octo")]
```
patriksvensson: The category is used for the [documentation](http://cakebuild.net/dsl/), so something like `Octopus Deploy` would probably describe the aliases better.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployRunner.cs
matdube: Good point, thanks for the comments.  I'll update the PR accordingly.

patriksvensson:
Hello @matdube 

Everything looks good to me, but I was wondering about the arguments.

Since `server` and `apikey` always are required (according to [this](http://docs.octopusdeploy.com/pages/viewpage.action?pageId=360596)), wouldn't it be better to move them to `OctopusDeploySettings` (which contain shared settings) instead of having them as method arguments, and move the `project` parameter to `CreateReleaseSettings`?

This would give us the signature:

```csharp
OctoCreateRelease(this ICakeContext context, CreateReleaseSettings settings)
```

This would match other tools (like NuGet Push) better. 

Since the alias is the contract with the user (and difficult to change after it's been released), I want to get this right from the beginning. I hope you don't mind the nitpicking :)

What do you think?

src/Cake.Common/Tools/OctopusDeploy/CreateReleaseArgumentBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            _builder.Append("--apiKey");
+            _builder.AppendSecret(_apiKey);
+
+            AppendArgumentIfNotNull("username", _settings.Username);
+
+            if (_settings.Password != null)
+            {
+                _builder.Append("--" + "password");
+                _builder.AppendQuoted(_settings.Password);
```
devlead: Shouldn't this be appended secret?

patriksvensson: Yes, good catch! Use `AppendQuotedSecret` here to make sure it's not displayed in build logs etc.

devlead:
@matdube merged last evening, forgot to say thank you for contributing :+1:

devlead:
@matdube also be sure to add all your email addresses to GitHub so you get recognized as an contributor see 7fc0149e769eb21ff7a4618b28044ba5760769f3 or just go to https://github.com/settings/emails and add any email you have so it gets linked to your account.

PR #252, opened 2015-05-16
Refactoring of script alias handling and code generation.

patriksvensson:
Refactoring of how we process scripts to make it easier to perform script code transformation which we'll need to get Mono.CSharp running.

Do not merge this since it's a work in progress.


devlead:
Merge conflict resolved and merged.

PR #253, opened 2015-05-20
Fluent Arguments

SharpeRAD:
Adds a fluent interface to both the ArgumentBuilder and ProcessSettings
that allows arguments to be neatly chained.


SharpeRAD:
If you merge this and deploy a new package I can release my Powershell add-on that makes use of the fluent arguments.

PS: I :heart: cake

patriksvensson:
Hello @SharpeRAD,

First of all, let me thank you for contributing to Cake.
Cake :heart:  you to! :smile: 

I have some objections with the PR though.

1. I don't really see the need for an IProcessArgumentList<T> since all extensions for the ProcessBuilder already are fluent. Am I missing something here?

2. The NamedParameter is a good initiative, but it only covers one named parameter case, where the argument looks like `-key value`. It won't cover other cases such as `-key=value`, `/key=value`, `/key:value` (or other permutations of those). The reasoning about the ProcessArgumentBuilder is to be generic, so we haven't implemented any specific parameters for this reason. I actually think it would be better if you could provide this functionality directly from your PowerShell project for now.

3. No examples or use case of the added functionality in the PR description.

4. No unit tests for the added functionality.

This is not meant as discouragement in any way, but I think we need to have a discussion about the functionality proposed in this PR.

Best regards
Patrik

SharpeRAD:
Sorry I should have included an example! I'm allowing classes that use the ArgumentBuilder to share the same extension methods, making for some very clean alias calls:

```C#
StartProcess("Example.exe", ​new ProcessSettings().Append("Accept").Append("Request"));
```

It allows add-ons to use the ArgumentBuilder in new contexts, eg: in my PowershellSettings class.

Did i mention  I :heart: cake

SharpeRAD:
I accept your concerns regarding NamedArgument and address them in my latest commit with a format field:

```C#
StartProcess("Example.exe", ​new ProcessSettings().AppendNamed("Action", "AcceptPR", "/{0}={1}"));
```

You can also set the default format in the ArgumentBuilder constructor:

```C#
new ProcessArgumentBuilder("/{0}={1}").AppendNamed("Action", "AcceptPR");
```

SharpeRAD:
Fixed a few silly mistakes, added some missing extension methods and where possible replaced existing Append calls with AppendNamed equivalents. 

Final commit shows that the proposed changes work for all existing use cases.

patriksvensson:
Hello @SharpeRAD,

First of all, let me apologize for the late answer.

I'm really happy that you're entusiastic about your changes, but I'm afraid that we cannot accept this pull request in it's current state. It's simply too many things implemented at once, and no unit tests for the new/changed functionality.

I think the interface for the argument builder is a good idea, but I don't think we should change every call in the code base to be fluent (I'm thinking about the commit [Putting Fluent arguments to use](https://github.com/SharpeRAD/cake/commit/7481a247350cdce4d797566425a8cfdeb1b9c3cb)). I also think that this functionality would benefit from a discussion since it affects most tools and aliases.

The named parameter addition is also something that I feel is something that could be better implemented outside of Cake as an addin. This is something that is nice to have, but the construct we have today is kept simple for a reason. The NamedParameter does not feel very intutitive and I prefer to keep the functionality around the ArgumentBuilder simple and rather build some kind of wrapper that is more specialized in a case like this.

Please see the [Contribution Guidelines](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md), and make sure you submit one PR per feature. Also start an issue about things that you would like to work on to make sure that the additions you want is aligned with the Cake project.

We're happy to assist/continue the discussion if you have any follow-up questions or feedback. Come to out [Gitter room](https://gitter.im/cake-build/cake) and we'll be happy to continue the discussion.

Best regards
Patrik

PR #259, opened 2015-06-01
Added filter predicate overloads to globber & clean directory methods

devlead:
Sometimes when working with recursive globbering it can lead into a rabbit hole of loads of folders, given at best bad performance and at worst hitting the .Net max path length limitation.

This PR proposes you to set a predicate for folders / files you know there's no need to explorer / include.

Sample DSL usage `GetFiles` with predicate:
```csharp
Func<IFileSystemInfo, bool> exclude_node_modules =
    fileSystemInfo=>!fileSystemInfo.Path.FullPath.EndsWith(
                            "node_modules",
                            StringComparison.OrdinalIgnoreCase);

var solutions = GetFiles("./**/*.sln", exclude_node_modules);
```
Sample DSL `CleanDirectories` with predicate:
```csharp
Task("Clean")
    .Does(() =>
{
    // Clean solution directories.
    foreach(var path in solutionPaths)
    {
        Information("Cleaning {0}", path);
        CleanDirectories(path + "/**/bin/" + configuration, exclude_node_modules);
        CleanDirectories(path + "/**/obj/" + configuration, exclude_node_modules);
    }
});
```

### DIFF
--- a/src/Cake.Common.Tests/Unit/IO/DirectoryAliasesTests.cs
+++ b/src/Cake.Common.Tests/Unit/IO/DirectoryAliasesTests.cs
@@ -52,7 +52,7 @@ namespace Cake.Common.Tests.Unit.IO
                 // When
                 var result = DirectoryAliases.Directory(context, "./temp");
 
-                // Then                
+                // Then
                 Assert.IsType<ConvertableDirectoryPath>(result);
             }
         }
@@ -80,7 +80,7 @@ namespace Cake.Common.Tests.Unit.IO
 
                 // When
                 var result = Record.Exception(() =>
-                    context.CleanDirectory(null));
+                    DirectoryAliases.CleanDirectory(context, null));
 
                 // Then
                 Assert.IsArgumentNullException(result, "path");
@@ -96,7 +96,7 @@ namespace Cake.Common.Tests.Unit.IO
                 context.FileSystem.Returns(fixture.FileSystem);
 
                 // When
-                context.CleanDirectory(directory);
+                DirectoryAliases.CleanDirectory(context, directory);
 
                 // Then
                 Assert.Empty(fixture.FileSystem.GetDirectory(directory).GetFiles("*", SearchScope.Recursive));
@@ -119,7 +119,7 @@ namespace Cake.Common.Tests.Unit.IO
                     .ToArray();
 
                 // When
-                context.CleanDirectory(directory, wherePredicate);
+                DirectoryAliases.CleanDirectory(context, directory, wherePredicate);
 
                 // Then
                 Assert.Empty(fixture.FileSystem.GetDirectory(directory).GetFiles("*", SearchScope.Recursive).Where(wherePredicate));
@@ -136,7 +136,7 @@ namespace Cake.Common.Tests.Unit.IO
                 context.FileSystem.Returns(fixture.FileSystem);
 
                 // When
-                context.CleanDirectory(directory);
+                DirectoryAliases.CleanDirectory(context, directory);
 
                 // Then
                 Assert.Empty(fixture.FileSystem.GetDirectory(directory).GetDirectories("*", SearchScope.Recursive));
@@ -158,7 +158,7 @@ namespace Cake.Common.Tests.Unit.IO
                     .ToArray();
 
                 // When
-                context.CleanDirectory(directory, wherePredicate);
+                DirectoryAliases.CleanDirectory(context, directory, wherePredicate);
 
                 // Then
                 Assert.Empty(fixture.FileSystem.GetDirectory(directory).GetDirectories("*", SearchScope.Recursive).Where(wherePredicate));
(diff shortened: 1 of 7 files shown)


src/Cake.Common.Tests/Unit/IO/DirectoryAliasesTests.cs
```diff
@@ -196,6 +196,48 @@ public void Should_Create_Directory_If_Missing()
...
+                    .ToArray();
+
+                var preGoodbyeFiles = context.FileSystem
+                    .GetDirectory("/Temp/Goodbye/")
+                    .GetFiles("*.*", SearchScope.Recursive)
+                    .Select(file=>file.Path.FullPath)
+                    .ToArray();
+
+                // When
+                context.CleanDirectory(directory, predicate=>predicate.Path.FullPath!="/Temp/Goodbye");
```
patriksvensson: Use the `DirectoryAliases` type instead of calling the extension method to make it clear what's under test.

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -144,6 +144,30 @@ public static void CleanDirectories(this ICakeContext context, string pattern)
         }
 
         /// <summary>
+        /// Cleans the directories matching the specified pattern.
+        /// Cleaning the directory will remove all it's content but not the directory itself.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="pattern">The pattern to match.</param>
+        /// <param name="predicate">The predicate used to filter directories based on file system information.</param>
+        /// <example>
```
patriksvensson: Move `example` tag so that parameters are at the bottom.

src/Cake.Common/IO/GlobbingAliases.cs
```diff
@@ -41,6 +41,39 @@ public static FilePathCollection GetFiles(this ICakeContext context, string patt
         }
 
         /// <summary>
+        /// Gets all files matching the specified pattern.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="pattern">The glob pattern to match.</param>
+        /// <param name="predicate">The predicate used to filter files based on file system information.</param>
+        /// <returns>A <see cref="FilePathCollection" />.</returns>
+        /// <example>
```
patriksvensson: Move `example` tag so that parameters are at the bottom.

src/Cake.Core/IO/Globber.cs
```diff
@@ -56,6 +56,19 @@ public Globber(IFileSystem fileSystem, ICakeEnvironment environment)
...
+
+        /// <summary>
+        /// Returns <see cref="Path" /> instances matching the specified pattern.
+        /// </summary>
+        /// <param name="pattern">The pattern to match.</param>
+        /// <param name="predicate">The predicate used to filter directories based on file system information.</param>
+        /// <returns>
+        ///   <see cref="Path" /> instances matching the specified pattern.
+        /// </returns>
+        public IEnumerable<Path> Match(string pattern, Func<IFileSystemInfo, bool> predicate)
```
patriksvensson: My opinion here is that I would rather see the filtering be done in the alias methods to keep the Globber clean and straight forward.

devlead: But then it's too late, it will already walked thru and gotten an exception or taken a long time.

RichiCoder1:
Out of curiosity, is there any reason this is a predicate rather than an exclude pattern.

devlead:
@RichiCoder1 predicate is more powerful as it let's you typed go against any property on the objects i.e. only apply to files, or hidden etc.

src/Cake.Core/IO/Globber.cs
patriksvensson: Ok, I understand. Maybe we should remove the other overload then, and introduce the predicate less method as an extension method to keep the interface lean.

src/Cake.Common.Tests/Unit/IO/DirectoryAliasesTests.cs
devlead: Sure copied an existing test, will fix that thru out the class it was already inconsistent, though what you do miss then is testing if it's an extension method though which is required for aliases ;)

src/Cake.Common/IO/DirectoryAliases.cs
devlead: Ok will go thru and fix that and all existing example inconsistencies in that class.

src/Cake.Common/IO/GlobbingAliases.cs
devlead: Ok will go thru and fix that and all existing example inconsistencies in that class.

src/Cake.Core/IO/Globber.cs
devlead: Sure can do.

devlead:
Refactored according to feedback.

patriksvensson:
Looks good to me now. OK to merge.

PR #264, opened 2015-06-20
Fixed issue where method was missing on interface declaration.

patriksvensson:
The method `UpdatedBuildVersion` was missing in the `IAppVeyorProvider` interface declaration. Also fixed some minor, build system related StyleCop issues.


patriksvensson:
Crucial that this PR is merged before next release since all Cake builds will fail otherwise when updating the Cake tool version :smile:

Issue #267, opened 2015-06-30
Report formatting messed up when long task names


marcosnz:
If a task name is longer than 30 characters then the report is messed up and hard to read. I am dynamically creating task names which is leading to longer than usual names.
```
Task                          Duration
--------------------------------------------------
Build-Order                   00:00:00.0081275
Nuget-Restore                 00:00:04.1434154
Build-ReleaseDebug-Common-Library00:00:06.2929425
Build-ReleaseDebug-Services-All00:00:20.0597646
Build-ReleaseDebug-Module-Analysis00:00:03.9095751
Build-ReleaseDebug-Module-Documents00:00:04.7935222
Build-ReleaseDebug-Module-Messaging00:00:02.2321549
Build-ReleaseDebug-Site-App   00:00:23.4270291
Build-ReleaseDebug-Site-Assets00:01:11.4500985
Build-ReleaseDebug-Site-Authentication00:00:04.0556270
Build-ReleaseDebug-Site-Instance00:00:02.4231884
Build-ReleaseDebug-Site-Manage00:00:19.0323366
Build-ReleaseDebug-Site-My    00:00:25.8938716
Build-ReleaseDebug-Site-Pay   00:00:02.9447329
Build-ReleaseDebug-Site-Widget00:00:01.7430201
Default                       00:00:00.0117797
--------------------------------------------------
Total:                        00:03:12.4211861
```

marcosnz:
The #268 PR leaves most reports the same. If any of the task names are greater than 30 characters long it widens the report e.g. the above now looks like:
```
Task                                   Duration
-----------------------------------------------------------
Build-Order                            00:00:00.0107507
Nuget-Restore                          00:00:03.3841406
Build-ReleaseDebug-Common-Library      00:00:03.2489604
Build-ReleaseDebug-Services-All        00:00:09.3761721
Build-ReleaseDebug-Module-Analysis     00:00:01.9135796
Build-ReleaseDebug-Module-Documents    00:00:02.7120511
Build-ReleaseDebug-Module-Messaging    00:00:01.6872557
Build-ReleaseDebug-Site-App            00:00:13.5204454
Build-ReleaseDebug-Site-Assets         00:00:44.0441283
Build-ReleaseDebug-Site-Authentication 00:00:02.1955409
Build-ReleaseDebug-Site-Instance       00:00:01.5250945
Build-ReleaseDebug-Site-Manage         00:00:10.7844529
[...]

PR #268, opened 2015-06-30
Fix formatting when long task names used. Fix for #267

marcosnz:


patriksvensson:
Thanks! I have been irritated about this for months but never got around to actually do something about it! Will merge this tonight after work! :+1:

devlead:
@marcosnz this is now merged, thanks for your contribution! :+1:

Issue #81, opened 2014-07-31
Mono support


patriksvensson:

diegobfernandez:
Is there anything I could do to help this to move on?
I'm kind of desperate for an alternative build system for .NET, but I'm using mono.

patriksvensson:
Hello @diegobfernandez!

We absolutely need help with this one! I actually had a conversation about this with @devlead just a couple of days ago, so the timing is perfect.

However, we're not sure if we should add support for Mono via something like [Mono.CSharp](http://docs.go-mono.com/index.aspx?link=N:Mono.CSharp) (available on NuGet as well) or try to get this working with CoreClr using Roslyn. Maybe the `Mono.CSharp` approach would be the best one for now until CoreClr "settled" a little more.

I think the easiest approach would be to create a new [IScriptSessionFactory](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Scripting/IScriptSessionFactory.cs) and [IScriptSession](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Scripting/IScriptSession.cs), and configure the container in [Program.cs](https://github.com/cake-build/cake/blob/develop/src/Cake/Program.cs) to use those if running on Mono.

If you want to give this a shot, we would appreciate it! :+1:

patriksvensson:
*This is a brain dump for the Mono.CSharp support.*

Let's say we have the following Cake script.

```csharp
public class Test
{
	public int MagicNumber
	{
		get { return 42; }
	}

	public void DoMagic()
	{
		Console.WriteLine("Hello from {0}", MagicNumber);
	}
}

public int MagicNumber()
{
	return 99;
}

public void DoStuff(Action<int> stuff)
{
	Console.WriteLine("Doing stuff");
}

Task("Hello")
   .Does(() =>
{
	new Test().DoMagic();
});
```


To get this to run using Mono.CSharp, we would have to evaluate the different blocks individually. We would also need to rewrite the methods to something that could be executed individually outside a class scope, such as a delegate variable.

**1) Extract and evaluate the class.**

```csharp
public class Test
{
	public int MagicNumber
	{
		get { return 42; }
	}

	public void DoMagic()
	{
		Console.WriteLine("Hello from {0}", MagicNumber);
	}
}
```

**2) Rewrite the MagicStuff method to a delegate.**

```csharp
Func<int> MagicNumber = delegate() {
	return 99;
}
```

**3) Rewrite the DoStuff method to a delegate.**

[...]

PR #269, opened 2015-07-01
Initial work on Mono.CSharp scripting support

Redth:
This is a work-in-progress PR for supporting Mono.

I took some advice from #81 and created an implementation of `IScriptSession` and `IScriptEngine` which use the `Mono.CSharp` script hosting.

This doesn't work yet, there's some obvious issues to work through, but the scaffolding is there for someone to help jump in who might be more familiar with `Mono.CSharp`.

Currently the code is generated almost identically to how the Roslyn generator works.  The `Evaluator` doesn't like this, and I believe things like the aliases will have to be wrapped up in a class to make work.

Any input is appreciated.


devlead:
Cool, mono support is high on my wish list!

Hopefully at one point we'll be able to leverage the Roslyn work done for Mono/XamarinStudio
https://github.com/mono/roslyn

There are a few StyleCop issues in the code, think that's one reason it fails on AppVeyor now.

Really cool to see someone's starting to pursue this :+1:

gep13:
@Redth I noticed in the Gitter Chat that you mentioned you are using xBuild to run the build under Mono.  As a result, you might not see the StyleCop errors that are being generated.  You can however seen them listed out in the MSBuild log, which starts here:

https://ci.appveyor.com/project/cakebuild/cake/build/0.4.2-build-524#L283

Hope that helps!

Redth:
@gep13 Thanks, it also appears that the StyleCop addin for Xamarin Studio does the job too :)

gep13:
> @Redth said...
Thanks, it also appears that the StyleCop addin for Xamarin Studio does the job too :)

Good to know, thanks! :smile_cat:

Redth:
Closing in favour of #280

PR #271, opened 2015-07-02
DownloadFile alias

Redth:
I do a lot of downloading of external files in my build scripts. 

This adds a new `HttpAliases` class with (so far) a single `DownloadFile(string url, DirectoryPath outputPath)` alias which does exactly what it looks like.

Eventually there could be more HTTP related aliases here such as `DownloadString`


Redth:
Whoops, didn't mean to base this off of the mono support... Will resend.

PR #273, opened 2015-07-02
Added DownloadFile alias

Redth:
I do a lot of downloading of external files in my build scripts. 

This adds a new `HttpAliases` class with (so far) a single `DownloadFile(string url, DirectoryPath outputPath)` alias which does exactly what it looks like.

Eventually there could be more HTTP related aliases here such as `DownloadString`

### DIFF
--- a/src/Cake.Common/IO/Net/HttpAliases.cs
+++ b/src/Cake.Common/IO/Net/HttpAliases.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.Annotations;
+using Cake.Core.Diagnostics;
+using Cake.Core.IO;
+
+namespace Cake.Common
+{
+    /// <summary>
+    /// Contains functionality related to HTTP operations
+    /// </summary>
+    [CakeAliasCategory("HTTP Operations")]
+    public static class HttpAliases
+    {
+        /// <summary>
+        /// Downloads the specified File over HTTP to the specified local output path
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="url">URL of file to download.</param>
+        /// <param name="outputPath">Where to download the file to.</param>
+        [CakeMethodAlias]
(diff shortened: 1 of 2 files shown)


src/Cake.Common/IO/Net/HttpAliases.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                throw new ArgumentNullException("outputPath");
+            }
+
+            context.Log.Write(Verbosity.Verbose, LogLevel.Information, "Downloading File: {0}", url);
+
+            // We track the last posted value since the event seems to fire many times for the same value
+            var percentComplete = 0;
+
+            var http = new System.Net.WebClient();
```
devlead: As we're using this in a sync fashion we should wrap in a using block, so streams/responses gets deallocated.

devlead:
@Redth hi looks good just left one comment.

devlead:
@Redth Looks good, I'll take a look later to day and merge if I don't find anything else.

RichiCoder1:
Random idea would also be an alias which does the same thing, but drops it in the appropriate "temp" directory and passed back a path/`IFilePath`. Fun food for thought.

patriksvensson:
I like @RichiCoder1's idea as well. Ok if I add tomorrow?

devlead:
@RichiCoder1 @patriksvensson download string / bytes would be nice additions to. if your at it;)

patriksvensson:
Well, download a file as string/byte feels unsafe. Wouldn't it be better to
return a stream in that case and add convenience methods to convert that to
string/byte array? That way, everyone will opt in for that functionality
instead of doing it like that by mistake.

devlead:
@patriksvensson methods allready exists on the standard .net WebClient which this pr uses 
[`string DownloadString(String)`](https://msdn.microsoft.com/en-us/library/fhd1f0sw(v=vs.110).aspx) and [`byte[] DownloadData(string address)`](https://msdn.microsoft.com/en-us/library/xz398a3f(v=vs.110).aspx)
so would be easy to add aliases for.

PR #274, opened 2015-07-04
Changed namespace and added overloads for HTTP aliases.

patriksvensson:
* Changed namespace to `Cake.Common.Net`.
* Added overloads that accept an `System.Uri`.
* Added overload that generates a temporary filename.
* Added sub category for aliases for documentation.


patriksvensson:
Merged

Issue #276, opened 2015-07-07
Error thrown if addin dll can't be loaded


marcosnz:
If I reference ```#addin MSBuild.Extension.Pack``` I get errors in ```Cake.Core.Scripting.ScriptAliasFinder.FindAliases``` like ```System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.```
The error is raised because some of the dlls installed by tthe addin rely on other dlls that may not be installed if you aren't targeting that particular dll. 
But there are no aliases in the dlls so don't I think we need to fail at this point.
I will add a PR soon that logs a warning but continues.

PR #277, opened 2015-07-07
#276 log warning if dll not be loaded in FindAliases

marcosnz:
The code traps the load error and logs warning but continues without failing.
Also logs diagnostic information re Fusion binding if Verbosity is Diagnostic.


patriksvensson:
Merged :+1:

PR #280, opened 2015-07-08
Mono support

Redth:
This adds basic mono support and XBuild support.

For scripts that have defined classes, properties, or methods, you will need to wrap them in `#region "Extras"` code regions to be properly processed.  This is a current limitation of the mono scripting and should eventually be fixed.

You can invoke `XBuild` to build projects and solutions.  There is also a `MSOrXBuild` alias which will route to `MSBuild` on windows and `XBuild` on linux/mac.  You can still explicitly use `XBuild` on windows if you like.

There are still some limitations in locating tools used in some aliases, but the basic scripting itself should be functional :+1:


patriksvensson:
Merged.

gep13:
woot! :+1:

PR #284, opened 2015-07-17
Added alias to detect if build script is running on Windows or a Unix based system.

patriksvensson:
Two variants depending on what you prefer to use.

```csharp
if (IsRunningOnWindows())
{
    Information("Windows!");
}
```

or

```csharp
if (IsRunningOnUnix())
{
    Information("Not Windows!");
}
```


patriksvensson:
Merged

PR #288, opened 2015-07-17
Nuget Proxy

SharpeRAD:
Adds an Alias for setting the Nuget proxy

### DIFF
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\xunit.runner.visualstudio.2.0.1\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\packages\xunit.runner.visualstudio.2.0.1\build\net20\xunit.runner.visualstudio.props')" />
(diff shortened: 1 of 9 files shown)


patriksvensson:
Will take a closer look at it later but it looks good from what I can see :smile:

src/Cake.Common.Tests/Cake.Common.Tests.csproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\packages\xunit.runner.visualstudio.2.0.1\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\packages\xunit.runner.visualstudio.2.0.1\build\net20\xunit.runner.visualstudio.props')" />
```
patriksvensson: Should be removed.

src/Cake.Common.Tests/packages.config
```diff
@@ -4,4 +4,5 @@
   <package id="xunit.abstractions" version="2.0.0" targetFramework="net45" />
   <package id="xunit.core" version="2.0.0" targetFramework="net45" />
   <package id="xunit.extensibility.core" version="2.0.0" targetFramework="net45" />
+  <package id="xunit.runner.visualstudio" version="2.0.1" targetFramework="net45" />
```
patriksvensson: Should not be added.

patriksvensson:
It looks good to me, but I don't think we should add support for the xUnit Visual Studio runner. At least not in this commit. I also don't think we should add a build.bat in this commit.

Other than that it looks good to me :smile:

SharpeRAD:
Sorry force of habit! I add both by default now as its nice to have a build shortcut in the solution folder and execute individual tests from the explorer. Should be good to merge now. 

Would you be opposed to me adding them in another PR?

patriksvensson:
No problem! I wouldn't be opposed to adding the visual studio runner, but the bat script feels a little bit redundant in my opinion since it would be a bootstrapper for the bootstrapper.

patriksvensson:
Merged :+1:

SharpeRAD:
YAY! Thanks for the blog mention too :smile:

patriksvensson:
:balloon: :smile: :tada:

PR #291, opened 2015-07-19
Rewrote FakeFileSystem to behave more like a real file system.

patriksvensson:
This means that it's much easier to write tests that touches the file system. The old fake file system was kind of cumbersome to work with, but now it should be much easier.

This also means that old tests that uses `NSubstitute` to set up globbing expectations now can create a real globber and setup the file system accordingly.

```csharp
// Prepare
var environment = FakeEnvironment.CreateUnixEnvironment();
var fileSystem = new FakeFileSystem(environment);
var globber = new Globber(fileSystem, environment);

// Setup a file system.
fileSystem.CreateDirectory("/Working");
fileSystem.CreateDirectory("/Temp");
fileSystem.CreateFile("/Temp/Hello.txt").SetContent("Hello World");
fileSystem.CreateFile("/Temp/Hidden.txt").Hide();

// Try getting a file.
var file = fileSystem.GetFile("/Temp/Hidden.txt");
Debug.Assert(file.Exists);

// Try finding a file.
var matches = globber.GetFiles("./**/Hello.txt");
Debug.Assert(matches.Count() == 1);
```

### DIFF
--- a/src/Cake.Testing/Fakes/FakeFileSystemTree.cs
+++ b/src/Cake.Testing/Fakes/FakeFileSystemTree.cs
@@ -1,0 +1,275 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Cake.Core;
+using Cake.Core.IO;
+using Path = Cake.Core.IO.Path;
+
+namespace Cake.Testing.Fakes
+{
+    internal sealed class FakeFileSystemTree
+    {
+        private readonly PathComparer _comparer;
+        private readonly FakeDirectory _root;
+
+        public PathComparer Comparer
+        {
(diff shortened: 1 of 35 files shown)


src/Cake.Testing/Fakes/FakeFileSystemTree.cs
```diff
@@ -0,0 +1,245 @@
...
+
+            return directory;
+        }
+
+        public void DeleteDirectory(FakeDirectory fakeDirectory, bool recursive)
+        {
+            var root = new Stack<FakeDirectory>();
+            var result = new Stack<FakeDirectory>();
+
+            root.Push(fakeDirectory);
```
patriksvensson: Should make sure that `fakeDirectory` really exist before pushing it on the root.

PR #294, opened 2015-07-20
changed msbuild verbosity

### LINKED ISSUE #282: Cake uses wrong MSBuild verbosity parameter
When using the extension method [MSBuildSettings.SetVerbosity(Verbosity.Verbose)](http://cakebuild.net/api/cake.common.tools.msbuild/949206fc/ba92223e), MSBuild is called with `/v:verbose` instead of `/v:detailed`.

```csharp
MSBuild("./Project.sln", settings =>
    settings.SetVerbosity(Verbosity.Verbose));
```

**Correct mapping should be:**

* Quiet => `quiet`
* Minimal => `minimal`
* Normal => `normal`
* Verbose => `detailed`
* Diagnostic => `diagnostic`

ChrisMissal:
MSBuild uses the flag 'detailed' instead of 'verbose'. This applies
the correct mapping to fix the discrepancy.

Closes #282 - "Cake uses wrong MSBuild verbosity parameter"


patriksvensson:
Perfect! :+1:

Issue #285, opened 2015-07-17
Add Support For Roundhouse


ChrisMissal:
As an avid user of [Roundhouse](https://github.com/chucknorris/roundhouse) for database migrations, I'd like to see it supported as a Cake tool.

gep13:
:+1:  @ferventcoder might be interested in hearing about this :smile_cat:

ferventcoder:
:+1:

ferventcoder:
This could possibly be what UppercuT uses for it's next engine, provided it runs on mono.

RichiCoder1:
It's dated at this point, but I'd imagine that support would probably be best added through an addon like https://github.com/RichiCoder1/Cake.AliaSql.

@ferventcoder I'd have to double check with @patriksvensson, but Mono support was recently added and is being stabilized and expanded. Would probably love testers and users :)

ferventcoder:
UC supports a ton of different tools though - how is the plugin engine for this @patriksvensson ?

gep13:
>@ferventcoder said...
This could possibly be what UppercuT uses for it's next engine, provided it runs on mono.

The grunt work has been done here:
https://github.com/cake-build/cake/pull/280

I have been thinking about using it on a couple projects, just need to get the cycles...

ferventcoder:
Perhaps I can get @drusellers excited about cake?

RichiCoder1:
@ferventcoder Cake has the notions of [Aliases](http://cakebuild.net/docs/fundamentals/aliases), which usually provide an excellent starting point for third party tools an extensions. Cake itself uses them to implement most of it support, as most of Cake.Common is simply commonly used aliases.

The more the merrier :D

@gep13 Been chewing on introducing Cake to Gui for a while. Already use it extensively at my company. About the only problem we've had is poor editing experience and it's something I've been bugging the omnisharp team about.

gep13:
>@RichiCoder1 said...
Been chewing on introducing Cake to Gui for a while

I have been thinking the same.  Will be sorry to see psake leave, it has served me very well!

patriksvensson:
@ferventcoder Plugins is mainly supported via script aliases (like @RichiCoder1 already said). There is also some pre-processor directives that can be used to import tools and assemblies directly from NuGet. Now when I think about it, I should probably add more documentation around extensibility :smile: 

@gep13 Come to the dark side, we have :cake:

gep13:
[...]

PR #297, opened 2015-07-20
Added support for Roundhouse DB Migrations

ChrisMissal:
Initial implementation for Roundhouse support (Issue #285).

### DIFF
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -81,6 +81,7 @@
     <Compile Include="Fixtures\Tools\NUnitRunnerFixture.cs" />
     <Compile Include="Fixtures\ProcessFixture.cs" />
     <Compile Include="Fixtures\Tools\OctopusDeployReleaseCreatorFixture.cs" />
+    <Compile Include="Fixtures\Tools\RoundhouseRunnerFixture.cs" />
     <Compile Include="Fixtures\Tools\SignToolResolverFixture.cs" />
     <Compile Include="Fixtures\Tools\SignToolSignRunnerFixture.cs" />
     <Compile Include="Fixtures\TextTransformationFixture.cs" />
@@ -149,6 +150,7 @@
     <Compile Include="Unit\Tools\NUnit\NUnitRunnerTests.cs" />
     <Compile Include="Unit\Tools\NUnit\NUnitSettingsTests.cs" />
     <Compile Include="Unit\Tools\OctopusDeploy\OctoCreateReleaseTests.cs" />
+    <Compile Include="Unit\Tools\Roundhouse\RoundhouseRunnerTests.cs" />
     <Compile Include="Unit\Tools\SignTool\SignToolResolverTests.cs" />
     <Compile Include="Unit\Tools\SignTool\SignToolSignRunnerTests.cs" />
     <Compile Include="Unit\Tools\WiX\LightRunnerTests.cs" />
--- a/src/Cake.Common.Tests/Fixtures/Tools/RoundhouseRunnerFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/RoundhouseRunnerFixture.cs
@@ -1,0 +1,44 @@
+﻿using Cake.Common.Tools.Roundhouse;
+using Cake.Core;
+using Cake.Core.IO;
+using NSubstitute;
+
+namespace Cake.Common.Tests.Fixtures.Tools
+{
+    internal sealed class RoundhouseRunnerFixture
+    {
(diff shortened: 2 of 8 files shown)


ChrisMissal:
@patriksvensson This does the trick to get RH tool support. I can imagine more aliases being created once I start using this more, or from others using it, but I wanted to keep things simple to start.

:cake:

patriksvensson:
I think that the alias should be called `RoundhouseMigrate` rather than just `Migrate`. This would be good so it won't be ambiguous with other tools like EF migrations or FluentMigrator if we want to introduce support for them in the future.

Is any of the data sensitive in the sense that you wouldn't want them to appear in a build log such as database connection strings? In that case I would suggest appending a [SecretArgument](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/Arguments/SecretArgument.cs). 

The ProcessArguments are composable so you could do like this:

```csharp
var secret = new SecretArgument(new TextArgument("--amg=986b0e4a"));
builder.AppendQuoted(secret);
```

This would appear as `rh.exe "[REDACTED]"` in the build log.

Great work btw :+1:

ChrisMissal:
@patriksvensson Good catches, fixed, thanks!

RichiCoder1:
Random food for thought might also be making Roundhouse a property like AppVeyor and such, and then expose the methods on it rather then have them be straight aliases. I won't the be the end user though, so it's just a thought.

ChrisMissal:
I'm open to that, but I've only just begun using Cake, but that seems to make sense to me.

patriksvensson:
I think it would fit in better with the other tool if we just exposed alias methods like it is now.

AppVeyor is a little bit special since it has a lot of nested properties - and to be honest I'm not super happy with it since discoverability is really bad.

Good idea though :smile:

ChrisMissal:
Digging in a bit more, I'm not sure there's much to be gained by that approach. Since everything needed is on the settings object, the method calls wouldn't have arguments, so passing the settings in would essentially be the same thing. Here's an example of how I'm using it now:

```csharp
var target = Argument("target", "Migrate");

var settings = new RoundhouseSettings {
    ToolPath = @"path\to\rh",
    Environment = "LOCAL",
    ConnectionString = "Server=localhost;Database=Blarg;Trusted_Connection=True;MultipleActiveResultSets=true",
    SqlFilesDirectory = @"src\DatabaseMigration",
    Silent = true,
};

Task("Rebuild")
  .IsDependentOn("Drop")
  .IsDependentOn("Migrate");

Task("Migrate")
  .Does(() => {
    Migrate(settings);
  });

Task("Drop")
  .Does(() => {
    Drop(settings);
  });

RunTarget(target);
```

Correct me if I'm wrong, but that change would look more like this:

```csharp
var target = Argument("target", "Migrate");

var settings = new RoundhouseSettings {
    ToolPath = @"path\to\rh",
    Environment = "LOCAL",
    ConnectionString = "Server=localhost;Database=Blarg;Trusted_Connection=True;MultipleActiveResultSets=true",
    SqlFilesDirectory = @"src\DatabaseMigration",
    Silent = true,
};
var rh = Roundhouse.Create(settings); // or something to get a context based on settings?

Task("Rebuild")
  .IsDependentOn("Drop")
  .IsDependentOn("Migrate");

Task("Migrate")
  .Does(() => {
    rh.Migrate();
  });

Task("Drop")
  .Does(() => {
    rh.Drop();
  });

RunTarget(target);
```

If that hurts discoverability, I'd vote against it.

***Mostly just thinking out loud here...***

RichiCoder1:
Oh, I was thinking more like:
```c#
var target = Argument("target", "Migrate");

var settings = new RoundhouseSettings {
    ToolPath = @"path\to\rh",
    Environment = "LOCAL",
    ConnectionString = "Server=localhost;Database=Blarg;Trusted_Connection=True;MultipleActiveResultSets=true",
    SqlFilesDirectory = @"src\DatabaseMigration",
    Silent = true,
};

Task("Rebuild")
  .IsDependentOn("Drop")
  .IsDependentOn("Migrate");

Task("Migrate")
  .Does(() => {
    Roundhouse.Migrate(settings);
  });

Task("Drop")
  .Does(() => {
    Roundhouse.Drop(settings);
  });

RunTarget(target);
```

Interesting thought creating an object though.

patriksvensson:
Yes, but I think it would be better to keep to the conventions used by the other tools. AppVeyor and TeamCity have so many properties and methods that it made sense at the tim. 

If I could break stuff like crazy I would probably remove alias properties completely :smile:

ChrisMissal:
Sounds like changes for the v1.0 release ;)

I don't have any more commits to add, so unless anybody has any other feedback, I'd call this PR good to go.

patriksvensson:
Merged! :+1:

PR #303, opened 2015-07-24
Fixes Issue #302: Add XmlPoke Aliases

RichiCoder1:


patriksvensson:
Nice! :+1: 
Will take a closer look at and merge this after the v0.5.1 release.

ChrisMissal:
Excellent! :+1:

src/Cake.Common/Xml/XmlPokeAliases.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// Set the value of, or remove, target nodes.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="source">The source xml to transform.</param>
+        /// <param name="destination">The destination to write too.</param>
+        /// <param name="xpath">The xpath of the nodes to set.</param>
+        /// <param name="value">The value to set too. Leave blank to remove the selected nodes.</param>
+        /// <param name="settings">Additional settings to tweak Xml Poke behavior.</param>
+        [CakeMethodAlias]
+        public static void XmlPoke(this ICakeContext context, XmlReader source, XmlWriter destination, string xpath, string value, XmlPokeSettings settings)
```
patriksvensson: Don't see much reason to expose this. I suspect the two above methods will suffice.

src/Cake.Common/Xml/XmlPokeSettings.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    /// Contains settings for <see cref="XmlPokeAliases"/>
+    /// </summary>
+    public sealed class XmlPokeSettings
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="XmlPokeSettings"/> class.
+        /// </summary>
+        public XmlPokeSettings()
```
patriksvensson: Properties before constructor.

src/Cake.Common/Xml/XmlPokeAliases.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        /// <summary>
+        /// Set the value of, or remove, target nodes.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="filePath">The target file.</param>
+        /// <param name="xpath">The xpath of the nodes to set.</param>
+        /// <param name="value">The value to set too. Leave blank to remove the selected nodes.</param>
+        /// <param name="settings">Additional settings to tweak Xml Poke behavior.</param>
+        [CakeMethodAlias]
```
patriksvensson: Would be great if we could add an example of how to use the alias with the `example` XML tag.

src/Cake.Common/Xml/XmlPokeAliases.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// Set the value of, or remove, target nodes.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="sourceXml">The source xml to transform.</param>
+        /// <param name="xpath">The xpath of the nodes to set.</param>
+        /// <param name="value">The value to set too. Leave blank to remove the selected nodes.</param>
+        /// <param name="settings">Additional settings to tweak Xml Poke behavior.</param>
+        /// <returns>Resulting XML.</returns>
+        [CakeMethodAlias]
+        public static string XmlPoke(this ICakeContext context, string sourceXml, string xpath, string value, XmlPokeSettings settings = null)
```
patriksvensson: Would be great if we could add an example of how to use the alias with the `example` XML tag.

src/Cake.Common/Xml/XmlPokeSettings.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to preserve white space.
+        /// </summary>
+        public bool PreserveWhitespace { get; set; }
+
+        /// <summary>
+        /// Gets or sets namespaces to include for xpath recognition.
+        /// </summary>
+        public Dictionary<string, string> Namespaces { get; set; }
```
patriksvensson: Should probably be `IDictionary<string, string>` to follow other tool alias.

patriksvensson:
@RichiCoder1 I've taken a look at it now and it looks good.

I left some comments about some thoughts I had. Minor stuff :smile:

src/Cake.Common/Xml/XmlPokeAliases.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            if (settings == null)
+            {
+                settings = new XmlPokeSettings();
+            }
+
+            using (var resultStream = new MemoryStream())
+            using (var sourceReader = new StringReader(sourceXml))
+            {
+                var xmlReader = XmlReader.Create(sourceReader);
```
devlead: I believe `XmlReader` & `XmlWriter` inherit from `IDisposable` and should be wrapped in a using statement.

RichiCoder1: I thought so too, but it looks like that's redundant, as the Dispose on those two just clean up the underlying streams (which we're already doing).

src/Cake.Common/Xml/XmlPokeAliases.cs
```diff
@@ -0,0 +1,177 @@
...
+                settings = new XmlPokeSettings();
+            }
+
+            var sourceFile = context.FileSystem.GetFile(filePath);
+            if (!sourceFile.Exists)
+            {
+                throw new FileNotFoundException("Source File not found.", sourceFile.Path.FullPath);
+            }
+            
+            using (var sourceReader = sourceFile.Open(FileMode.Open, FileAccess.ReadWrite, FileShare.None))
```
devlead: Minor as `sourceFile`& `sourceReader` used for both read&write their naming might be a bit misleading, perhaps `file` & `fileStream` is more appropriate.

src/Cake.sln.DotSettings
```diff
@@ -15,6 +15,7 @@
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/Abbreviations/=NET/@EntryIndexedValue">NET</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/Abbreviations/=VS/@EntryIndexedValue">VS</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateStaticReadonly/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="_" Suffix="" Style="aaBb" /&gt;</s:String>
+	<s:String x:Key="/Default/CodeStyle/TypeScriptCodeStyle/ImportReferenceStyle/@EntryValue">ECMAScript6</s:String>
```
devlead: Don't think this file is necessary for this PR.

src/Cake.Common/Xml/XmlPokeSettings.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    public sealed class XmlPokeSettings
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="XmlPokeSettings"/> class.
+        /// </summary>
+        public XmlPokeSettings()
+        {
+            Namespaces = new Dictionary<string, string>();
+            Encoding = Encoding.Default;
```
devlead: As the most common encoding is `UTF8` and Default is unknown (based on system settings) I think `UTF8` should be the "default".

devlead:
Looks ok to me, left some minor comments and agree with the issues @patriksvensson raised.

src/Cake.Common/Xml/XmlPokeAliases.cs
devlead: Ok, just checked the docs which stated `Closes the XmlWriter and the underlying stream/TextReader (if Settings.CloseOutput is true).` which led me to believe it did something more. Haven't checked IL so might be as you say, if so it's a non issue.

devlead:
Also would be nice with at least one test case to ensure expected behavior and ti have as an sample of usage,

patriksvensson:
@RichiCoder1 i'm planning on making a v0.5.2 release tomorrow (or later today). Should I move this to the v0.5.3 milestone?

src/Cake.Common/Xml/XmlPokeAliases.cs
RichiCoder1: Will do.

src/Cake.Common/Xml/XmlPokeAliases.cs
RichiCoder1: Wil do! Will add.

src/Cake.Common/Xml/XmlPokeSettings.cs
RichiCoder1: Can do.

src/Cake.Common/Xml/XmlPokeAliases.cs
RichiCoder1: I'll test again to be sure.

RichiCoder1:
Updated to address comments and add examples.

patriksvensson:
Perfect. Great work! :+1:

patriksvensson:
Merged!

PR #304, opened 2015-07-25
Rewrote Mono script tokenizer and parser.

### LINKED ISSUE #301: Add proper tokenizer and parser for script blocks
The script block parser for the Mono script engine that we added to distinguish between lines of code that should be executed and scopes such as methods/classes is kind of unstable/naive.

I'm currently working on this.

patriksvensson:
Not perfect but it should work better now. There might still be some corner cases left though, but nothing I have encountered. Could also probably (definitely) be simplified a lot.

This should fix #301.


patriksvensson:
Merged.

PR #308, opened 2015-07-27
Merge from original

marcosnz:


marcosnz:
Sorry, don't think i did that right :(

PR #317, opened 2015-07-27
GH-316: Enable Specifying Target without Switch

### LINKED ISSUE #316: Enable Specifying Target without Switch
Currently Cakes command line arguments are structured like so:
`Usage: Cake.exe [build-script] [-switch=value]`
This requires you to specify your Target by passing in `-Target=Default`.

I'd like to be able to specify the target without the prefix like so:
`Usage: Cake.exe [build-script] [target] [-switch=value]`

Allowed:
- `cake`
- `cake build.cake`
- `cake install`
- `cake build.cake install`
- `cake build.cake -Target=Install`` (mostly for back compat)

Disallowed:
- `cake install build.cake`
- `cake build.cake build.stuff`
- `cake install -Target=SomethingElse`
- `cake build.cake install -Target=SomethingElse`

RichiCoder1:
Fixes #316: Allow accepting Target as a positional parameter as well as a switch.
Also rewrites the argument parser slightly to process all arguments,
appending positions, and then process them.


patriksvensson:
Sorry for not responding to this until now. Will take a look at it today :smile:

src/Cake/Arguments/Argument.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Cake.Arguments
+{
+    internal struct Argument
```
patriksvensson: No advantage to use a struct here. Should be a class.

src/Cake/Arguments/Argument.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        /// <summary>
+        /// Gets or sets the string value for this argument.
+        /// </summary>
+        public string Value { get; set; }
+
+        /// <summary>
+        /// Gets or sets the position in which the argument appeared.
+        /// </summary>
+        public uint Position { get; set; }
```
patriksvensson: Could be an `Int32`.

src/Cake/Arguments/ArgumentParser.cs
```diff
@@ -96,7 +64,7 @@ private static bool IsOption(string arg)
             return arg[0] == '-';
         }
 
-        private bool ParseOption(string arg, CakeOptions options)
+        private Argument ParseOption(string arg, uint position)
```
patriksvensson: Could be an `Int32`.

patriksvensson:
It seems to do the job, but the argument parser have become kind of complex. This was the case before the change so it's not your fault, but adding more logic to it didn't make it better :smile:

I think that most of the logic could be simplified in the future by writing a proper tokenizer, but we can postpone that until later.

Left some minor comments in the code.

/ Patrik

RichiCoder1:
Agreed on proper tokenizer support. This'll have to do in the meantime.

RichiCoder1:
Updated to address PR feedback.

PR #319, opened 2015-07-27
increase visibility of skipped tasks

### LINKED ISSUE #305: Increase visibility of skipped tasks
It's easy to miss that a task was skipped when looking at the build log.

```
Copy-Files
Executing task: Copy-Files
Finished executing task: Copy-Files
Skipping task: Package
```

Would me much nicer if the skipped task was written to the build log like executed tasks.

```
Package
Skipping task: Package
```

This is just an adjustment to the [DefaultExecutionStrategy](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/DefaultExecutionStrategy.cs#L85).

ChrisMissal:
resolves #305

### DIFF
--- a/src/Cake.Core/DefaultExecutionStrategy.cs
+++ b/src/Cake.Core/DefaultExecutionStrategy.cs
@@ -84,6 +84,10 @@ namespace Cake.Core
         {
             if (task != null)
             {
+                _log.Verbose(string.Empty);
+                _log.Verbose("========================================");
+                _log.Verbose(task.Name);
+                _log.Verbose("========================================");
                 _log.Verbose("Skipping task: {0}", task.Name);
             }
         }


patriksvensson:
Great!

A small suggestion; the verbosity of the other added lines should probably be `Verbose` as well. Otherwise it will look like the task was executed when using `normal` verbosity.

I didn't mention this in the issue, so it's completely my fault.

ChrisMissal:
@patriksvensson This approach matches `PerformSetup`, `PerformTeardown`, and `Execute`. For example:
https://github.com/ChrisMissal/cake/blob/issue305/src/Cake.Core/DefaultExecutionStrategy.cs#L30-L34

Should I change them all to match?

patriksvensson:
Yes, you are completely right about this.

However, skipped tasks are different, because the build log would be very confusing if verbosity was set to `Normal`. In the example below, the task `Package` would appear in the build log, but not in the summary report.

Without looking closely at the report, the impression would be that `Package` was executed although it shouldn't.

```
Package

Default

Task                          Duration
--------------------------------------------------
Clean                         00:00:00.0186653
Restore-NuGet-Packages        00:00:00.3709869
Patch-Assembly-Info           00:00:00.0191192
Build                         00:00:05.5307239
Run-Unit-Tests                00:00:01.9940009
Copy-Files                    00:00:00.0129129
Zip-Files                     00:00:00.0404134
Default                       00:00:00.0055309
--------------------------------------------------
Total:                        00:00:08.0066584
```

ChrisMissal:
Ah! Thanks for explaining, I amended the commit.

RichiCoder1:
Does the build summary also show skipped in the place of duration? That'd be cool.

ChrisMissal:
It's possible if it's something that would be beneficial. I would imagine it would be off by default.

patriksvensson:
Good idea. Maybe not for this PR but maybe we could output SKIPPED in the duration column in yellow if verbosity >= verbose (or something like that).

RichiCoder1:
Exactly.

patriksvensson:
Merged :+1:

PR #322, opened 2015-07-30
Globber rewrite

### LINKED ISSUE #275: Globber performance
I am working with a largish repo and have found that the GetFiles method of the Globber is significantly slower than the MSBuild equivalent I am trying to replace..
If you take a look at [this Gist](https://gist.github.com/marcosnz/74285d2e7dd14048fb47) you will see that I am just getting a count of assembly info files. The MSBuild example does it in less than a second where as the cake version takes about a minute.

patriksvensson:
This is an attempty to refactor and speed up the globber. it's faster and much easier to follow and debug. Next step is to identify corner cases and add proper tests for them.

**This might contain bugs**, but would appreciate if people would try it out to make sure we don't break any core functionality in Cake.

### Benchmarks

`Globber.Match(string pattern)` and `CleanDirectory(string pattern)` benchmarks using the old and new globber on a directory root with **3,964** files and **1,238** folders, using the expression `C:/Source/Git/patriksvensson/**/bin`.

* **Old globber**
  * Matching: 7.4157083 seconds
  * Cleaning: 7.7243717 seconds

* **New globber**
  * Matching: 0.4628712 seconds
  * Cleaning: 0.7900062 seconds

Resolves #275


patriksvensson:
Merged.

PR #327, opened 2015-08-03
Fix for using (IDisposable) statement

devlead:
if space comes between `using` and `(` in using `IDisposable` statement it gets treated as an namespace. example
```csharp
using (new MemoryStream())
{
    new StringBuilder();
}
```

fails before this fix.


patriksvensson:
Merged :+1:

PR #333, opened 2015-08-06
Fix for PlatformTarget not used in MSBuild runner

### LINKED ISSUE #332: MSBuildRunner ignores PlatformTarget setting provided.
MSBuildRunner does not contain implementation for PlatformTarget specification in MSBuild arguments string.

os-alan:
Implemented PlatformTarget specification in MSBuild arguments string. MSBuildRunner.cs simply doesn't have it.

Closes #332

UPD: set it to WIP state, to implement Patrik's suggestions (see comments)

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
@@ -335,6 +335,36 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
                     "/m /v:normal /p:\"Configuration\"=\"Release\" /target:Build \"/Working/src/Solution.sln\"");
             }
 
+            [Theory]
+            [InlineData(PlatformTarget.MSIL, "/m /v:normal /p:\"Platform\"=\"AnyCPU\" /target:Build \"/Working/src/Solution.sln\"")]
+            [InlineData(PlatformTarget.x86, "/m /v:normal /p:\"Platform\"=\"x86\" /target:Build \"/Working/src/Solution.sln\"")]
+            [InlineData(PlatformTarget.x64, "/m /v:normal /p:\"Platform\"=\"x64\" /target:Build \"/Working/src/Solution.sln\"")]
+            public void Should_Append_Platform_As_Property_To_Process_Arguments(PlatformTarget? platform, string argumentsString)
+            {
+                // Given
+                var fixture = new MSBuildRunnerFixture(false, true);
+                fixture.Settings.SetPlatformTarget(platform.Value);
+
+                // When
+                fixture.Run();
+
+                // Then
+                fixture.AssertReceivedArguments(argumentsString);
+            }
+
+            [Fact]
+            public void Should_Omit_Platform_Property_In_Process_Arguments_If_It_Is_Null()
+            {
+                // Given
+                var fixture = new MSBuildRunnerFixture(false, true);
+                
+                // When
+                fixture.Run();
+
+                // Then
+                fixture.AssertReceivedArguments("/m /v:normal /target:Build \"/Working/src/Solution.sln\"");
+            }
+
             [Fact]
             public void Should_Set_Working_Directory()
             {
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs
@@ -18,14 +18,14 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
                 Assert.Equal(MSBuildToolVersion.Default, settings.ToolVersion);
             }
 
-            [Fact]
-            public void Should_Set_Default_Platform_Target_To_MSIL()
+            [Fact()]
+            public void Should_Set_Default_Platform_Target_To_Null()
             {
                 // Given, When
                 var settings = new MSBuildSettings();
 
                 // Then
-                Assert.Equal(PlatformTarget.MSIL, settings.PlatformTarget);
+                Assert.Null(settings.PlatformTarget);
             }
 
             [Fact]
@@ -81,6 +81,19 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
 
                 // Then
                 Assert.Equal(string.Empty, settings.Configuration);
+            }
+        }
+
+        public sealed class ThePlatformProperty
+        {
+            [Fact]
+            public void Should_Be_Null_By_Default()
+            {
+                // Given, When
+                var settings = new MSBuildSettings();
+
+                // Then
+                Assert.Null(settings.PlatformTarget);
             }
         }
 
--- a/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
+++ b/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Globalization;
 using System.Linq;
 using Cake.Core;
@@ -64,6 +65,13 @@ namespace Cake.Common.Tools.MSBuild
                 builder.Append(string.Concat("/p:\"Configuration\"=", configuration.Quote()));
             }
 
+            // Buid for a specific platform?
+            if (settings.PlatformTarget.HasValue)
+            {
+                var platform = settings.PlatformTarget.Value;
+                builder.Append(string.Concat("/p:\"Platform\"=", GetPlatformName(platform).Quote()));
+            }
+
             // Got any properties?
             if (settings.Properties.Count > 0)
             {
@@ -89,6 +97,21 @@ namespace Cake.Common.Tools.MSBuild
             builder.AppendQuoted(solution.MakeAbsolute(_environment).FullPath);
 
             return builder;
+        }
+
+        private static string GetPlatformName(PlatformTarget platform)
+        {
+            switch (platform)
+            {
+                case PlatformTarget.MSIL:
+                    return "AnyCPU";
+                case PlatformTarget.x86:
+                    return "x86";
+                case PlatformTarget.x64:
+                    return "x64";
+                default:
+                    throw new InvalidEnumArgumentException("platform", (int)platform, typeof(PlatformTarget));
+            }
         }
 
         private static string GetVerbosityName(Verbosity verbosity)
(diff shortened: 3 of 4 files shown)


src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -91,6 +98,17 @@ private ProcessArgumentBuilder GetArguments(FilePath solution, MSBuildSettings s
             return builder;
         }
 
+        private static string GetPlatformName(PlatformTarget platform)
+        {
+            switch (platform)
+            {
+                case PlatformTarget.MSIL:
+                    return "AnyCPU";                                    
+                default:
```
patriksvensson: I think it would be good to handle the different cases explicitly and throw a `InvalidOperationException` here.

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -64,6 +64,13 @@ private ProcessArgumentBuilder GetArguments(FilePath solution, MSBuildSettings s
                 builder.Append(string.Concat("/p:\"Configuration\"=", configuration.Quote()));
             }
 
+            // Buid for a specific platform?
+            if (settings.PlatformTarget != default(PlatformTarget))
```
patriksvensson: Would be better if we made `PlatformTarget` nullable and only set the platform target if it's not null. This would be more in line with how we do it at other places.

patriksvensson:
@os-alan Thanks for this! Not sure how `PlatformTarget` never was properly implemented but good that you found it! :+1: 

It looks good to me, but I left some minor comments that I think should be addressed before merging.

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
os-alan: I had same though originally, but explicitly specifying "AnyCPU" platform causes many other tests to fail (because before this change, AnyCPU was expected as default value) and I was too lazy to fix them. But, if you find this to be good idea as well, then it worth implementing :)

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
os-alan: You think `InvalidOperartionException` would suitable here? Maybe `ArgumentOutOfRangeException` ?

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
patriksvensson: Ok, I understand what you mean.

The platform target is used to chose what version of MSBuild to be used, and this conflicts.
Since we used `MSIL` as the default platform target, tests are failing.

We could make the PlatformTarget nullable and replace the `GetAlternativeToolPaths` [MSBuildRunner.cs](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs#L146).

```csharp
protected override IEnumerable<FilePath> GetAlternativeToolPaths(MSBuildSettings settings)
{
    if (settings == null)
    {
    }

    var platform = settings.PlatformTarget ?? PlatformTarget.MSIL;
    var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion, platform);
    if (path != null)
    {
        return new[] { path };
    }

    return Enumerable.Empty<FilePath>();
}
```

We would also need to remove the assignment of the default PlatformTarget in the [MSBuildSettings constructor](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs#L80).

Finally, we should change (and rename) the [test for the default values in MSBuildSettings](https://github.com/cake-build/cake/blob/develop/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs#L22) to reflect these changes:

```csharp
[Fact]
public void Should_Set_Default_Platform_Target_To_Null()
{
    // Given, When
    var settings = new MSBuildSettings();

    // Then
    Assert.Null(settings.PlatformTarget);
}
```

What do you think?

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
```diff
@@ -335,6 +335,27 @@ public void Should_Append_Configuration_As_Property_To_Process_Arguments()
                     "/m /v:normal /p:\"Configuration\"=\"Release\" /target:Build \"/Working/src/Solution.sln\"");
             }
 
+            [Theory]
+            [InlineData(null, "/m /v:normal /target:Build \"/Working/src/Solution.sln\"")]
```
patriksvensson: Create a new tests for this to avoid if statements in the test.

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs
```diff
@@ -18,16 +18,17 @@ public void Should_Set_Default_Tools_Version_To_Default()
...
-            public void Should_Set_Default_Platform_Target_To_MSIL()
+            [Fact()]
+            public void Should_Set_Default_Platform_Target_To_Null()
             {
                 // Given, When
                 var settings = new MSBuildSettings();
 
                 // Then
-                Assert.Equal(PlatformTarget.MSIL, settings.PlatformTarget);
+                Assert.Equal(default(PlatformTarget?), settings.PlatformTarget);
```
patriksvensson: I think `Assert.Null(settings.PlatformTarget)` would be more appropriate here.

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs
```diff
@@ -84,6 +85,20 @@ public void Should_Be_Empty_By_Default()
...
+        public sealed class ThePlatformProperty
+        {
+            [Fact]
+            public void Should_Be_Null_By_Default()
+            {
+                // Given, When
+                var settings = new MSBuildSettings();
+
+                // Then
+                Assert.Equal(default(PlatformTarget?), settings.PlatformTarget);
```
patriksvensson: I think `Assert.Null(settings.PlatformTarget)` would be more appropriate here.

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
os-alan: That's exactly what I did :+1:

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -150,7 +172,8 @@ protected override IEnumerable<FilePath> GetAlternativeToolPaths(MSBuildSettings
                 throw new ArgumentNullException("settings");
             }
 
-            var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion, settings.PlatformTarget);
+            var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion,
+                                                       settings.PlatformTarget.HasValue ? settings.PlatformTarget.Value : PlatformTarget.MSIL);
```
patriksvensson: Move this to it's own line for brevity like: `var platform = settings.PlatformTarget ?? PlatformTarget.MSIL`

patriksvensson:
Added some comments again :smile:

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
patriksvensson: I don't think that `ArgumentOutOfRangeException` fits well for this, since it's not an array type. Otherwise you could use `InvalidEnumArgumentException`.

os-alan:
Fixed again :-)

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
```diff
@@ -335,6 +335,37 @@ public void Should_Append_Configuration_As_Property_To_Process_Arguments()
...
+            [Theory]
+            [InlineData(PlatformTarget.MSIL, "/m /v:normal /p:\"Platform\"=\"AnyCPU\" /target:Build \"/Working/src/Solution.sln\"")]
+            [InlineData(PlatformTarget.x86, "/m /v:normal /p:\"Platform\"=\"x86\" /target:Build \"/Working/src/Solution.sln\"")]
+            [InlineData(PlatformTarget.x64, "/m /v:normal /p:\"Platform\"=\"x64\" /target:Build \"/Working/src/Solution.sln\"")]
+            public void Should_Append_Platform_As_Property_To_Process_Arguments(PlatformTarget? platform, string argumentsString)
+            {
+                // Given
+                var fixture = new MSBuildRunnerFixture(false, true);
+                
+                if (platform.HasValue)
```
patriksvensson: I think this if-statement should be omitted.

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
```diff
@@ -335,6 +335,37 @@ public void Should_Append_Configuration_As_Property_To_Process_Arguments()
...
+                // Then
+                fixture.AssertReceivedArguments(argumentsString);
+            }
+
+            [Fact]
+            public void Should_Omit_Platform_Property_In_Process_Arguments_If_It_Is_Null()
+            {
+                // Given
+                var fixture = new MSBuildRunnerFixture(false, true);
+                // When
```
patriksvensson: Line break

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs
```diff
@@ -84,6 +85,20 @@ public void Should_Be_Empty_By_Default()
...
+            {
+                // Given, When
+                var settings = new MSBuildSettings();
+
+                // Then
+                Assert.Null(settings.PlatformTarget);
+            }
+        }
+
+
```
patriksvensson: Remove extra line break

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsTests.cs
```diff
@@ -18,16 +18,17 @@ public void Should_Set_Default_Tools_Version_To_Default()
...
             {
                 // Given, When
                 var settings = new MSBuildSettings();
 
                 // Then
-                Assert.Equal(PlatformTarget.MSIL, settings.PlatformTarget);
+                Assert.Null(settings.PlatformTarget);
             }
 
+
```
patriksvensson: Remove extra line break

patriksvensson:
@os-alan Looks good to me. Some final minor comments and we're good to merge.

Would also be nice if you could rebase your work to one commit with a commit message that's more in line with others such as "Fix for PlatformTarget not used in MSBuild runner" or similar.

Great work! :+1:

os-alan:
Hi Patrik! 

>Would also be nice if you could rebase your work to one commit with a commit message that's more in line with others such as "Fix for PlatformTarget not used in MSBuild runner" or similar.

Fixed formatting and rebased. Hope, it will pass now :sweat_smile:

patriksvensson:
Hello @os-alan!

Yes, everything looks good now and I will merge this later tonight! :+1: 

/ Patrik

patriksvensson:
Merged :+1:

PR #336, opened 2015-08-08
Changed TeamCityOutput to nullable boolean.

### LINKED ISSUE #320: Better Design around Fixie's TeamCity output
When the `--TeamCity` option is specified, Fixie will use that value to produce TeamCity styled formatting in its output. When the value is omitted, Fixie checks for an environment variable. This part isn't of any concern to Cake, but this should be better documented in the `FixieRunner` usage.

Also, for better discoverability and ease of use, the `TeamCityOutput` enum could be simplified to `Nullable<bool>`.

patriksvensson:
When the `--TeamCity` option is specified, Fixie will use that
value to produce TeamCity styled formatting in its output. When the
value is omitted, Fixie checks for an environment variable.

Also, for better discoverability and ease of use, the
`TeamCityOutput` enum has been changed to a nullable bool.

Resolves #320


ChrisMissal:
:+1:

PR #339, opened 2015-08-09
Added using-alias-directive support.

### LINKED ISSUE #328: Using alias directives not supported
To solve conflicting type names C# supports using aliases i.e:
`using ClassAlias = N1.N2.Class;`
[MSDN reference](https://msdn.microsoft.com/en-us/library/aa664765(v=vs.71).aspx)

Currently this gets treated as an namespace in Cake and fails.

Haven't looked into details yet if this is doable or not.

patriksvensson:
Seems to work correctly with Roslyn, but on Mono there seem to
be a bug in the Evaluator that throws a NullReferenceException.
Since the exception is so vague, I added a better error message
for the user.

I've filed a bug report with Xamarin for this that can
be found at https://bugzilla.xamarin.com/show_bug.cgi?id=32894#c0

Resolves #328


patriksvensson:
Merged

PR #343, opened 2015-08-11
Extend nunit switches - Closes #330

### LINKED ISSUE #330: NUnit: support for /domain, /process, /apartment options
I noticed that the NUnit feature doesn't currently support the `/domain` and `/process` options.

Would anyone object to me taking a stab at adding them?

Here are their descriptions, from the [NUnit command line docs](http://www.nunit.org/index.php?p=consoleCommandLine&r=2.6.4):

> ### Controlling the Use of Processes

> The **/process** option controls how NUnit loads tests in processes. The
following values are recognized.

> - **Single**     All the tests are run in the nunit-console process. This is the default. 
> - **Separate**     A separate process is created to run the tests. 
> - **Multiple**     A separate process is created for each test assembly, whether specified on the command line or listed in an NUnit project file. 

> **Note:** This option is not available using the .NET 1.1 build of nunit-console.

> ### Controlling the Use of AppDomains

> The **/domain** option controls of the creation of AppDomains for running
tests. The following values are recognized:

> - **None**     No domain is created - the tests are run in the primary domain. This normally requires copying the **NUnit** assemblies into the same directory as your tests. 
> - **Single**     A test domain is created - this is how NUnit worked prior to version 2.4 
> - **Multiple**     A separate test domain is created for each assembly 

> The default is to use multiple domains if multiple assemblies are listed on
the command line. Otherwise a single domain is used.

> ### Controlling the Apartment in Which Tests Run

> The **/apartment** option may be used to specify the ApartmentState (STA or
MTA) of the test runner thread. Since the default is MTA, the option is only
needed to force execution in the Single Threaded Apartment.

> **Note:** If a given test must **always** run in a particular apartment, as is the case with many Gui tests, you should use an attribute on the test rather than specifying this option at the command line.

jrnail23:


patriksvensson:
@jrnail23 Could you rebase this against develop to a single commit so I can merge this? Would appreciate it! Thanks!

jrnail23:
@patriksvensson, I just got back from vacation, sorry to take so long to respond.  I'll work on this today.

jrnail23:
@patriksvensson, I think this is ready to merge now.

devlead:
@jrnail23 this is now merged, thanks for you contribution :+1:

PR #344, opened 2015-08-11
Closes #342: Killable processes

### LINKED ISSUE #342: Can we make IProcess disposable or give it a stop method, etc.?
Last night I sat down to write a Cake add-on for IIS Express.
The use case I'm trying to enable is that in one build task I can start one or more IIS express instances (as new processes), then in subsequent tasks, I would run acceptance tests, then I'd kill the IIS Express process in a `Teardown` block.

To implement my add-in, I figured I'd use `IProcessRunner` to start my IIS Express processes, returning the process to be managed by the consuming script (to be stopped, disposed, etc.).  Then I realized that the existing `IProcess` interface doesn't actually expose any way to stop or dispose a process.

First of all, is there a better way to achieve this? And second, is it feasible to extend `IProcess` to enable this behavior? I realize that would be a breaking change, so it wouldn't be trivial to manage.

jrnail23:


patriksvensson:
Should resolve #342

patriksvensson:
Looks good to me :+1:

patriksvensson:
@jrnail23 Could you rebase this against develop to a single commit so I can merge this? Thanks!

jrnail23:
@patriksvensson, I'll give this a try today.

jrnail23:
@patriksvensson, is this suitable now?  I'll await your feedback before taking the same approach to PR #343

jrnail23:
@patriksvensson, I believe this is ready to be merged now.
FYI, I have some other `IProcess`-related enhancements that I've been working on locally. I'll create a new issue to propose them and get feedback before submitting any code.

devlead:
@jrnail23 this is now merged, thanks for your contribution :+1:  You introduced extra whitespace in all `<summary>` xml documentation, this introduced allot of extra diffs, for future PR's if possible try to only change lines needed. Makes it allot easier for us to see what's changed/relevant for each PR.

jrnail23:
@devlead, sorry about that, a stupid oversight on my part.  I frequently use ReSharper silent cleanup, which I figured would be OK for Cake since you guys have a settings file in the solution.
I may have some user-specific or machine-wide settings that override the Cake solution settings.
In any case, I know those extraneous changes are a real pain in the ass.  Won't happen again.

devlead:
@jrnail23  no worries.

jrnail23:
@devlead, I found the resharper settings on my machine that were overriding the defaults expected by Cake Team-Shared settings (which had no specific settings in place for reformatting XML comments).
Would you like me to specify the settings that would enforce the current XML comment formatting conventions to prevent stuff like this from happening again? ("No indent at all" inside elements would match the convention, or "Do not change indentation" would just prevent any automated indentation changes.

devlead:
@jrnail23 likely wouldn't do any harm. So sure go ahead. Personally I leave off any auto reformatting just to reduce noise in feature commits. But then I'm more sensitive than most when it comes to white space ;)

PR #346, opened 2015-08-11
Add GitVersion alias.

RichiCoder1:
Also reworked process and process arguments.

### DIFF
--- a/src/Cake.Common/Tools/GitVersion/GitVersionAliases.cs
+++ b/src/Cake.Common/Tools/GitVersion/GitVersionAliases.cs
@@ -1,0 +1,115 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.Annotations;
+
+namespace Cake.Common.Tools.GitVersion
+{
+    /// <summary>
+    /// Contains functionality related to GitVersion.
+    /// <see href="http://gitversion.readthedocs.org/en/latest/">GitVersion Documentation</see>
+    /// <see href="https://www.nuget.org/packages/GitVersion.CommandLine/">GitVersion NuGet Package</see>
+    /// </summary>
+    [CakeAliasCategory("GitVersion")]
+    public static class GitVersionAliases
+    {
+        /// <summary>
+        /// Retrives the GitVersion output.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <returns>The git version info.</returns>
+        /// <example>
+        /// <para>Update the assembly info files for the project.</para>
+        /// <para>Cake task:</para>
+        /// <code>
+        /// <![CDATA[
+        /// Task("UpdateAssemblyInfo")
+        ///     .Does(() =>
+        /// {
+        ///     GitVersion(new GitVersionSettings {
+        ///         UpdateAssemblyInfo = true
+        ///     });
+        /// });
+        /// ]]>
+        /// </code>
+        /// <para>Get the git version info for the project using a dynamic repository.</para>
+        /// <para>Cake task:</para>
+        /// <code>
+        /// <![CDATA[
+        /// Task("GetVersionInfo")
+        ///     .Does(() =>
+        /// {
+        ///     var result = GitVersion(new GitVersionSettings {
+        ///         UserName = "MyUser",
+        ///         Password = "MyPassword,
+        ///         Url = "http://git.myhost.com/myproject.git"
+        ///         Branch = "develop"
+        ///         Commit = EnviromentVariable("MY_COMMIT")
+        ///     });
+        ///     // Use result for building nuget packages, setting build server version, etc...
+        /// });
+        /// ]]>
+        /// </code>
+        /// </example>
+        [CakeMethodAlias]
+        public static GitVersion GitVersion(this ICakeContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            return GitVersion(context, new GitVersionSettings());
+        }
+
+        /// <summary>
+        /// Retrives the GitVersion output.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="settings">The GitVersion settings.</param>
+        /// <returns>The git version info.</returns>
+        /// <example>
+        /// <para>Update the assembly info files for the project.</para>
+        /// <para>Cake task:</para>
+        /// <code>
+        /// <![CDATA[
+        /// Task("UpdateAssemblyInfo")
+        ///     .Does(() =>
+        /// {
+        ///     GitVersion(new GitVersionSettings {
+        ///         UpdateAssemblyInfo = true
+        ///     });
+        /// });
+        /// ]]>
+        /// </code>
+        /// <para>Get the git version info for the project using a dynamic repository.</para>
+        /// <para>Cake task:</para>
+        /// <code>
+        /// <![CDATA[
+        /// Task("GetVersionInfo")
+        ///     .Does(() =>
+        /// {
+        ///     var result = GitVersion(new GitVersionSettings {
+        ///         UserName = "MyUser",
+        ///         Password = "MyPassword,
+        ///         Url = "http://git.myhost.com/myproject.git"
+        ///         Branch = "develop"
+        ///         Commit = EnviromentVariable("MY_COMMIT")
+        ///     });
+        ///     // Use result for building nuget packages, setting build server version, etc...
+        /// });
+        /// ]]>
+        /// </code>
+        /// </example>
+        [CakeMethodAlias]
+        public static GitVersion GitVersion(this ICakeContext context, GitVersionSettings settings)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            var gitVersionRunner = new GitVersionRunner(context.FileSystem, context.Environment, context.ProcessRunner, context.Globber, context.Log);
+            return gitVersionRunner.Run(settings);
+        }
+    }
+}
--- a/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
+++ b/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
@@ -1,0 +1,172 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.Serialization.Json;
+using System.Text;
+using Cake.Core;
+using Cake.Core.Diagnostics;
+using Cake.Core.IO;
+using Cake.Core.Tooling;
+
+namespace Cake.Common.Tools.GitVersion
+{
+    /// <summary>
+    /// The GitVersion runner.
+    /// </summary>
+    public sealed class GitVersionRunner : Tool<GitVersionSettings>
+    {
+        private ICakeLog _log;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GitVersionRunner"/> class.
+        /// </summary>
+        /// <param name="fileSystem">The file system.</param>
+        /// <param name="environment">The environment.</param>
+        /// <param name="processRunner">The process runner.</param>
+        /// <param name="globber">The globber.</param>
+        /// <param name="log">The log.</param>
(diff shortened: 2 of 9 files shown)


gep13:
Woot! It is great to see this getting added! @JakeGinnivan

Am I right in saying that I can call this alias, and get back the json output from GitVersion for later use in my build script?

RichiCoder1:
It parses the JSON output, but yup!

gep13:
Sweet! And this addition was using the latest v3 version of GitVersion?  Is it possible to have cake understand the difference between 2 versions of the same tool?

RichiCoder1:
I just used what was ever latest to be honest. >_>

gep13:
>@RichiCoder1 said...
I just used what was ever latest to be honest. >_>

Makes sense!

@patriksvensson is there a way to have Cake support multiple versions of a tool?  Say the input arguments change, get removed, etc.

RichiCoder1:
You could theoretically have an external package that has something like `Cake.GitVersionV2.dll` `Cake.GitVersionV3.dll`. I don't think there's anything in cake that supports this out of the box.

patriksvensson:
Question is, does it differ that much between versions? In the case of xUnit and xUnit v2 we've added a XUnit2 runner since there is both semantic and breaking changes in their API (and both are widely used), but as long as there's only additions I would say that it's up to the user not to use functionality not supported by the version they're using.

If we really must support different versions (a lot of them), I would actually suggest creating NuGet packages for the deprecated ones.

Whatbdo you think?

gep13:
GitVersion v2 to GitVersion v3 was essentially a complete rewrite, and various new arguments were added onto the command line implementation.  Going forward, there hopefully won't be that many changes, but in this one, there was quite a bit.  I was just curious as to how this was handled.  I think for now, the suggestion would be to upgrade GitVersion to the latest, and not support the older one.

patriksvensson:
The plan from the beginning was to support general tool that's widely used and seldom change such as MSBuild.

To be honest there is no modus operandi in place for how to handle this, but I'm open for suggestions.

JakeGinnivan:
I think not supporting v2 is a good idea. The current plan for v4 is to give the command line interface an overhaul, but that will just be switches not so much functionality. We might be able to have a compat shim which we ship to make it easy for people to transition. If we provide a shim it would just be detecting if the shim is there, then calling it instead. We also might be able to detect legacy command line args and translate transparently.

If we don't add any of those compatibility options then based on the major version you are running against you would use different command line switches. If cake could read the major version of the exe then generate the appropriate command line switches then it could be the same extension.

This is pretty awesome! :+1: We should definitely link and add some notes to the GitVersion docs to make this as discoverable as possible

src/Cake.Common/Tools/GitVersion/GitVersionSettings.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <summary>
+        /// Gets or sets the path to the log file.
+        /// </summary>
+        public FilePath LogFilePath { get; set; }
+    }
+
+    /// <summary>
+    /// The git version output type.
+    /// </summary>
+    public enum GitVersionOutput
```
patriksvensson: Move this to a separate file.

src/Cake.Common/Tools/NuGet/SetApiKey/NuGetSetApiKey.cs
```diff
@@ -55,13 +55,15 @@ public void SetApiKey(string apiKey, string source, NuGetSetApiKeySettings setti
...
-            string output = null;
+
             var processSettings = new ProcessSettings
             {
                 Arguments = GetArguments(apiKey, source, settings),
                 RedirectStandardOutput = true
             };
-            Run(settings, null, settings.ToolPath, processSettings, process => output = string.Join("\r\n", process.GetStandardOutput()));
+
+            var process = Run(settings, null, settings.ToolPath, processSettings);
```
patriksvensson: Why those changes? Doesn't seem related to GitVersion.

src/Cake.Common/Tools/NuGet/SetProxy/NuGetSetProxy.cs
```diff
@@ -54,13 +54,14 @@ public void SetProxy(string url, string username, string password, NuGetSetProxy
                 throw new ArgumentNullException("settings");
             }
 
-            string output = null;
             var processSettings = new ProcessSettings
             {
                 Arguments = GetArguments(url, username, password, settings),
                 RedirectStandardOutput = true
             };
-            Run(settings, null, settings.ToolPath, processSettings, process => output = string.Join("\r\n", process.GetStandardOutput()));
```
patriksvensson: Why those changes? Doesn't seem related to GitVersion.

src/Cake.Common/Tools/NuGet/Sources/NuGetSources.cs
```diff
@@ -133,9 +133,8 @@ public bool HasSource(string source, NuGetSourcesSettings settings)
                 RedirectStandardOutput = true
             };
 
-            var result = false;
```
patriksvensson: Why those changes? Doesn't seem related to GitVersion.

patriksvensson:
I think the GitVersion functionality is good, but the PR does too much.

* GitVersion support
* Adds a SwitchArgument
* Returns the process (already solved in #344)
* Defines a new method for `IProcess`.

I think it would be better if the PR were more focused around the GitVersion support and make the other tweaks as separate PR:s (unless I'm mistaken and the functionality is required for the GitVersion tool).

patriksvensson:
I spoke with @RichiCoder1 in the [Gitter room](https://gitter.im/cake-build/cake), and the plan is to postpone merging this PR until #344 has been merged.

src/Cake.Common/Tools/GitVersion/GitVersionAliases.cs
```diff
@@ -0,0 +1,111 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.Annotations;
+
+namespace Cake.Common.Tools.GitVersion
+{
+    /// <summary>
+    /// Contains funcionality related to GitVersion.
+    /// </summary>
+    [CakeAliasCategory("GitVersion")]
```
gep13: @RichiCoder1  When you come back around to this, would it be possible to change to to be GitTools, and then add an additional alias category to the GitVersion call below?  Similar to what is done here in the ReSharper tools? 

https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/DupFinder/DupFinderAliases.cs

Eventually, I would like to add aliases for all the GitTools, including GitReleaseManager, GitReleaseNotes, as well as GitVersion.  Assuming that @patriksvensson and @devlead  is ok with that :smile_cat:

gep13:
@patriksvensson  @devlead I have started taking a look at this, and I think I have fixed up all the main concerns that there were in @RichiCoder1 original PR.  This PR now only adds GitVersion alias, and nothing else.

I am nothing into one issue though...

When trying to test with either this:

```
Task("GetVersionInfo")
    .Does(() => 
{
    var result = GitVersion(new GitVersionSettings {
        RepositoryPath = "c:/github/organisations/gittools/gitreleasemanager"
    });
    
    Console.WriteLine(result.ToString());
    Console.ReadLine();
});
```

or this:

```
Task("GetVersionInfo")
    .Does(() => 
{
    var result = GitVersion();
    
    Console.WriteLine(result.ToString());
    Console.ReadLine();
});
```

I am getting the following errors:

image

My eyes are starting to glaze over now, really quite sleepy, but if you can see anything obviously wrong, please let me know and I will fix up.

Plan on looking at this again tomorrow, cheers!

src/Cake.Common/Tools/GitVersion/GitVersionAliases.cs
gep13: @devlead with regard to the above comment that I made, I can't actually see how this would work anymore.  I thought I did, but now I am not so sure.  Is there a way to group all these tools under the GitTools alias, but still also group by both GitReleaseManager and GitVersion?

patriksvensson:
@gep13 It looks like the `GitVersion` alias is missing the [CakeMethodAlias](http://cakebuild.net/api/cake.core.annotations/8283ce29) attribute.

If the `GitVersionSettings` isn't in the same namespace as the alias, you can also add a [CakeNamespaceImport](http://cakebuild.net/api/cake.core.annotations/0a570436) attribute to tell Cake what extra namespaces to add. You can see an example of this [here](https://github.com/cake-build/cake/blob/5426a07c672ac2c81534d4cc2883ee7e59e7277f/src/Cake.Common/IO/FileAliases.cs#L34).

gep13:
@patriksvensson that was exactly the problem, thank you! :-)

Did you also see my comment on the line above?

gep13:
@patriksvensson @devlead @RichiCoder1 I "think" this is ready for review as well.  I have done some basic testing, and seems to work to me, interested in your thoughts.  Thanks!

patriksvensson:
@gep13 Not sure about what to do. If there is a need to support multiple categories for an alias on the website, we can add it.

gep13:
What I would love to see would be:

GitTools
  - GitVersion
  - GitReleaseManager
    - Create
    - Publish
    - ...

In the case of ReSharper, the nesting is simply

ReSharper
  - DupFinder
  - InspectCode

since there is no further aliases exposed.  So, you perhaps a change is needed to accommodate this.  It's just an idea, I am happy to leave it as:

GitVersion
GitReleaseManager
  - Create
  - Publish
  - ...

Would have just been nice to show that these some from the same place is all.

patriksvensson:
@gep13 I understand. So three levels instead of two? Might take a little bit of work to get that to work. Maybe we can leave it as it is and revisit this later when we know what we need to do.

gep13:
@patriksvensson oh, absolutely, there is no immediate need for this.  I would see this as a "nice to have" :-)

gep13:
@patriksvensson since this is not actually my PR, I can't change the title of the PR, any chance you can change it?  Thanks!

devlead:
@gep13 what's the status here?

gep13:
@devlead it is next on my list.  The PR was ready to go, but now needs to be refactored to use the new base classes that @patriksvensson created.  I forgot about this one the other day when I was refactoring, as I was filtering open PR's based on who started them, i.e. me, so this one got missed out.  Hopefully get to this today.

devlead:
Excellent @gep13 :+1:

gep13:
@devlead ok, assuming all goes well with the builds, and you are happy, I think this is ready.

src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            if (settings.OutputType != GitVersionOutput.BuildServer)
+            {
+                var jsonString = string.Empty;
+
+                Run(settings, GetArguments(settings), new ProcessSettings { RedirectStandardOutput = true },
+                process => jsonString = string.Join("\n", process.GetStandardOutput()));
+
+                var deserializer = new DataContractJsonSerializer(typeof(GitVersion));
+                return deserializer.ReadObject(new MemoryStream(Encoding.UTF8.GetBytes(jsonString))) as GitVersion;
```
devlead: Any special reason the `MemoryStream` isn't disposed/within a using statement here?

devlead:
@gep13 Looks good, just left one minor comment.

src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
gep13: none that I can think of. This was some of the original code from @RichiCoder1 that I hadn't really looked at.  Will refactor that just now, couple minutes...

devlead: Excellent :+1:

gep13:
@devlead pushed a new version, let me know if this is what you were thinking.

devlead:
@gep13 excellent thanks :+1:

Issue #348, opened 2015-08-12
Is RunTarget actually necessary?


olivier-spinelli:
Hello to every contributors of this great project!

I read the existing issues to look for an answer but I may missed it. IMHO, the fact that RunTarget is accessible from within the script is a weakness (calling RunTarget from everywhere except the very last step SHOULD nearly always result in a rather strange behavior... at best).

Shouldn't the "target" simply be handled at the host level? And the Task named "Default" be the default when no target name is provided?

RichiCoder1:
Correct me if I'm wrong @patriksvensson, but that's because Cake itself doesn't have a way of kicking off the "Target" when the script is done executing. You have to manually kick it off. Cake files are just C# scripts. All cake does is provide some pre-processing smarts and a global helper object that aliases can be merged with.

olivier-spinelli:
Yes, I have been thinking this.... before. But the fact is I can't see any way of using a build script without:
1 - specifying a target (exactly like I specify a script name): then there is no need to RunTarget from the script itself.
2 - not specifying a target: in such case the name "Default" is provided by the script itself as the default value to the target argument...

The fact is that CakeEngine.RunTarget needs a "target": The only benefit I can see is that the actual "Default" name is chosen by the script itself...
To conclude, i'd say that acting that the default entry point is "Default":
1 - Unifies the "interface" of a script.
2 - Removes these two lines (the first one and the last one of the current scripts).
3 - Removes the risk (from a beginner) of a call to RunTarget in the middle of the script.

It seems as simple as implementing that in CakeOptions/ArgumentParser: "Default" is automatically set when no "target" is specified on the arguments.

Did I miss something?

patriksvensson:
I agree. I've actually been thinking about renaming `RunTarget` to `SetTarget` and change the semantics for some time. Cake would then do the running automatically with 'Default' as target if no target was explicitly set.

RichiCoder1:
Would go nicely with my changes to make target a positional parameter, as we could just auto set the Target to "Default" if none is specified.

patriksvensson:
I have some code for this lying around so I can spike this out in  a PR if no one oppose it :)

RichiCoder1:
Nope. One more step towards the Just run `cake` dream :)

patriksvensson:
:D 

Or did @oliver-spinelli want to take this?

olivier-spinelli:
With pleasure!
May I propose : 
[...]

PR #352, opened 2015-08-13
SetTarget implementation (Issue #348).

olivier-spinelli:
Current Implementation is quite simple (ScriptHost.SetTarget sets the CakeEngine.Target property and the CakeEngine.RunTarget uses it).
Thanks to the default value of CakeEngine.RunTarget target parameter, this is a non breaking change: the parameter can now be null and the Target property is used. I updated the unit tests accordingly.

I did not change ScriptHost.RunTarget (I just marked it obsolete - that is useless for script). IMO it should call SetTarget and emit a warning in a first time and removed in a future release.


devlead:
Closing due inactivity.

gep13:
Relates to #348

PR #354, opened 2015-08-13
Referencing external nuget packages from your addin

SharpeRAD:
Tweaks the AddinProcessor to look for a "packages.config" file inside the addin directory then adds any referenced packages to cake. 

The great thing about this approach is you can copy your dll's "packages.config" file to the nuget package during your build script. Resolves an issue I have with my new addin [Cake.IIS (https://github.com/SharpeRAD/Cake.IIS) such that you only need to include one reference to pull in multiple nuget packages :smile:


patriksvensson:
@SharpeRAD I think there might be something wrong with the PR, since it references 31 commits. I also would like to see some tests that covers this functionality to be able to merge it since it affects a lot of functionality related to addins.

RichiCoder1:
Isn't this something that should be handled using good old fashion Nuget Dependencies?

PR #370, opened 2015-08-21
MSBuild Platform target was misspelled fixes #359

### LINKED ISSUE #359: MSBuild Platform target misspelled
The MSBuild Platform target produces the literal `AnyCPU` in [MSBuildRunner.cs](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs#L107) which should be `Any CPU` (notice the space).

**Workaround:**

Append a `WithProperty` as specified below, and remove the `SetPlatformTarget` call until this is fixed.

```csharp
MSBuild(solution, new MSBuildSettings()
    .WithProperty("Platform", "Any CPU")
    .SetConfiguration(configuration));
```

devlead:
This fixes an type that caused code like this to fail:
```csharp
        MSBuild(solution, settings =>
            settings.SetPlatformTarget(PlatformTarget.MSIL)
                .WithTarget("Build"));
```

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
@@ -336,7 +336,7 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
             }
 
             [Theory]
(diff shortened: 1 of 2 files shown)


patriksvensson:
Merged :+1:

PR #379, opened 2015-08-28
Extended ProcessAliases with methods returning IProcess

jrnail23:
added `StartNewProcess` alias methods to `ProcessAlias`.

Note: I'm not in love with the method name, but I was limited by the existing methods' names.
The existing ones, named `StartProcess` don't just start the process, they really execute the process and report the exit code.
My new ones do actually just start the process, leaving the user to interact with the resulting process in the manner of their choosing.  So in a perfect world where we didn't have to worry about breaking changes, my new methods would be named `StartProcess`, and the old ones would be renamed to `ExecuteCommand` or `ExecuteProcess`.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/ProcessFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/ProcessFixture.cs
@@ -31,9 +31,19 @@ namespace Cake.Common.Tests.Fixtures
             return Context.StartProcess(filename);
         }
 
+        public IProcess StartNewProcess(string filename)
+        {
+            return Context.StartAndReturnProcess(filename);
+        }
+
         public int Start(string filename, ProcessSettings settings)
         {
             return Context.StartProcess(filename, settings);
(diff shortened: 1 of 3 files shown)


jrnail23:
(need to clean up commits)

patriksvensson:
Hmm. Maybe we could call it `StartAndReturnProcess` or something like that?
What do you think?

jrnail23:
That's certainly more intention-revealing.  Works for me.

jrnail23:
@patriksvensson, I just updated this with your suggested name change.
Feel free to merge when/if you like.

patriksvensson:
@jrnail23 Would you mind rebasing this against develop into a single commit? Thanks! :smile:

jrnail23:
@patriksvensson -- all done, ready to go at your convenience

devlead:
@jrnail23 merged, thanks :+1:

PR #382, opened 2015-08-31
[WIP] Support more functionality from System.Diagnostics.Process in IProcess

jrnail23:
-- Exited event
-- ErrorDataReceived event
-- OutputDataReceived event
-- Redirect standard error
-- HasExited property
-- Environment variables for process
-- Start method on IProcess


jrnail23:
(need to clean up commits)

jrnail23:
FYI, I've got a bunch of integration/functional tests for this (local only), since the ProcessRunner and ProcessWrapper are pretty much untestable using tradtional unit testing (due to direct usage of System.Diagnostics.Process) but I'm pretty sure we're not going to want to add them as-is.  I'll follow up with more details on what I've done, and maybe someone will have a good idea of how to deal with them.

jrnail23:
(still tweaking this as I discover issues in exploratory testing)

patriksvensson:
Sorry for the delay.

While I appreciate your contribution have some opinions about this PR.
Please take this as constructive criticism.

Cake's process wrapper was never meant as a full fledged Process implementation.
It's there to provide an easy way to unit test logic that executes processes.
More advanced scenarios in a Cake script would probably rather benefit from
using the `System.Diagnostics.Process` class. If this functionality is
supposed to be used from an addin, we could expose this more advanced
functionality in some other way. Maybe an "advanced" adapter to an
existing `IProcess` or exposing the underlying `System.Diagnostics.Process`.

Personally I don't think that we should expose events from the
`IProcess` interface. We've used `Func`/`Action` delegates for callbacks
before so I think that would follow Cake's code standard more. It would probably
be more straigh forward in a Cake script to use.

My suggestions:

* Replace the Exited event with a `WaitHandle`.
* Replace output events with a single delegate registration.

jrnail23:
@patriksvensson thanks for the constructive feedback.
My intent here was indeed to use this in an add-in I'm working on.
I'll come back to this as soon as I get to a stopping point with the TaskSetup/Teardown stuff.

patriksvensson:
Yes, do that :smile: 
I would like to find a solution that works for you and others who're writing addins.

jrnail23:
@patriksvensson, just getting back to your suggestions now...

First of all, to confirm my understanding of what you're saying: 
* You'd prefer to see this functionality as a separate tool/alias within Cake.Core.
* You'd prefer to use `Func`/`Action` for callbacks instead of .net events/handlers.
* You want to combine the output / error events into a single delegate (presumably with an `IsError` flag in the callback arg)

does that all sound correct?

I'm not sure what you mean by 
> Replace the Exited event with a `WaitHandle`

Can you elaborate?  I assume you mean simply setting up an action callback to be invoked instead of the event handler.

jrnail23:
@patriksvensson, pinging you for feedback here -- I'm sure the little one has you pretty busy, just wanted to see which direction I should take here.

patriksvensson:
@jrnail23 Sorry for not responding to this sooner!

My suggestions were just suggestions :) I would prefer to keep IProcess as a higher level abstraction like you say, but I would like to find a way to solve your more advance use case as well. Maybe IProcess need to evolve into a wrapper around Process, I haven't ruled that out. What I would like is to somehow separate the advanced functionality and to keep testing of the process simple and to keep the interfaces as close to the rest of the Cake code as possible (i.e. no events).

With WaitHandle I meant something like a ManualResetEvent that we expose as a WaitHandle, but I'm not sure about this to be honest.

I appreciate all the work you've done with this and would like to find a solution were we keep the simplicity but still provides an option to do more advanced stuff like you want to do.

I'm writing this on my phone, so sorry if I repeat myself or sound ranty :)

jrnail23:
@patriksvensson, thanks again for the feedback.  
I just finished up a revision that removes the events in favor of action-based callbacks. I'll push it up to this PR so you can see what that looks like, and hopefully we can figure out the best way forward from there.

jrnail23:
@patriksvensson, @devlead: Hey guys, I'd like to figure out how to move forward with this, as it's blocking my other add-ins that I want to create for my actual work.
I completely understand if this approach doesn't feel right for Cake as a whole, so please just let me know if that's the case, and I may just re-create it as a separate addin.
Otherwise, I'd like to get it into a state where I can develop against it.

Are there any further modifications you'd like to see that would make you more comfortable including it in the core?

patriksvensson:
@jrnail23 Sorry for the delay (again), I simply forgot to follow up on this. Sorry about this.

To be honest, I don't know. I would like to find a solution to proceed with this without bloating `IProcess` too much. Perhaps is is possible to add this functionality as an addin somehow until we have figured this out since you depend on it for other stuff?

@devlead Do you have any suggestions here?

devlead:
Yeah, add-in sounds like a great starting point. Begin there ave see if any obstacles exist in core due you to do so.

jrnail23:
I just finished an initial cut at making the process stuff into an addin, then I updated my iis express addin to use it.
No issues with core, and it just works.
One little thing though -- maybe it would be a good idea to publish `Cake.Core.Testing` as a nuget package, to help out devs creating addins?

jrnail23:
Actually, maybe this didn't work as I hoped.  It looks like when referencing my iis express addin as an addin (`#addin`), Cake doesn't add a reference to my `Cake.Process` dll.  It does pull down the nuget package under my addins folder, it's just not referencing it from there.

patriksvensson:
@jrnail23 That should probably be supported. I'll add an issue for this and we can continue the discussion.

devlead:
@jrnail23 there's now a Cake.Testing #416 and addin imports references aliases  #415

So you should be able to progress with your addin.

devlead:
Closing due inactivity.

PR #388, opened 2015-09-02
Remove nuspec requirement if sufficent parameters supplied

devlead:
This will allow to pack Nuget packages without having a nuspec file as long as enough parameters supplied in the Cake file and nuspec directory exists making less friction to create packages and cleaner repos.

### DIFF
--- a/src/Cake.Common/Tools/NuGet/Pack/NuspecProcessor.cs
+++ b/src/Cake.Common/Tools/NuGet/Pack/NuspecProcessor.cs
@@ -19,6 +19,17 @@ namespace Cake.Common.Tools.NuGet.Pack
             _log = log;
         }
 
+        public FilePath Process(NuGetPackSettings settings)
+        {
+            var nuspecFilePath = settings.OutputDirectory
+                                    .CombineWithFilePath(string.Concat(settings.Id, ".nuspec"))
+                                    .MakeAbsolute(_environment);
+
+            var xml = LoadEmptyNuSpec();
+
+            return ProcessXml(nuspecFilePath, settings, xml);
+        }
+
         public FilePath Process(FilePath nuspecFilePath, NuGetPackSettings settings)
         {
             // Make the nuspec file path absolute.
@@ -36,6 +47,11 @@ namespace Cake.Common.Tools.NuGet.Pack
             _log.Debug("Parsing nuspec...");
             var xml = LoadNuspecXml(nuspecFile);
 
+            return ProcessXml(nuspecFilePath, settings, xml);
+        }
+
+        private FilePath ProcessXml(FilePath nuspecFilePath, NuGetPackSettings settings, XmlDocument xml)
+        {
             // Process the XML.
             _log.Debug("Transforming nuspec...");
             NuspecTransformer.Transform(xml, settings);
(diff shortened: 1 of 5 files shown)


ChrisMissal:
:+1:

patriksvensson:
So if the provided nuspec isn't found and a minimum amount of specific parameters are provided, it will create the NuGet package anyway? Maybe I'm missing something, but that doesn't sound right. What if you misspell the filename to the nuspec? I would want to get a "File not found" exception or something that halts execution of the build script so I could address the error.

Maybe I'm just misunderstanding something...

devlead:
@patriksvensson you are correct, I will adjust PR.

src/Cake.Common/Tools/NuGet/Pack/NuspecProcessor.cs
```diff
@@ -45,6 +62,24 @@ public FilePath Process(FilePath nuspecFilePath, NuGetPackSettings settings)
...
+            xml.LoadXml(@"<?xml version=""1.0"" encoding=""utf-8""?>
+<package xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema"">
+  <metadata xmlns=""http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd"">
+    <id></id>
+    <version>0.0.0</version>
+    <authors></authors>
+    <description></description>
+  </metadata>
+  <files>
+    <file src=""LICENSE"" />
```
patriksvensson: Is a `LICENSE` file always required?

src/Cake.Common/Tools/NuGet/Pack/NuspecProcessor.cs
```diff
@@ -24,17 +24,34 @@ public FilePath Process(FilePath nuspecFilePath, NuGetPackSettings settings)
...
+            XmlDocument xml;
+
+            // Make sure the nuspec file exist.
             if (!nuspecFile.Exists)
             {
-                const string format = "Could not find nuspec file '{0}'.";
-                throw new CakeException(string.Format(CultureInfo.InvariantCulture, format, nuspecFilePath.FullPath));
-            }
+                // Check sufficient parameters supplied voiding the need for nuspec template
+                if (
```
patriksvensson: Could we break this statement out to a method? Would be easier to read and follow.

patriksvensson:
Left some comments for you. Minor stuff.

src/Cake.Common/Tools/NuGet/Pack/NuspecProcessor.cs
devlead: no, files are replaced when specified(so wontr end up in nuspec that gets packed), but will remove.

src/Cake.Common/Tools/NuGet/Pack/NuspecProcessor.cs
devlead: Will take a look.

devlead:
Rethought and rewrote the PR so instead add an new `Pack(NuGetPackSettings settings)` alias leaving `Pack(FilePath nuspecFilePath, NuGetPackSettings settings)` working same as before, but still trying to reuse as much code as possible.

Required settings are
* OutputDirectory
* Id
* Version
* Authors
* Description
* Files

Example usage:
```csharp
NuGetPack(
    new NuGetPackSettings {
                OutputDirectory = "/Working/",
                Id = "nonexisting",
                Version = "1.0.0",
                Description = "The description",
                Authors = new[] { "Author #1", "Author #2" },
                Files = new NuSpecContent[] {
                    new NuSpecContent { Source = "LICENSE" }
                }
    }
);
```

PR #391, opened 2015-09-04
Changed NUnit argument prefix from '/' to '-'.

samvik:
This according to nunit documentation.
http://www.nunit.org/index.php?p=consoleCommandLine&r=2.6

Also renamed file with incorrect capitalization to be able to build on Linux.


patriksvensson:
@samvik Good catch! :+1: 

Since Windows support both hyphen and slash, maybe it would be better to simply change the argument to a hyphen?

samvik:
That would be the most clean solution. I'm all for that. Just wasn't sure if that would introduce any side effects on windows.

patriksvensson:
@samvik Good thinking! Just update the PR when you have time and I'll merge it :smile:

devlead:
Hi @samvik, please when you have the opportunity update PR and rebase against latest and we'll look at merging this.

samvik:
I should be up to date with develop now.

devlead:
@samvik merged! Thanks :+1:

PR #397, opened 2015-09-04
Added homebrew install paths to Cake tool resolver

Redth:
Currently if you install cake via homebrew and execute it using the global alias `cake` and then proceed to use `CakeExecuteScript (..)` it will fail to find `Cake.exe` to run.

This patch simply looks for `Cake.exe` in the paths that homebrew will install cake and sorts it so that the newest version is returned first.


devlead:
@Redth when you have the opportunity could you rebase this into 1 commit.

Redth:
Squashed. :)

devlead:
@Redth merged, thanks! :+1:

PR #399, opened 2015-09-04
Enhance exception message to include name of missing cake argument

jrnail23:
Just something that annoyed me the other day -- I forgot to pass in an argument and the resulting exception message `"Argument was not set."` wasn't very helpful.

Now it'll be `"Argument 'MY_ARGUMENT_NAME' was not set."`


devlead:
merged, thank you :+1:

PR #407, opened 2015-09-11
Add path to nuget resolver for mono on osx 10.11

Redth:
On OSX 10.11 (El Capitan) we can no longer have `/usr/bin/nuget` as `/usr/bin/` is not writeable.  The new location for this symlink will be `/usr/local/bin/nuget` so we should check for that.

### DIFF
--- a/src/Cake.Core/IO/NuGet/NuGetToolResolver.cs
+++ b/src/Cake.Core/IO/NuGet/NuGetToolResolver.cs
@@ -13,6 +13,8 @@ namespace Cake.Core.IO.NuGet
         private readonly IGlobber _globber;
         private IFile _cachedPath;
 
+        private static readonly FilePath[] unixSystemPaths = new[] { new FilePath("/usr/local/bin/nuget"), new FilePath("/usr/bin/nuget") };
+
         /// <summary>
         /// Initializes a new instance of the <see cref="NuGetToolResolver" /> class.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Redth:
I should clarify that this new symlink will be created by future mono installs (the next release of mono will have it).

devlead:
`usr/local` is fairly standard among posix system so seems reasonable, might there be more standard paths? just wondering as the code seems a bit duplicated now.
Would it make sense to have an read only array of paths to iterate over?
Or have something like an `bool TryCheckFilePathExists(FilePath testPath, out existsPath)` to reduce duplication?

Redth:
Well, in this case, the mono install will put it one of two places, and formerly it will be in /usr/bin, but from 10.11 on, it will be /usr/local/bin...

We could introduce a loop for a set of readonly paths if you'd prefer, just not sure there will be more than two ever, at least in the case of NuGet, which this resolver is already sort of a special case for.

devlead:
I get it, just had a gut reaction an feeling this part of the code was identical except for the path.
```csharp
var nugetUnixPath = new FilePath("/usr/local/bin/nuget"); 
if (_fileSystem.Exist(nugetUnixPath{ 
    _cachedPath = _fileSystem.GetFile(nugetUnixPath); 
    return _cachedPath.Path; 
} 
```

devlead:
I.e. basically same code here too in way.
https://github.com/Redth/cake/blob/patch-1/src/Cake.Core/IO/NuGet/NuGetToolResolver.cs#L110

Just thinking we perhaps could increase readability with a helper function testing path, only caching if valid, if so returning true and outputting the cached path.

devlead:
Don't want to be a dry _fundamentalist_;), just thinking out loud.

Redth:
I've turned it into a for loop which is a bit more readable.. Not sure it necessarily needs its own function for now...  I'm pretty motivated to see this make 0.5.4 though since it breaks things for me right now :)

devlead:
@Redth merged, thanks :+1:

PR #409, opened 2015-09-11
Added tool cache & changed to PS command

devlead:
* Added caching of tools to speed up builds
* Changed to use PowerShell command to fix AppVeyor layout issues


patriksvensson:
Great! Forgot about the packages.config in the tools folder when I added package caching. This will hopefully spead up the builds even more!

PR #411, opened 2015-09-13
Added verbose message when glob pattern did not match anything.

### LINKED ISSUE #408: NUnit runner does not throw if glob pattern matched nothing
The call to the NUnit runner with a glob pattern that will not find any files will try to execute NUnit without arguments. We should throw if this is the case.

```csharp
NUnit("this_matches_nothing/**/*.Tests.dll");
```

patriksvensson:
Sometimes when providing a glob pattern to certain aliases such
as NUnit or XUnit, a mistake in the pattern will result in
zero files being matched. This is kind of annoying since the
call to the alias will silently pass. This might be by design
but we should at least write a verbose message saying that no
files or directories were matched.

Closes #408


patriksvensson:
Just realized I need to add one more test due to a change.
Do not merge this yet.

patriksvensson:
Nevermind, this PR is ready to be merged.

PR #418, opened 2015-09-16
Resolves #414: additional xunit2 switches

### LINKED ISSUE #414: XUnit2: add support for additional command line switches
The following command line switches are among those missing from Cake's XUnit2 support:

- [x] `-parallel`
- [x] `-maxthreads`
- [x] `-trait`
- [x] `-notrait`

jrnail23:


src/Cake.Common/Tools/XUnit/XUnit2Settings.cs
```diff
@@ -7,6 +9,14 @@ namespace Cake.Common.Tools.XUnit
     /// </summary>
     public sealed class XUnit2Settings
     {
+        private readonly IDictionary<string, IList<string>> _traitsToInclude =
```
patriksvensson: Could be made a auto-property and initialized in the constructor.

src/Cake.Common/Tools/XUnit/XUnit2Settings.cs
```diff
@@ -7,6 +9,14 @@ namespace Cake.Common.Tools.XUnit
...
+            new Dictionary<string, IList<string>>(StringComparer.OrdinalIgnoreCase);
+
+        private readonly IDictionary<string, IList<string>> _traitsToExclude =
```
patriksvensson: Could be made a auto-property and initialized in the constructor.

src/Cake.Common/Tools/XUnit/XUnit2SettingsExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+    public static class XUnit2SettingsExtensions
+    {
+        /// <summary>
+        /// Adds a trait to the settings, to include in test execution.
+        /// </summary>
+        /// <param name="settings">The settings.</param>
+        /// <param name="name">The trait name.</param>
+        /// <param name="values">The trait values.</param>
+        /// <returns>The same <see cref="XUnit2Settings"/> instance so that multiple calls can be chained.</returns>
+        public static XUnit2Settings IncludingTrait(this XUnit2Settings settings, string name, params string[] values)
```
patriksvensson: The name (`IncludingTrait`) is good, but we might want to follow other Cake aliases and use `WithTrait` here.
What do you think?

patriksvensson:
@jrnail23 Looks good! Left some comments about some minor things. :+1:

src/Cake.Common/Tools/XUnit/XUnit2SettingsExtensions.cs
jrnail23: @patriksvensson, would the inverse of that (to correspond with the `TraitsToExclude` setting) be `WithoutTrait`?
Do you think that makes the builder more fluent, in terms of expressing intent to the user?

patriksvensson: @jrnail23 Not sure to be honest. In this case maybe `IncludeTrait` and `ExcludeTrait` would be better.

jrnail23: Yeah, I like that.

jrnail23:
@patriksvensson, I applied your suggestions, so this should be good to go now :smile:

devlead:
@jrnail23 Merged, thanks :+1:

PR #420, opened 2015-09-20
Added script analyzer.

patriksvensson:
To analyze script files outside of Cake, I've separated the script analysis and script processing. This so that the Omnisharp integration can analyse scripts without actually downloading NuGet packages and such.

To analyze a script, simply create an instance of the `ScriptAnalyzer` and provide it with the appropriate settings. Doing this will return a `ScriptAnalyzerResult` that contain a tree containing all referenced scripts and their references, tools, addins, using alias directives and namespaces.

If you want to download the tools and addins, simply instansiate the `ScriptProcessor` class that will accept a `ScriptAnalyzerResult`.

I've also added a lot of new tests to cover the functionality, and I've also manually tested the application to make sure that I haven't broken anything.

```csharp
// Analyze the script.
var analyzer = new ScriptAnalyzer(fileSystem, environment, log);
var result = analyzer.Analyze("./build.cake");

// Iterate through all scripts included in build.cake.
ScriptInformation script = result.Script;
Information("{0} requires {1} addin(s).", script.Path, script.Addins.Count);

// Iterate all scripts included by the script.
foreach(ScriptInformation included in script.Includes)
{
  Information("{0} directly references script {1}", script.Path, included.Path);
}

// Install all addins and tools.
var scriptProcessor = new ScriptProcessor(fileSystem, environment, log, installer);
var addinAssemblies = scriptProcessor.InstallAddins(result, "./Addins");
var toolFiles = scriptProcessor.InstallTools(result, "./Tools");
```


RichiCoder1:
Very awesome! I'll jump on this asap.

patriksvensson:
Updated usage example.

devlead:
@patriksvensson solid👍  get conflicts sorted and up-to-date with base branch.

patriksvensson:
Will do :smile:

patriksvensson:
This should be ready for merge once the CI builds finish doing their stuff :smile:

jrnail23:
@patriksvensson, what happens when we install a nuget package that has dependencies on other nuget packages?
The `#addin` directive will allow the analysis to include the primary package, but it won't include any transitive dependencies.
That said, my PR(#435) won't handle this yet either. The only way at current (that I know of) to identify a nuget package's dependencies is by actually downloading the package. I think nuget's v3 API may support such info via REST, but I assume we're not there yet.

patriksvensson:
@jrnail23 Yes, you're right, but I think this will suffice for now. Plugins (like the one for Omnisharp) would need to download the addin anyway to be able to reference the assemblies. Maybe we can add functionality to probe a dowloaded NuGet package depending on framework?

jrnail23:
@patriksvensson, I agree -- probing a downloaded nuget package seems to be the next logical step once #435 is in place.
So yeah, in terms of providing a "complete" analysis of the script (including transitive dependencies), we're probably out of luck, until the nugets actually get downloaded.
I only brought this up because I saw that one of your objectives in creating the script analyzer was to avoid having to download and install dependencies up front in order to get an analysis result.

Honestly though, if we were talking about a VS project instead of a script, wouldn't the same limitation apply?  With that in mind, I don't think this is a huge problem.

PR #422, opened 2015-09-22
Added main cake repo requirement for MyGet publish

devlead:
I wan't do be able to automatically build my fork of Cake too on AppVeyor. Currently it fails when trying to publish to MyGet because it lacks the proper ApiKey, obviously forks shouldn't publish to Cake MyGet feed.

So what I propose is that we check which repo is built and only if it's `cake-build/cake` should we publish to MyGet, this enables automatic builds of forks with the standard `build.cake`


patriksvensson:
Looks good to me :+1: 
Should I merge it or will you?

PR #429, opened 2015-10-01
Allow setting the AssemblyConfiguration attribute

michael-wolfenden:
This is useful for checking the configuration (eg Debug, Release) that an assembly was built using


patriksvensson:
Looks good for me. :+1: 
I will include this in the v0.5.5 release.

Issue #373, opened 2015-08-24
Issue with External nugets used directly via #addin directive


jrnail23:
I did some experimenting with using external nuget packages through the `#addin` directive, and I found that it only works when there is only a single dll file for each dll name in the referenced nupkg file.

Before I get into detail here, I should say that I assume what I'm trying to do here isn't really what the developer of Cake's `#addin` directive intended when writing it (there's really no documentation that I could find to explain the intended usage for the various ways to reference such dependencies).

I'm sure my initial explanation was about as clear as mud, so here's an example:

Let's say I want to use [Polly](https://www.nuget.org/packages/Polly) in my Cake script, so I do this:

```C#
#r "./tools/Polly/lib/net45/Polly.dll"
using Polly;

Task("Default")
	.Does(() => Policy.Handle<IOException>()
		.WaitAndRetry(10, i => TimeSpan.FromMilliseconds(50))
		.Execute(() => CleanDirectory(@"c:\MyFakeBinaries")));

RunTarget("Default");
```
Ok, so I run that, and it works.

However, if I want to use Polly as an addin like this (because I don't want to have to specify the path to Polly.dll for various reasons -- among them being that I might like to specify the tools directory at runtime), I'm out of luck:

```C#
#addin Polly
using Polly;

Task("Default")
	.Does(() => Policy.Handle<IOException>()
		.WaitAndRetry(10, i => TimeSpan.FromMilliseconds(50))
		.Execute(() => CleanDirectory(@"c:\MyFakeBinaries")));

RunTarget("Default");
```
Results in this:
<pre>
All packages listed in packages.config are already installed.
Installing 'Polly 2.2.3'.
Successfully installed 'Polly 2.2.3'.
Addin: Polly, adding Reference C:/Dev/tools/Addins/Polly/lib/net35/Polly.dll
Addin: Polly, adding Reference C:/Dev/tools/Addins/Polly/lib/net40/Polly.dll
Addin: Polly, adding Reference C:/Dev/tools/Addins/Polly/lib/net45/Polly.dll
Addin: Polly, adding Reference C:/Dev/tools/Addins/Polly/lib/portable-net45+netcore45+wpa81+wp8+MonoAndroid1+MonoTouch1/Polly.dll
<b>Error: error CS1704: An assembly with the same simple name 'Polly' has already been imported.  Try removing one of the references (e.g. 'C:\Dev\tools\Addins\Polly\lib\net45\Polly.dll') or sign them to enable side-by-side.</b>
[...]

PR #430, opened 2015-10-02
Change the way assembly references are loaded.

### LINKED ISSUE #415: The #addin directive should import referenced aliases
If a NuGet package (added with the `#addin` directive) reference another assembly in the package that contains an alias, it will not be referenced properly.

(Reported by @jrnail23 in #382)

patriksvensson:
We've been using Assembly.LoadFile to load script assembly references
which work great until you load an assembly where there is a
dependency to a previously loaded assembly. By using Assembly.LoadFrom
we make sure that assemblies are loaded properly.

Closes #415

### DIFF
--- a/src/Cake.Core/Scripting/ScriptRunner.cs
+++ b/src/Cake.Core/Scripting/ScriptRunner.cs
@@ -77,7 +77,7 @@ namespace Cake.Core.Scripting
             {
                 if (host.Context.FileSystem.Exist((FilePath)reference))
                 {
-                    var assembly = Assembly.LoadFile(reference);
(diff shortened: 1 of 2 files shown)


jrnail23:
@patriksvensson, I'm finally getting back to some Cake stuff, and I think you're right, this shouldn't have much (if any) impact on my NuGet addin work (#373).  At first I thought you might've been looking to alter the part of `AddInDirectiveProcessor` that adds the references, but now I see you're attacking the other end, where the ScriptRunner actually loads those references.
Looks good to me :smile:

Issue #431, opened 2015-10-03
Issue with globbing using dot character in conjunction with wildcard


molehillrocker:
We currently face an issue with globbing using wildcards in conjunction with a dot character.

**Scenario:**
We have multiple unit test projects that all end with ``.Test`` and one integration test project that ends with ``IntegrationTest``. Using ``src/**/*.Test.dll`` as seach pattern finds all unit test projects but also the integration test project.

**Solution structure:**
src
|-- Project.A
|-- Project.A.Test
|-- Project.B
|-- Project.B.Test
|-- Project.IntegrationTest

I guess this is a bug, but if I'm doing something wrong please direct me to a proper way of doing this. :smile: 

The current workaround is to specify all unit test projects manually, but obviously this enforces us to adjust the build file to for every unit test project we add.

molehillrocker:
I wrote a test case that shows the error. I currently have no Git available, which is why I post it here.

```
[Fact]
public void Should_Return_Files_For_Pattern_Ending_With_Character_Wildcard_And_Dot()
{
	// Given
	var fixture = new GlobberFixture(true);
	fixture.FileSystem.CreateFile("C:/Working/Project.A.Test.dll");
	fixture.FileSystem.CreateFile("C:/Working/Project.B.Test.dll");
	fixture.FileSystem.CreateFile("C:/Working/Project.IntegrationTest.dll");

	// When
	var result = fixture.Match("C:/Working/*.Test.dll");

	// Then
	Assert.Equal(2, result.Length);
	Assert.ContainsFilePath(result, "C:/Working/Project.A.Test.dll");
	Assert.ContainsFilePath(result, "C:/Working/Project.B.Test.dll");
}
```

molehillrocker:
I managed to install Git over here and made a PR: #433

patriksvensson:
I've merged the associated PR and will close this issue.
Thank you! :smile:

PR #432, opened 2015-10-03
Treat a . (dot) in glob pattern as dot instead of character wildcard

molehillrocker:
- A dot inside a pattern token is passed to the regex as common dot, where
  it is treated as an arbitrary character instead of real dot character.
- By prefixing with a backslash in the regex, the dot is treated as real
  character.


gep13:
@molehillrocker is there a reason that you closed and deleted this PR? Seem like a sensible suggestion to me.

molehillrocker:
I became confused due to the fact that the issue is labeled as #431 and my contribution as #432. I thought I'd named my branch wrong and changed it to #432, which ended in a PR with label #433. :smile: 

Lesson learned: Branch name has no effect on the id of the PR.

gep13:
Ah, gotcha.  Yeah, each Pull Request is actually a new issue, so you won't get a PR numbered the same as the Issue that it is intended to fix.

PR #433, opened 2015-10-03
Treat a . (dot) in glob pattern as dot instead of character wildcard

molehillrocker:
- A dot inside a pattern token is passed to the regex as common dot, where
  it is treated as an arbitrary character instead of real dot character.
- By prefixing with a backslash in the regex, the dot is treated as real
  character.


molehillrocker:
Is it possible to re-trigger the verification build? It seems that a timeout occured. :hushed:

patriksvensson:
Thanks for finding this! :+1: :smile:

molehillrocker:
Thanks for this nice tool! :smile:

PR #439, opened 2015-10-09
Fixes #437  for Powershell 2.0

### LINKED ISSUE #437: Minimum required PowerShell version
Today we ran into an issue that some colleagues could not run the build tasks I defined. After some investigation we found out that they had an older PowerShell version (``v2.0``) installed which failed to run the ``build.ps1``. My PowerShell version is ``3.0``.
Does it make sense to add a check for the PowerShell version on the startup of a build script and fail the build with a proper error message that hints the user to install an up-to-date version of PowerShell, or at least print the required version to the output?

After upgrading to ``v3.0``, everything works as expected. :+1:

phillipsj:


build.ps1
```diff
@@ -11,10 +11,16 @@ Param(
     [switch]$SkipToolPackageRestore
 )
 
+
```
devlead: No need for extra space

build.ps1
```diff
@@ -35,7 +41,13 @@ if($Mono.IsPresent) {
 
 # Try download NuGet.exe if do not exist.
 if (!(Test-Path $NUGET_EXE)) {
-    Invoke-WebRequest -Uri http://nuget.org/nuget.exe -OutFile $NUGET_EXE
+    try{
+        Invoke-WebRequest -Uri http://nuget.org/nuget.exe -OutFile $NUGET_EXE
```
devlead: Put nuget URL in $NUGET_URL variable so same it's used for both Invoke-WebRequest & WebClient, I prefer https URL is used.

gep13:
Do we really need this extra handling for PowerShell 2?

Since PowerShell is backwards compatible, why not simply use the PowerShell 2 code as the default, it will work equally well in later versions of PowerShell.

phillipsj:
Good point.  I didn't think much about it, just trying to help out.

devlead:
Agree @gep13 :+1:

gep13:
Sorry, didn't mean to sound harsh, certainly not what I intended. Just wanted to check of there is a reason to do the check. We have similar issues in Chocolatey, but our default is to use what works in PS2, knowing it will work in later versions.

Sent from my Windows Phone
________________________________

phillipsj:
I didn't think it was harsh at all.  I just didn't stop to think about it.
Should I make the changes?

devlead:
@phillipsj yes, change according to @gep13 feedback. Rebase into one commit and I'll look at it after that.

PR #440, opened 2015-10-10
Fixes #437 Converting PS 3 to PS 2 commands

### LINKED ISSUE #437: Minimum required PowerShell version

phillipsj:
Since PowerShell maintains backwards compatibility this was the
suggested solution.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -11,9 +11,11 @@ Param(
     [switch]$SkipToolPackageRestore
 )
 
+$PSScriptRoot = split-path -parent $MyInvocation.MyCommand.Definition;
 $TOOLS_DIR = Join-Path $PSScriptRoot "tools"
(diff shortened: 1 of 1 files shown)


gep13:
Looks good to me! :+1:

devlead:
Merged, thanks @phillipsj for your contribution :+1:

Issue #442, opened 2015-10-11
Eliminate Whitespace


gep13:
As discussed here:

https://gitter.im/cake-build/cake?at=56197d714e0fa3e5544804b0

gep13:
In order to help with this type of thing, I use a number of Visual Studio Extensions, and one that I would recommend is the Trailing Whitespace Visualizer:

https://visualstudiogallery.msdn.microsoft.com/a204e29b-1778-4dae-affd-209bea658a59

This includes this option:

![image](https://cloud.githubusercontent.com/assets/1271146/10417309/a789caba-702e-11e5-9c47-e04f049e14eb.png)

Which makes it really easy to maintain going forward.

gep13:
Another one is the EditorConfig extension which you can find here:

https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

Which helps keep things in order once you start moving between IDE's.  All it takes is the addition of a .editorconfig file, where you specify what settings, as a project, you want to use.

PR #443, opened 2015-10-11
(GH-442) Eliminated Whitespace across all files

gep13:
- A couple files also had mixed CR/LF and LF, I defaulted these to use
the Windows default


devlead:
@gep13 thanks for your contribution :+1:

Issue #148, opened 2015-01-07
MyGet build support


patriksvensson:
As @gep13 says [in a comment](https://github.com/cake-build/cake/pull/146#issuecomment-68878328) to PR #146, it would be nice with MyGet support.

> You might also want to throw MyGet into the mix as well :-)
> I have just added some MyGet helpers to psake:
> https://github.com/psake/psake-contrib/blob/develop/myget.psm1
> They expose similar Service Messaging to what TeamCity does.

gep13:
Ooo, I had forgotten about this one.  @patriksvensson you want me to take this and see what I can come up with?

devlead:
@gep13 go ahead :+1:

PR #456, opened 2015-10-18
Add MyGet Build System Support

gep13:
As documented here:

http://docs.myget.org/docs/reference/build-services

As requested here https://github.com/cake-build/cake/issues/148

### DIFF
--- a/src/Cake.Common/Build/MyGet/MyGetProvider.cs
+++ b/src/Cake.Common/Build/MyGet/MyGetProvider.cs
@@ -1,0 +1,173 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Linq;
+using Cake.Core;
+
+namespace Cake.Common.Build.MyGet
+{
+    /// <summary>
+    /// Responsible for communicating with MyGet.
+    /// </summary>
+    public sealed class MyGetProvider : IMyGetProvider
+    {
+        private const string MessagePrefix = "##myget[";
+        private const string MessagePostfix = "]";
+        private static readonly Dictionary<string, string> _sanitizationTokens;
+        private readonly ICakeEnvironment _environment;
+
+        static MyGetProvider()
+        {
+            _sanitizationTokens = new Dictionary<string, string>
+            {
+                { "|", "||" },
+                { "\'", "|\'" },
+                { "\n", "|n" },
+                { "\r", "|r" },
+                { "[", "|['" },
+                { "]", "|]" }
+            };
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MyGetProvider"/> class.
+        /// </summary>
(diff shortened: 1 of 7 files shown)


gep13:
cc @maartenba @xavierdecoster I would welcome any comments/feedback on this PR from a MyGet point of view, thanks! :smile_cat:

src/Cake.Common/Build/MyGet/MyGetProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether the current build is running on MyGet.
+        /// </summary>
+        /// <value>
+        /// <c>true</c> if the current build is running on MyGet; otherwise, <c>false</c>.
+        /// </value>
+        public bool IsRunningOnMyGet
+        {
+            get { return !string.IsNullOrWhiteSpace(_environment.GetEnvironmentVariable("BuildRunner")); }
```
xavierdecoster: When running on MyGet, the `BuildRunner` environment variable should always have the value `MyGet`.
Perhaps change this line to something like this?

```csharp
get
{
  var buildRunner = _environment.GetEnvironmentVariable("BuildRunner");
  return !string.IsNullOrWhiteSpace(buildRunner) && string.Equals("MyGet", buildRunner, StringComparison.OrdinalIgnoreCase)
}
```

src/Cake.Common/Build/MyGet/MyGetProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+                { "name", name },
+                { "value", value }
+            });
+        }
+
+        /// <summary>
+        /// Write a status message to the MyGet build log.
+        /// </summary>
+        /// <param name="message">Message contents.</param>
+        /// <param name="status">Build status.</param>
```
xavierdecoster: Perhaps document that the status defaults to `NORMAL` and can be set to `NORMAL`, `WARNING`, `FAILURE` or `ERROR`?

xavierdecoster:
Really like the contribution! Thanks :)
Two minor comments, but overall looks good to me.

gep13:
@xavierdecoster thanks for taking a look.  I will take a look at your comments :+1:

src/Cake.Common/Build/MyGet/MyGetProvider.cs
patriksvensson: Maybe you can introduce an enum for this instead of magic strings.

```csharp
public enum MyGetStatus
{
   Failure,
   Error,
   Warning,
   Normal
}
```

Or map the `Verbose` enum to the MyGet status codes.

gep13:
@xavierdecoster @patriksvensson I have fixed up this PR with the suggestions that we made, let me know what you think.

xavierdecoster:
:shipit: looks good to me!

gep13:
@devlead @patriksvensson rebased and ready

PR #458, opened 2015-10-18
Fail build on finding Duplicates

gep13:
This is a first stab at implementing the following:

https://github.com/cake-build/cake/issues/452

@patriksvensson  @devlead  What are your thoughts on this approach?

@yvschmid I would be interested to hear your thoughts as well.

Thanks!

### DIFF
--- a/src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
+++ b/src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
@@ -91,5 +91,10 @@ namespace Cake.Common.Tools.DupFinder
         /// Gets or sets a value indicating whether to show duplicates text in the report.
         /// </summary>
         public bool ShowText { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to throw an exception on finding duplicates
+        /// </summary>
+        public bool ThrowExceptionOnFindingDuplicates { get; set; }
     }
 }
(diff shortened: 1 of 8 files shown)


devlead:
@gep13  branch is out-of-date and won't currently build, just ping when you've had time to revisit this pr. And we'll review.

gep13:
:+1: yip, it is quite a bit out of date, but it is on my list to loop back to, thanks!

src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
```diff
@@ -91,5 +91,15 @@ public sealed class DupFinderSettings : ToolSettings
...
+
+        /// <summary>
+        /// Gets or sets an optional flag for if should throw exception on finding duplicates
+        /// </summary>
+        public bool? ThrowOnDuplicates { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether analysis of results file should be completed
+        /// </summary>
+        public bool PerformResultsFileAnalysis { get; set; }
```
DevWouter: The name and description of this property doesn't explain much and was a bit unclear to me. After looking at the usage a better description would be `AnalyzeOutputResult` and the description could be: "Gets or sets a value indicating whether the file-result should be analyzed."

gep13: @DevWouter yip, that makes sense.  Based on some of the other feedback, this property may actually not even be required, but I will decide that when I look again at this, which will hopefully be tonight.

gep13:
@devlead @patriksvensson ok, I have re-worked this PR, any chance you can take a look?

There currently is one failing test, and I wanted to get your thoughts on how to proceed.  The analysis of the Output File can *only* occur when the Output File property in the Settings has been set.  However, when this is the case, it then attempts to analyse the file, which obviously doesn't exist:

image

What would be your recommendation on how to correct this problem?

*NOTE:* I have decided to always analyze the DupFinder Results file, and write warnings to the output log when they are detected, but only actually fail the build when users have made an informed decision to do so.  This works best in terms of backwards compatibility for people who are already using this alias.

gep13:
@patriksvensson if you are looking for things to do, I think this one is ready for review/pulling in :-)

Issue #451, opened 2015-10-17
InspectCode: Provide ability to fail the build on detected issues


gep13:
Similar to the way that the Cake.Slack and Cake.Gitter Addin's provide the ability to fail the build if a chat message is not posted, Cake should provide the ability to fail the build if there are inspectcode errors detected.

Issue #449, opened 2015-10-17
DupFinder: Add ability to perform XSL Transform of output file


gep13:
Cake should provide the ability to provide a default XSL transform of the resulting dupfinder report xml file, into a human readable html file.

In addition, Cake users should be able to provide their own custom XSL file for performing the transformation.

gep13:
As discussed with @patriksvensson and @devlead this is not something that will be included in Cake.

PR #459, opened 2015-10-19
Fail build on finding InspectCode Violations

gep13:
This is a first stab at implementing the following:

#451

@patriksvensson @devlead What are your thoughts on this approach?

@yvschmid I would be interested to hear your thoughts as well.

Thanks!


gep13:
@yvschmid so in general, am I right in saying that you think that including the ability to parse the results file makes sense?  Also the ability to perform an XSL Transform on the resulting XML file to make a more human readable report (I have created another Issue to cover that work).

What are your thoughts on NOT performing the analysis when you are passing in a config file?  Good?  Bad?

Thanks for the review!

patriksvensson:
@gep13 Yes, I agree with @yvschmid that you should use `_fileSystem.GetFile()` and use the file system abstractions.

I'm not sure about adding XSL transform though. It feels like it would be more specialized if you instead added documentation for how to do this with the XSLTransform utility. Otherwise we should add this functionality for all other tools that produces some form of XML (xUnit, NUnit, MSTest, Fixie etc) for consistency. Just my opinion.

gep13:
Ok, I will make the suggested change to use the File System abstractions.

With regard to the XSL Transform, my thoughts were the following...

* Provide a default XSL Transform to take the XML file and make it a human readable HTML file
* Provide the ability for the consumer to override the default XSL file with one of their own choice

If we don't provide this ability in the Cake Addin, then the flow becomes a little bit jarring.  For example, assume we add the ability in this PR to throw an exception if there is a problem after analysis, we would need to add a try/catch around the call to `InspectCode`, and after checking the exception, we would then need to make a second call to the XSL alias to transform the report (which is really when you want to read it), as opposed to setting a property `PerformXslTransform` when you originally call `InspectCode`.

I much prefer the latter option, but that is just my opinion.  If this isn't something you want included in the built-in Cake alias, I am happy to close off issues https://github.com/cake-build/cake/issues/450 and https://github.com/cake-build/cake/issues/449.

devlead:
@gep13 I like the possibility do get more verbose logging of errors. The _"reporting"_ part is neat, but kinda agree with @patriksvensson . we should give it some thought if we could do something generic. Many tools to have an xml report. So don't argue with it's usefulness.

gep13:
@patriksvensson and this one :+1:

patriksvensson:
@gep13 Getting some merge conflicts on this. Could you fix and rebase? :)

PR #461, opened 2015-10-19
Add GitReleaseNotes alias

gep13:
@devlead @patriksvensson would love it if you could review this.

Thanks!

### DIFF
--- a/src/Cake.Common/Tools/GitReleaseNotes/GitReleaseNotesAliases.cs
+++ b/src/Cake.Common/Tools/GitReleaseNotes/GitReleaseNotesAliases.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.Annotations;
+using Cake.Core.IO;
+
+namespace Cake.Common.Tools.GitReleaseNotes
+{
+    /// <summary>
+    ///  Contains functionality related to ReSharper's duplication finder
+    /// </summary>
+    [CakeAliasCategory("GitReleaseNotes")]
+    public static class GitReleaseNotesAliases
+    {
+        /// <summary>
+        /// Generates a set of release notes based on the commit history of the repository and specified settings.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="outputFile">The output file.</param>
+        /// <param name="settings">The settings.</param>
+        [CakeMethodAlias]
+        [CakeAliasCategory("GitReleaseNotes")]
+        public static void GitReleaseNotes(this ICakeContext context, FilePath outputFile, GitReleaseNotesSettings settings)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            var runner = new GitReleaseNotesRunner(context.FileSystem, context.Environment, context.ProcessRunner, context.Globber);
+            runner.Run(outputFile, settings);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


gep13:
@jakeginnivan would you be in a position to take a look at this as well?  Would love your input, thanks!

gep13:
I am just looking again at this, and I think I need to change the GitReleaseNotes Alias.  Looking at this:

https://github.com/GitTools/GitReleaseNotes/blob/master/src/GitReleaseNotes/Program.cs#L35

It would appear that the OutputFile is a mandatory command line parameter, which makes sense, so I will refactor the Alias to account for that.

devlead:
@gep13 Agree mandatory arguments should be parameters.

gep13:
:+1:

src/Cake.Common/Tools/GitReleaseNotes/IssueTracker.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace Cake.Common.Tools.GitReleaseNotes
+{
+    /// <summary>
+    /// The IssueTracker to be used.
+    /// </summary>
+    public enum IssueTracker
```
devlead: @gep13 as this basically becomes a global available enum we should probably prefix if with GitReleaseNotes.

gep13: As in `GitReleaseNotesIssueTracker`?

devlead: Yes. All alias's namespaces  are added when imported, names are basically global and currently this usages is pretty "local".

gep13: Gotcha, that makes sense, I will fix this up.

gep13:
@devlead I have fixed up the issues that you mentioned in the earlier review of this, any chance you can take another look, and let me know what you think?  Cheers!

gep13:
@devlead looks like the Travis build failed due to timeout, can you kick it again?

patriksvensson:
@gep13 Kicked!

gep13:
@devlead @patriksvensson I "think" I have refactored this tool as well, would appreciate any feedback, thanks!

src/Cake.Common/Tools/GitReleaseNotes/GitReleaseNotesAliases.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.Annotations;
+
+namespace Cake.Common.Tools.GitReleaseNotes
+{
+    using global::Cake.Core.IO;
```
patriksvensson: Is there a reason this is using the global alias (and inside the namespace)?

patriksvensson:
@gep13 Left a comment. Other than that this looks good to me.

src/Cake.Common/Tools/GitReleaseNotes/GitReleaseNotesAliases.cs
gep13: Yes :-( ReSharper is currently fighting me, and I haven't sat down and looked into it yet.  I have ReSharper set up with all the default rules, including putting usings with the namespaces, which isn't what cake does.  As a result, each time I have to remember to put them back properly.  I am starting to move away from that rule in general though, and I need to turn off that refactoring within ReSharper, I just need to find where it is :smile: 

I also found another instance of `global::` within one of the files for GitReleaseManager, and although not related to this PR, I took the opportunity to change it just now in this PR.  Let me know if you want me to back that out.

gep13:
@patriksvensson updated and rebased.

devlead:
@gep13 merged, thanks :+1:

PR #465, opened 2015-10-20
 Added CDATA attribute to releasenotes

gep13:
@devlead this is quite crude, but also almost effective :smile_cat: 

The end result is the following:

image

Are you doing some escaping of the strings that are added into the nuspec file?  Add initial glance, I couldn't see anything obvious.

### DIFF
--- a/src/Cake.Common.Tests/Properties/Resources.resx
+++ b/src/Cake.Common.Tests/Properties/Resources.resx
@@ -140,9 +140,9 @@
     &lt;licenseUrl&gt;https://license.com&lt;/licenseUrl&gt;
     &lt;requireLicenseAcceptance&gt;true&lt;/requireLicenseAcceptance&gt;
     &lt;iconUrl&gt;https://icon.com&lt;/iconUrl&gt;
-    &lt;releaseNotes&gt;Line #1
+    &lt;releaseNotes&gt;&lt;![CDATA[Line #1
 Line #2
-Line #3&lt;/releaseNotes&gt;
+Line #3]]&gt;&lt;/releaseNotes&gt;
   &lt;/metadata&gt;
   &lt;files&gt;
     &lt;file src="tools\**" target="tools" /&gt;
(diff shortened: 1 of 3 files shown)


devlead:
Don't you need to use `XmlDocument.CreateCDataSection`? Think it's otherwise treated as a xml value.

gep13:
No idea, I will do some checking, I was going for the most pragmatic approach, i.e. just throw it in :smile_cat: 

Will do some digging.

RichiCoder1:
Ditto what @devlead.

gep13:
@devlead @RichiCoder1 ok, I have fixed up this PR, which now results in this:

image

Thoughts?

patriksvensson:
The tests are failing (since they're not updated) but as soon as those are adjusted I'm ready to merge this.

RichiCoder1:
LGTM

gep13:
@patriksvensson I was sure that I ran the tests on this one, but it must have been one of the other branches that I was thinking about.  Sorry about that, I will get that fixed up today.

gep13:
@patriksvensson @devlead on the subject of these tests though, have you guys heard of the following:

https://github.com/approvals/ApprovalTests.Net

We use it over on the GitVersion project to make assertions based on the results of Unit Tests.  You can see an example here:

https://github.com/GitTools/GitVersion/blob/039a43b092134c6a42902fc162fcb4cbc1265c14/src/GitVersionCore.Tests/JsonVersionBuilderTests.cs#L26

Here is it making an assertion against the result, and an approved version of the result, which you can see here:

https://github.com/GitTools/GitVersion/blob/039a43b092134c6a42902fc162fcb4cbc1265c14/src/GitVersionCore.Tests/JsonVersionBuilderTests.Json.approved.txt

i.e. a file that is checked into source control, rather than in the embedded resources of the project.  If a Unit Test fails, the failing text will be written out to a file on the file system, which means that you can then do a direct comparison to the files, using something like KDiff, or whatever tool you like.

It is driven by conventions, where it looks for a file with the same name as the name of the unit test.

Just thought I would mention it, as I thought that it might be something useful for you to include within the code base.

gep13:
@devlead @patriksvensson I have just rebased this PR on the latest develop, and assuming the builds pass, I think it is ready for merging.

Also, did you see my question/suggestion above?

patriksvensson:
@gep13 Can you rebase this into one commit?

patriksvensson:
@yvschmid There has been some changes to the `Tool` base class. Please rebase against `develop` and inherit from `Cake.Core.Tooling.Tool<TSettings>`.

gep13:
@patriksvensson done

Also, was the comment for @yvschmid meant to be in this PR?

patriksvensson:
@gep13 Hmmm. No. Meant his PR :)

gep13:
@devlead @patriksvensson can one of you give this a kick? The Travis CI build failed.

devlead:
@gep13 kicked it :cake:

devlead:
@gep13 thanks.

PR #480, opened 2015-11-02
Fixing Typos

gep13:
- Made use of the ReSpeller Extension for ReSharper

### DIFF
--- a/src/Cake.Common/Tools/Chocolatey/ApiKey/ChocolateyApiKeySettings.cs
+++ b/src/Cake.Common/Tools/Chocolatey/ApiKey/ChocolateyApiKeySettings.cs
@@ -55,6 +55,6 @@ namespace Cake.Common.Tools.Chocolatey.ApiKey
         /// Gets or sets a value indicating whether to run in allow unofficial mode.
         /// </summary>
         /// <value>The allow unofficial flag</value>
-        public bool AllowUnoffical { get; set; }
+        public bool AllowUnofficial { get; set; }
     }
 }
--- a/src/Cake.Common/Tools/Chocolatey/Pack/ChocolateyPackSettings.cs
+++ b/src/Cake.Common/Tools/Chocolatey/Pack/ChocolateyPackSettings.cs
@@ -186,6 +186,6 @@ namespace Cake.Common.Tools.Chocolatey.Pack
         /// Gets or sets a value indicating whether to run in allow unofficial mode.
         /// </summary>
         /// <value>The allow unofficial flag</value>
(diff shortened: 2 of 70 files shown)


DevWouter:
There are a few cases where `_field` is replaced by `this._field` since the rest of the source doesn't use this you might want restore this to the original behaviour.

gep13:
@DevWouter  Ah, good catch!  That is ReSharper getting ahead of itself, I will correct.

gep13:
I have fixed up the issue pointed out by @DevWouter

DevWouter:
Looks good now. :+1:

src/Cake.Common/Tools/Chocolatey/ApiKey/ChocolateyApiKeySettings.cs
```diff
@@ -55,6 +55,6 @@ public sealed class ChocolateyApiKeySettings : ToolSettings
         /// Gets or sets a value indicating whether to run in allow unofficial mode.
         /// </summary>
         /// <value>The allow unofficial flag</value>
-        public bool AllowUnoffical { get; set; }
+        public bool AllowUnofficial { get; set; }
```
devlead: This is a breaking change as it's an public API, though new so perhaps now's the time. Don't think many have taken an dependency on this yet, but for those that have their code will break.

src/Cake.Common/Tools/Chocolatey/Pack/ChocolateyPackSettings.cs
```diff
@@ -186,6 +186,6 @@ public sealed class ChocolateyPackSettings : ToolSettings
         /// Gets or sets a value indicating whether to run in allow unofficial mode.
         /// </summary>
         /// <value>The allow unofficial flag</value>
-        public bool AllowUnoffical { get; set; }
+        public bool AllowUnofficial { get; set; }
```
devlead: Same here.

devlead:
@gep13 think it looks good, there's breaking changes which we really try to avoid. But in this case makes sense to do an exception,

Please get branch up to date and I'll look gain.

gep13:
@devlead rebased and pushed.

That is a very good point about the potential breaking change.  Although in this case, the Chocolatey Alias is slated for release in 0.5.6, so not currently available outwith MyGet Sources, so I "think" in this instance, we are ok.  Right?

devlead:
@gep13 yes in this case it's "ok", it's just get a bad gut feeling when seeing rename of public members :wink:

Issue #474, opened 2015-10-29
Command line parameters should follow conventions


patriksvensson:
We should add support for the `--longoption=value` convention and (sometime in the future) deprecate the (non conventional) `-longoption=value`.

gep13:
Why deprecate?  Why not support both?

patriksvensson:
@gep13 Because it's wrong :smile:

Jokes aside, we want Cake to be intuitive, and for a *nix person this is confusing. Multiple letters combined with a single dash is multiple switches in their short form.

I.e. `-test` is the same as `-t -e -s -t`.

gep13:
Ah, I see, that makes sense :+1:

DevWouter:
Let's see if I can get this to work :)

DevWouter:
Alright, since I haven't updated the issue with some details, here they are.

This is mostly about the parser.

First a few valid arguments:
```ps
# Long keys are always two dashes followed by at least two characters
.\cake --longkey value
# Short keys are always dash followed by a single char
.\cake -s short-key
# Duplicate keys are supported by the parser (might not be supported by the data-structure)
# In the following example the "keys" property should be a list.
.\cake --keys a --keys b
# Supports quotes
.\cake --double-quote "here we a space"
.\cake --single-quote 'here we a space'
# Supports escaped quotes
.\cake --double-quote "write a \"quote\""
.\cake --single-quote 'write a \'quote\''
# Values are not required
.\cake --no-value --one-with-value a_value --yet-no-value-again
```

And one of the last thing I have been thinking of is using list seperator like git uses. However unles there is use-case for it we might rather avoid it.
```powershell
.\cake --k1 --k2 -- And here is a long argument which can be anythig
```

DevWouter:
From the chat with @patriksvensson 

**tl;dr:** Multiple arguments are supported by the parser but can't be used since `ICakeArguments` can't support it. 

```
patriksvensson 13:59
Well, I'm not against supporting multiple keys, but that would break the ICakeArguments interface. That's my main concern.

DevWouter 14:01
@patriksvensson Yes, but what the parsers supports can be more than the data-structure.
@patriksvensson We will still throw an error if that happens, but it won't be thrown by the parser as the grammar is correct.
```

DevWouter:
Reading through the ticket I noticed that the flags aren't supported yet. It should be quite simple to add them now.  First I want to collect some feedback on the PR (since it is my first) and then see what else requires work so we can close this ticket.

[...]

PR #481, opened 2015-11-02
Improve argument parser to use unix-arguments

DevWouter:
Related to #474 

The original version used a non-conventional single-dash argument as a way
of passing arguments. This has been changed to double-dash arguments for
long options and single-dash argument for single character options.

To prevent this commit from becoming a breaking change the old behaviour
is also supported. It might be removed in the future in favor of the new
style, however until documentation and features have been updated this
should have no effect for everyone.

In the new approach we have also introduced a new way to create additional
options in `CakeOptions`. The command line key name is now decided by an
attribute. In the future we might wanna expend that with documentation
attribute so that we can generate the help on the fly. (Also it would be
really cool if we could let cake parse the build file for additional
options).

All the newly introduced code has been written using TDD behaviour and
exception for two error messages (so that it would support the new
behaviour) none of the original UnitTests have been changed.

### DIFF
--- a/src/Cake.Tests/Cake.Tests.csproj
+++ b/src/Cake.Tests/Cake.Tests.csproj
@@ -76,6 +76,8 @@
     </Compile>
     <Compile Include="Unit\Arguments\ArgumentParserTests.cs" />
     <Compile Include="Unit\Arguments\ArgumentTokenizerTests.cs" />
+    <Compile Include="Unit\Arguments\UnixCommandArgumentParserTests.cs" />
+    <Compile Include="Unit\Arguments\UnixArgumentParserTests.cs" />
     <Compile Include="Unit\CakeApplicationTests.cs">
       <SubType>Code</SubType>
     </Compile>
--- a/src/Cake.Tests/Unit/Arguments/ArgumentParserTests.cs
+++ b/src/Cake.Tests/Unit/Arguments/ArgumentParserTests.cs
@@ -85,7 +85,7 @@ namespace Cake.Tests.Unit.Arguments
 
                 // Then
                 Assert.Null(result);
-                Assert.True(fixture.Log.Entries.Any(x => x.Message == "Multiple arguments with the same name (unknown)."));
+                Assert.True(fixture.Log.Entries.Any(x => x.Message == "Using a argument-name more than once is not allowed: -unknown"));
             }
 
             [Theory]
@@ -152,7 +152,7 @@ namespace Cake.Tests.Unit.Arguments
                 parser.Parse(arguments);
 
                 // Then
-                Assert.Equal("More than one build script specified.", fixture.Log.Entries[0].Message);
+                Assert.Equal("An \"--key\" should preceed the value.", fixture.Log.Entries[0].Message);
             }
 
             [Theory]
--- a/src/Cake.Tests/Unit/Arguments/UnixArgumentParserTests.cs
+++ b/src/Cake.Tests/Unit/Arguments/UnixArgumentParserTests.cs
@@ -1,0 +1,94 @@
+﻿using Cake.Arguments;
+using Xunit;
+
+namespace Cake.Tests.Unit.Arguments
+{
+    public class UnixCommandArgumentNameTests
+    {
+        [Theory]
+        [InlineData("--test")]
+        [InlineData("-t")]
+        public void Constructor_Should_Store_Valid_Value(string name)
+        {
+            var arg = new CommandArgumentNameAttribute(name);
+            Assert.Equal(name, arg.Name);
+        }
+
+        [Fact]
+        public void Should_Throw_If_Name_Is_Null()
+        {
+            var result = Record.Exception(() => new CommandArgumentNameAttribute(null));
+
+            Assert.IsArgumentException(result, "name", "CommandArgumentName can not be empty or null");
+        }
+
+        [Fact]
+        public void Should_Throw_If_Name_Has_No_Dashes()
+        {
+            var result = Record.Exception(() => new CommandArgumentNameAttribute("no-dash"));
+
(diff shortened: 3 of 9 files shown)


patriksvensson:
Hello @DevWouter,

My main concern with this PR is that it have over 900 additions. While perhaps more generic and extendable than what we already have, I think this PR goes beyond the scope of support double dashes when parsing command line arguments.

I think that your ideas about making the command line parser more conforming to standards is very sound, but I think we should use a more conservative approach. I think that supporting double dash and different quoting could be implemented in the existing solution without a complete rewrite.

I'm sorry, but I don't feel like we can accept this PR in it's current state. While I really appreciate your contribution, I don't think this is in line with what the issue originally was about.

What are your thoughts?

DevWouter:
_(Most of this are explanations and argumenting skip to the end to see my conclusion)_

Hello @patriksvensson,

Thanks for taking a look at it. As I see it you have the following concerns:

1. The size of the PR
2. No rewrite was required
3. Not in line with the original issue

### The size of the PR

The size of the PR can be explained due to the fact that 7 commits where squashed to one. Conceptually the following steps (TDD-style) were taken:

* Extracting the parser (which resulted in `UnixCommandParser`)
* Rewrite the parser (old: `ArgumentParser`, new: `UnixCommandParser`)
* Rewrite the processing of arguments (old: `ArgumentParser`, new: `ArgumentParser` and `CommandArgumentNameAttribute`)
* Add backwards compatibility

It might have been better if each one of those was a seperate PR.

### Why the rewrite?

The rewrite is a result of seperation of concern. The old `ArgumentParser` did three things:

1. Parse the command line
2. Proces the result of parsing 
3. Fill the correct fields of `CakeOptions` using names.

Granted the old one did this in one move, but the new one is better seperated

1. Use `UnixCommandParser` to proces the arguments
2. Use `CommandArgumentNameAttribute` to define the special arguments in `CakeOptions`
3. Use `ArgumentParser` as a strategy fill `CakeOptions`

This means that in the future:

* Changing the command parser means only touching `UnixCommandParser` (for example `rm -rf` would be translated as if it was `rm -r -f`)
* Adding/removing arguments means only touching `CakeOptions`
* Changing the strategy to fill `CakeOptions` means only touching `ArgumentParser`

**I agree** that the same *functional result* could be achieved with far less work, but this approach requires less legwork in the future.

### Not in line with the original issue

The issue doesn't describe much but this PR does way more then what is requested. About the extra work we can ask the following questions:

Was the extra work asked? No.
Was the extra work necessary? No, it could work without.
Is the extra work an improvement? Yes, better SoC and addtional functionality.
Does the extra work "work"? Yes, it works, no need for "air-quotes".

## My opinion

The solution goes further then the scope requires and even though the improvements are fine they are unrelated to the issue. 
If this PR would **only** be about fixing the issue it would be best to reject it. 

However if we ignore that the 'PR should only be about the issue' then the only argument so far would be that the change is simply big. That alone should not be a reason to reject a PR.

## My feelings (since you don't care about my opinion :laughing: )

I think we should reject this PR unless the above provides you with another insight. The change adds value but only very little value of that change is actually used. Basically this technical value has a risk of becoming a technical debt.

In addition of all the parsers (commandline and otherwise) I have written I don't love this change as much as I know I should. It might be because I'm tired but when my gut tells me something is off then something often is.

Combine that with the arguments provided by @patriksvensson I also don't feel like excepting.

patriksvensson:
Hello @DevWouter,

First, I want to say that I do care about your opinion. I asked what your thoughts were about this because I think it's important that there is a discussion and mutual understanding. I also care about how you feel about this since I don't want anyone to take my opinion of a code change as personal critique. That was never my intention.

Your rewrite is absolutely more extensible and modularized and adhere more to the SOLID principles, but it (naturally) also introduces more complexity and a lot more code. I'm not sure if the change warrant a rewrite since the number of arguments are known and probably won't change in the foreseeable future. I prefer to cross that bridge when we come to it.

If you feel confident that rewriting the argument parser is the right way to go (I might very well be wrong about this), I think that we should properly discuss what we want from that rewrite and get buy in from the rest of the team. There are perhaps other things as well that we want to address with a rewrite e.g. separate tokenization and parsing which is done in one pass at the moment.

I also want to say that I really appreciate the work you've done even thought I had reservations about this PR, and I hope that this doesn't discourage you from continuing your work with Cake. I would love to see more involvement and contributions from you in the future.

DevWouter:
To clarify: The part about "not caring about opinions" is a joke. I know you respect it (why else would you ask for it?) but I also know you don't want to hurt/discourage me. By making it a joke we have had the chance to adres this in a positve way instead of when we need to justify or defend a remark that was badly received by the other party. 

From a purely technical point the changes are sound and I also see it as an improvement. If we would stop the discussion there then the only conclusion should be to accept it.

*However...*
As pointed out by both you and me the manner in which these changes are done is undesirable. The change changes too much, some of those individual changes should be discussed seperatly. 

*Meaning...*
Now we are forced to reject the PR if we want to reject one thing. To prevent that from happening I think we should reject this purely based on "this is not how we want to do it" and not even look at the value of the change. After rejecting we can use multiple PRs to get the same result, but that way every change gets the opportunity to be discussed allowing for better PRs in general.

*Unless...* 
We all think this change is perfect example of divine code in which case we should accept it. Sadly I don't think that.

Looking back at both our comments, I think we should close this PR. Do you agree?

devlead:
Hi @DevWouter,

Great feedback, closing this PR and splitting into multiple staged fine grained PR's sounds like an excellent idea.
Would make it allot easier to see intent, review, give & get feedback.

I'll close this for now and we'll revisit the issue in a new PR.

A good starting point would be a single focused PR that just adds what's needed to solve #474 and then follow up with suggested improvement/refactors in the form of Issues (to allow discussions), followed by PR's.

DevWouter:
@devlead Thanks for closing, will do!

PR #482, opened 2015-11-04
Added missing ReportGenerator parameter validation

devlead:
Fix missing context parameter validation.


yvschmid:
@devlead What's the point of this? Does this generate a more meaningfull error message (stacktrace)? :confused:

devlead:
Correct,  catches error in right place, if you build release you would've gotten errors on these as that triggers our StyleCop rules.I missed this on the review.

yvschmid:
How do you build release locally? `cake -configuration=Release` doesn't seem to do the same thing, at least for me.

devlead:
Release config in studio will trigger it.

yvschmid:
:+1: building with visual studio, I forgot about that :laughing:

PR #485, opened 2015-11-04
Fix for "Input string was not in a correct format." exception.

olivier-spinelli:
Fix for "Input string was not in a correct format." exception when logging a JSON string.

Considering the {} placeholder only when actual arguments are provided easily fixes it.
The second commit avoids to keep {{ in log output: an escaped {{ is transformed into one {.


olivier-spinelli:
Oups! Sorry, this breaks a test! I couldn't imagine this (that is clearly a fix for me) would break a test. I'll try to find some time to investigate this strange test,

patriksvensson:
Yes, it looks like your fix broke something :(
The tests doesn't run on Travis at the moment due to problems with xUnit on Mono.

devlead:
@olivier-spinelli had any time to look at above yet? Branch also out of date and needs to be rebased.

patriksvensson:
Closing this PR for now.

PR #487, opened 2015-11-04
Closes #486: CopyFile logs incorrect targetFilePath

### LINKED ISSUE #486: BUG: CopyFile logs incorrect targetFilePath
The log message written by `FileCopier.CopyFileCore` reports the source file path as the target file path:
```c#
context.Log.Verbose("Copying file {0} to {1}", absoluteFilePath.GetFilename(), absoluteFilePath);
```
So the log message makes it look like the file is just getting copied to itself (although the copy actually works correctly).

jrnail23:


devlead:
Looks sold @jrnail23 :+1: , could you get the branch up to date and I'll merge.

jrnail23:
@devlead, should be ready to go now, pending CI build

patriksvensson:
Restarted Linux build that had timed out when restoring NuGet packages.

jrnail23:
@patriksvensson, did it happen again?

patriksvensson:
@jrnail23 Hmm. Yes :cry: I'll restart it again.

I should really try to get the packages properly cached on Travis since this fail in more than 50% of the builds.

src/Cake.Common.Tests/Unit/IO/FileAliasesTests.cs
```diff
@@ -200,6 +202,22 @@ public void Should_Copy_File()
...
             }
+
+            [Fact]
+            public void Should_Log_Verbose_Message_With_Correct_Target()
+            {
+                // Given
+                var fixture = new FileCopyFixture();
+
+                // When
+                FileAliases.CopyFile(fixture.Context, "./file1.txt", "./target/file1.txt");
```
patriksvensson: Sorry for nitpicking, but a better test would be to test for `./file1.txt` and `./target/file2.txt`.
If I understand correctly the bug will output the filename of the target file first, which would give the same results.

jrnail23: I agree, but the fixture isn't very straightforward in terms of changing that part.  I did, however, start from a failing test.
I can still try to get that to work if you like.

patriksvensson: Ah, I see. I think it's OK then :smile:

patriksvensson:
This looks good to me :+1: 
Ok to merge?

patriksvensson:
@jrnail23 Looks like this needs to be rebased against develop.

jrnail23:
@patriksvensson, rebased again, ready to go (pending CI build)

devlead:
Merged @jrnail23, thanks for contributing :+1:

jrnail23:
:+1:

PR #489, opened 2015-11-05
Update version & copyright

devlead:


build.cake
```diff
@@ -75,7 +75,7 @@ Task("Patch-Assembly-Info")
         Version = version,
         FileVersion = version,
         InformationalVersion = semVersion,
-        Copyright = "Copyright (c) Patrik Svensson, Mattias Karlsson and contributors"
+        Copyright = "Copyright (c) Patrik Svensson, Mattias Karlsson, Gary Ewan Park and contributors"
```
gep13: @devlead related to this, should these also be updated?

image

devlead: Well as they're overwritten by Cake nuspec transformer I actually think it would be best to remove any properties specified by `build.cake` and only keep those unique to each individual project.
@patriksvensson what do you think?

gep13: Yes, this seems like a good idea to me :+1:

Issue #505, opened 2015-11-07
Create Chocolatey Package for Cake


gep13:
To allow people to install Cake globally, and run from everywhere.

patriksvensson:
:+1: to this :smile:

gep13:
Now that we have the Chocolatey Alias, it should be a breeze :-D

gep13:
@patriksvensson I was actually keeping this one open, as a reminder to change to use the new OutputDirectory setting that the ChocoPack command exposes, but that wasn't ready until the latest release.  Now I will be able to do that again, or should I create another issue to cover that?

patriksvensson:
Sorry! I didn't know. A new issue might be better for that but you decide :)

gep13:
It's all good.  The fact that it wasn't clear why it was still open, is enough of a reason to have a separate issue. Will create one just now :-)

PR #506, opened 2015-11-07
Create Chocolatey Package for Cake

gep13:
This is a first pass at creating the Chocolatey Package for Cake.

It isn't quite as simple to create a Chocolatey Package, as it is for a Nuget package, as there is no concept of BasePath and OutputDirectory for Chocolatey, so there is a little bit more work to do, but the end result is the same.

Fixes GH-505

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -153,6 +153,31 @@ Task("Zip-Files")
     Zip(binDir, packagePath, files);
 });
 
+Task("Create-Chocolatey-Packages")
+    .IsDependentOn("Copy-Files")
+    .WithCriteria(() => isRunningOnWindows)
+    .Does(() =>
+{
+  // Create Cake package.
+  ChocolateyPack("./nuspec/Cake.Portable.nuspec", new ChocolateyPackSettings {
+      Version = semVersion,
+	  ReleaseNotes = releaseNotes.Notes.ToArray(),
+      Files = new [] {
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Cake.exe")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Cake.Core.dll")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Cake.Core.xml")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Cake.Common.dll")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Cake.Common.xml")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/NuGet.Core.dll")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Mono.CSharp.dll")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Autofac.dll")},
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/LICENSE")}
+    }
+  });
+
+  MoveFileToDirectory(string.Concat("cake.portable.", semVersion, ".nupkg"), nugetRoot);
+});
+
 Task("Create-NuGet-Packages")
     .IsDependentOn("Copy-Files")
     .WithCriteria(() => isRunningOnWindows)
@@ -227,7 +252,7 @@ Task("Publish-MyGet")
         throw new InvalidOperationException("Could not resolve MyGet API key.");
     }
 
-    foreach(var package in new[] { "Cake", "Cake.Core", "Cake.Common", "Cake.Testing" })
+    foreach(var package in new[] { "Cake", "Cake.Core", "Cake.Common", "Cake.Testing", "cake.portable" })
     {
         // Get the path to the package.
         var packagePath = nugetRoot + File(string.Concat(package, ".", semVersion, ".nupkg"));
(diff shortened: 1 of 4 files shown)


gep13:
@patriksvensson @devlead let me know what you think.

build.cake
```diff
@@ -227,7 +251,7 @@ Task("Publish-MyGet")
         throw new InvalidOperationException("Could not resolve MyGet API key.");
     }
 
-    foreach(var package in new[] { "Cake", "Cake.Core", "Cake.Common", "Cake.Testing" })
+    foreach(var package in new[] { "Cake", "Cake.Core", "Cake.Common", "Cake.Testing", "cake.portable" })
```
patriksvensson: Should be upper case `P` in `Cake.Portable` right?

build.cake
```diff
@@ -153,6 +153,30 @@ Task("Zip-Files")
...
+Task("Create-Chocolatey-Packages")
+    .IsDependentOn("Copy-Files")
+    .WithCriteria(() => isRunningOnWindows)
+    .Does(() =>
+{
+  // Create Cake package.
+  ChocolateyPack("./nuspec/Cake.Portable.nuspec", new ChocolateyPackSettings {
+      Version = semVersion,
+      Files = new [] {
+        new ChocolateyNuSpecContent {Source = string.Concat("./../", binDir.Path.FullPath, "/Cake.exe")},
```
patriksvensson: Maybe we can add a variable that holds the absolute path to binDir before ChocolateyPack (or make all directories absolute when specifying them)?

Then we should be able to simply use something like this:

```csharp
new ChocolateyNuSpecContent { Source = binDir + File("Cake.exe") }
```

nuspec/Cake.Portable.nuspec
```diff
@@ -0,0 +1,31 @@
...
+    <tags>Cake Script Build</tags>
+    <copyright>Copyright (c) Patrik Svensson, Mattias Karlsson and contributors</copyright>
+    <licenseUrl>https://github.com/cake-build/cake/blob/develop/LICENSE</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <iconUrl>https://raw.githubusercontent.com/cake-build/graphics/master/png/cake-medium.png</iconUrl>
+    <!--<dependencies>
+      <dependency id="" version="__VERSION__" />
+      <dependency id="" />
+    </dependencies>-->
+    <releaseNotes>https://github.com/cake-build/cake/releases</releaseNotes>
```
patriksvensson: Couldn't we provide release notes from the build script? Or is Chocolatey releasenotes supposed to be an URL?

nuspec/Cake.Portable.nuspec
```diff
@@ -0,0 +1,31 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>cake.portable</id>
```
patriksvensson: Should it really be lower case for id? `Cake.Portable` looks better.

build.cake
gep13: yes, I think that makes a lot of sense :+1: I will fix this up.

nuspec/Cake.Portable.nuspec
gep13: No, there is no issue here.  I guess I was going with my gut feeling on only storing the Release Notes in one place, rather than duplicating.  The benefit of the Release Notes section in the Chocolatey Gallery is that you can use full markdown, so a hyperlink is actually rendered as that, rather than just text.  I can add the contents of the releasenotes.md file into here though :+1:

nuspec/Cake.Portable.nuspec
gep13: This is really in line with the guidance that we provide on the Chocolatey Wiki:

https://github.com/chocolatey/chocolatey/wiki/CreatePackages#naming-your-package

I used `choco new` to create the package which applies this guidance as well.  Although we can make it different, the result on chocolatey.org on the package page will be the lowercase version.  For example, the casing of this package, is GitVersion.Portable (I think it made it through before the guidance was finalised): https://chocolatey.org/packages/GitVersion.Portable is `GitVersion.Portable`, but is displays at `gitversion.portable` when shown in the installation step.

I can certainly make the ID capital, but depending on the moderation, it might be pushed back.

patriksvensson: I understand. Then we should follow the conventions.

nuspec/Cake.Portable.nuspec
patriksvensson: OK. You decide :)

build.cake
gep13: As discussed, going to leave this the way it is, based on recommendations from Chocolatey:

https://github.com/chocolatey/chocolatey/wiki/CreatePackages#naming-your-package

build.cake
gep13: @patriksvensson based on your suggestion, I have done this:

image

As making the directory absolute when it is declared causes other issues.  But when I try to run the build, I get this:

image

Should a DirectoryPath + File put the / in for me?

nuspec/Cake.Portable.nuspec
gep13: :+1:

nuspec/Cake.Portable.nuspec
gep13: I will pull in the releasenotes in the same way as we do for the NuGet Packages, and that way they are consistent, and if we change it to something else, we can change it everywhere.

build.cake
gep13: As discussed, going to roll this back to the way it was for now, until above is fixed.

patriksvensson: This is a bug. I suspect the `FilePath` (absoluteBinDir) and the `ConvertableFilePath` (Cake.exe) converts to string to satisfy the `+` (since they both have implicit conversion to string), and since FilePath has implicit conversion from string, the result gets wrong. Need to think about this one...

gep13:
@patriksvensson let me know what you think about the recent changes.  If you are happy, I will squash into 1 commit, and hopefully get this pulled in.

patriksvensson:
@gep13 Looks good to me. Tell me when it's done and I'll merge it for you :smile:

build.cake
devlead: Couldn't one for now use `binDir.CombineWithFilePath("Cake.exe")`

gep13:
@patriksvensson ok, good to go.

patriksvensson:
Merged :+1:

PR #509, opened 2015-11-09
[WIP] Changing Copyright Information

gep13:
So, although this was taken care of by @devlead here https://github.com/cake-build/cake/commit/e87b3159ebd8a3f636700fa547dee763b060f7d5 for any future packages, for the same reason that we keep the SolutionInfo.cs file in source, I think it makes sense to update the nuspec's to contain the full information as well.  This was after a conversation with @patriksvensson in Slack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Make sure you've read the [contribution guidelines](http://cakebuild.net/contrib
 
 ## License
 
-Copyright © 2014 - 2015, Patrik Svensson, Mattias Karlsson and contributors.
+Copyright © 2014 - 2015, Patrik Svensson, Mattias Karlsson, Gary Ewan Park and contributors.
 Cake is provided as-is under the MIT license. For more information see [LICENSE](https://github.com/cake-build/cake/blob/develop/LICENSE).
 
 * For Roslyn, see https://github.com/dotnet/roslyn/blob/master/License.txt
--- a/nuspec/Cake.Common.nuspec
+++ b/nuspec/Cake.Common.nuspec
@@ -11,7 +11,7 @@
     <projectUrl>https://github.com/cake-build/cake</projectUrl>
     <iconUrl>https://raw.githubusercontent.com/cake-build/graphics/master/png/cake-medium.png</iconUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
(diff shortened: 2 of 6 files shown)


devlead:
Couldn't we just remove the tags that are replaced by the build script so it's just one place to upgrade?

gep13:
I am happy to go down that route as well, if that is the general consensus.

gep13:
@patriksvensson what were your thoughts on this?  I am not fussed either way, I can totally see where @devlead is coming from.

devlead:
@gep13 @patriksvensson in my opinion if we can remove unused "code" the great and will make stuff clearer/more obvious.

gep13:
@devlead actually, I am not seeing in the build.cake file where the Copyright information would flow through to the generated nuget package.  We would need to add the setting of the Copyright property into the NugetPack command, right?  Currently, only the AssemblyInfo copyright is being updated, and I don't think that the Copyright information from the AssemblyInfo flows through to the nuget package.

Or am I missing something else?

devlead:
@gep13 you're absolutely correct... Allots been added in nuget pack since Cake build.cake was written I guess. Then I'll guess my recommendation would be to find all common data/properties between nuspec's & also assembly info, put in a new global anonymous class variable that can be reused in script.

Here @patriksvensson is free to chime in with some thoughts :wink: 

Actually nuspecs are no longer required by Cake i don't user them in anymore for other projects, I just supply everything to the nuget pack settings class.

patriksvensson:
I prefer NuSpecs to keep the build scripts clean.

We can clean up the NuSpecs but maybe it's out of the scope for the issue?

devlead:
@patriksvensson ok sure. That'll work, we can just merge this and discuss later :+1: Personally prefer as few dependencies & duplication as data as possible. Might bloat build.cake a couple of lines, but could also reduce errors :wink:

gep13:
Okay, to recap....

The changes in this PR are inline with how the build.cake file currently works, and therefore are required.  So we can go ahead and merge this in, once AppVeyor build responds correctly.

I will open another issue to track the discussion about whether to remove nuspecs and use Settings in NuGetPack instead.

devlead:
@gep13 yes :+1:

PR #513, opened 2015-11-11
Added string replacement token to error-log call.

jhogstrom:
Minor bug when reporting an error.


patriksvensson:
Merged :+1: Thank you! :smile:

Issue #510, opened 2015-11-09
Allow specifying an output directory for ChocolateyPack


vktr:
Chocolatey currently does not provide an option to specify the output directory for generated packages.

Cake can provide this functionality by setting the working directory when launching the `choco.exe` process.

PR #520, opened 2015-11-13
Added WorkingDirectory option for running Chocolatey

gep13:
As discussed in GH-510

### DIFF
--- a/src/Cake.Common/Tools/Chocolatey/Pack/ChocolateyPackSettings.cs
+++ b/src/Cake.Common/Tools/Chocolatey/Pack/ChocolateyPackSettings.cs
@@ -187,5 +187,10 @@ namespace Cake.Common.Tools.Chocolatey.Pack
         /// </summary>
         /// <value>The allow unofficial flag</value>
         public bool AllowUnofficial { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


gep13:
@patriksvensson is this the implementation that you were suggesting?

patriksvensson:
@gep13 Yes, but I think it might be better to call it `OutputPath` (or what the corresponding name is for NuGet). This will make it an easy change if/when Chocolatey adds this option.

patriksvensson:
@gep13 Did it work?

gep13:
>@patriksvensson said...
Did it work?

I didn't actually test it yet, wanted to make sure I was on the right path first.  Also, just away to sign off for the night, so will test over the weekend.

gep13:
@patriksvensson I have fixed this up as you suggested.  I have also verified that it is working by changing our own build.cake file to use that settings, rather than the previous MoveDirectory that was being used.

Think this is ready for merging, assuming CI passes.

gep13:
Well shite!

Of course that isn't going to work :-(

gep13:
@patriksvensson I have taken back out the modification of the build.cake file.  Once this change is merged in, I will add it back in as a separate PR.

Issue #522, opened 2015-11-14
Need ability to specify which MSBuild platform to use.


marcosnz:
For ny build scenario I have to build an "Any CPU" target on a.64bit server using the x86  version of MSBuild.
Currently cake uses the build target to determine which MSBuild version to use. In my case it chooses 64bit because the target is"Any CPU" and running on 64 bit.
Need to be able to specify x86 MSBuild in this case.

Suggested solution:
Add an MSBuildTarget property to MSBuildSettings (similar to PlatformTarget) which has three values: Auto (use default/current logic), x86 and x64. This would be passed to the MSBuildResolver.
A SetMSBuildTarget extension method will be added to MSBuildSettingsExtensions to enable fluent api.

PR #523, opened 2015-11-15
Add and use MSBuildTarget as opposed to PlatformTarget

marcosnz:
Issue #522

### DIFF
--- a/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
+++ b/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
@@ -65,7 +65,7 @@ namespace Cake.Common.Tools.MSBuild
                 builder.Append(string.Concat("/p:Configuration=", configuration));
             }
 
-            // Buid for a specific platform?
+            // Build for a specific platform?
             if (settings.PlatformTarget.HasValue)
             {
                 var platform = settings.PlatformTarget.Value;
@@ -109,6 +109,8 @@ namespace Cake.Common.Tools.MSBuild
                     return "x86";
                 case PlatformTarget.x64:
                     return "x64";
+                case PlatformTarget.ARM:
+                    return "arm";
                 default:
                     throw new InvalidEnumArgumentException("platform", (int)platform, typeof(PlatformTarget));
             }
@@ -173,7 +175,23 @@ namespace Cake.Common.Tools.MSBuild
                 throw new ArgumentNullException("settings");
             }
 
-            var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion, settings.PlatformTarget ?? PlatformTarget.MSIL);
+            MSBuildPlatform buildPlatform = settings.MSBuildPlatform;
+
+            // If we haven't explicitly set an MSBuild target then use the Platform Target
+            if (buildPlatform == MSBuildPlatform.Automatic)
+            {
+                switch (settings.PlatformTarget)
+                {
+                    case PlatformTarget.x86:
+                        buildPlatform = MSBuildPlatform.x86;
+                        break;
+                    case PlatformTarget.x64:
+                        buildPlatform = MSBuildPlatform.x64;
+                        break;
+                }
+            }
+
+            var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion, buildPlatform);
             if (path != null)
             {
                 return new[] { path };
--- a/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs
+++ b/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs
@@ -38,6 +38,12 @@ namespace Cake.Common.Tools.MSBuild
         public PlatformTarget? PlatformTarget { get; set; }
 
         /// <summary>
+        /// Gets or sets the MSBuild platform.
+        /// </summary>
+        /// <value>The MSBuild platform.</value>
+        public MSBuildPlatform MSBuildPlatform { get; set; }
+
+        /// <summary>
         /// Gets or sets the tool version.
         /// </summary>
         /// <value>The tool version.</value>
@@ -80,6 +86,7 @@ namespace Cake.Common.Tools.MSBuild
             ToolVersion = MSBuildToolVersion.Default;
             Configuration = string.Empty;
             Verbosity = Verbosity.Normal;
+            MSBuildPlatform = MSBuildPlatform.Automatic;
         }
     }
 }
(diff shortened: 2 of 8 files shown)


patriksvensson:
@marcosnz 

We actually want to take `PlatformTarget` into consideration as well.

I was thinking that we introduce a `MSBuildPlatform` that we check.
This so we can infer what architecture of MSBuild to use simply from the 
platform that should be built. If there's a special case (such as the Silverlight one), 
the user can override by specifying a specific `MSBuildPlatform`.

```csharp
public enum MSBuildPlatform
{
  // Autodetect platform from platfgorm target and environment
  Auto = 0,
  // Explicitly use x86
  x86 = 1,
  // Explicitly use x64
  x64 = 2
}
```

```csharp
private static DirectoryPath GetVisualStudioPath(ICakeEnvironment environment, PlatformTarget target, MSBuildPlatform platform, string version)
{
    // Get the bin path (default to x86).
    var programFilesPath = environment.GetSpecialPath(SpecialPath.ProgramFilesX86);
    var binPath = programFilesPath.Combine(string.Concat("MSBuild/", version, "/Bin"));

    if(platform == MSBuildPlatform.Auto)
    {
      // We want the platform target to decide
      // what MSBuild architecture to use.
      
      if (target == PlatformTarget.MSIL)
      {
          if (environment.Is64BitOperativeSystem())
          {
              binPath = binPath.Combine("amd64");
          }
      }
      if (target == PlatformTarget.x64)
      {
          binPath = binPath.Combine("amd64");
      }
    }
    else
    {
      // We want to use a specific MSBuild architecture.
      
      if (platform == MSBuildPlatform.x64)
      {
          binPath = binPath.Combine("amd64");
      }
    }
    
    return binPath;
}
```

@gep13 @devlead Any more thoughts about this?

marcosnz:
What about we set the MSBuildTarget based on PlatformTarget before passing to MSBuildReolver?
{code}
            MSBuildTarget target = settings.MSBuildTarget ?? MSBuildTarget.Default;

            if (target == MSBuildTarget.Default)
            {
                switch (settings.PlatformTarget)
                {
                    case PlatformTarget.x86:
                        target = MSBuildTarget.x86;
                        break;
                    case PlatformTarget.x64:
                        target = MSBuildTarget.x64;
                        break;
                }
            }

         var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion, target);
  {code}

src/Cake.Common/Tools/MSBuild/MSBuildTarget.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Cake.Common.Tools.MSBuild
+{
+    /// <summary>
+    /// Represents an MSBuild exe target.
+    /// </summary>
+    public enum MSBuildTarget
```
patriksvensson: Should be renamed to `MSBuildPlatform` to avoid confusion with [MSBuild targets](https://msdn.microsoft.com/en-us/library/ms171462.aspx). Will also follow Team Builds approach to this.

src/Cake.Common/Tools/MSBuild/MSBuildTarget.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Will build using MSBuild version based on PlatformTarget/Host OS.
+        /// </summary>
+        Default = 0,
```
patriksvensson: Would prefer this to be called `Automatic`.

src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs
```diff
@@ -38,6 +38,12 @@ public ISet<string> Targets
         public PlatformTarget? PlatformTarget { get; set; }
 
         /// <summary>
+        /// Gets or sets the MSBuild target.
+        /// </summary>
+        /// <value>The MSBuild target.</value>
+        public MSBuildTarget? MSBuildTarget { get; set; }
```
patriksvensson: Should not be nullable and should be explicitly set to `Automatic` in the constructor.

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -173,7 +173,23 @@ protected override IEnumerable<FilePath> GetAlternativeToolPaths(MSBuildSettings
                 throw new ArgumentNullException("settings");
             }
 
-            var path = MSBuildResolver.GetMSBuildPath(_fileSystem, _environment, settings.ToolVersion, settings.PlatformTarget ?? PlatformTarget.MSIL);
+            MSBuildTarget target = settings.MSBuildTarget ?? MSBuildTarget.Default;
```
patriksvensson: Since `MSBuildTarget` shouldn't be nullable, there is no need for null coalescing here.

patriksvensson:
@marcosnz Yes, that would work. Left you some comments as well. :+1:

marcosnz:
Thanks @patriksvensson! I have incorporated your comments. How does it look now?

patriksvensson:
@marcosnz I realized that the tests should not be changed. The behavior should be the same after your change so the tests should remain what they looked like before your changes. To test your changes, we should add new tests that test the combination of platform target and msbuild platform.

Other than that it looks good to me :+1:

marcosnz:
@patriksvensson - of course! I have added back original tests, kept in my change tests that test the new property and added a couple of tests that have both properties set to ensure we are getting the results we expect there.

Redth:
To throw another variable into the mix, I believe that Windows Phone 81 projects that you want to compile for ARM must use the x86 msbuild.exe...   So you'd need to invoke it something like 

`...\x86\msbuild.exe /p:Configuration=Release /p:Platform=ARM /target:Build "SomeWp81ARM.sln"`

Maybe we need `MSBuildPlatformTarget.Arm`?

patriksvensson:
@Redth I agree ARM support sounds like a good idea but we should probably add an issue for it and do that on a separate PR and not in this one.

As a workaround, you can send a custom platform by doing this:

```csharp
MSBuild("./src/Cake.sln", new MSBuildSettings()
    .WithProperty("platform", "ARM")
```

marcosnz:
@Redth @patriksvensson I think that ARM would be added to PlatformTarget enum, not MSBuildPlatform?

Redth:
@marcosnz yes i do think that's correct now, the only thing we need to ensure is that the logic of which msbuild.exe to select can allow a combination of: `MSBuildPlatform.x86 + PlatformTarget.Arm`.

Redth:
@patriksvensson currently your suggested workaround won't work since for whatever reason microsoft requires the windows phone projects to be built with the x86 msbuild.exe when specifying `/p:platform=arm`, and right now the amd64 gets picked if running on a 64 bit system and PlatformTarget.x86 is not selected.

marcosnz:
@Redth @patriksvensson I have added support for ARM. I added it into this PR because of the use-case's dependency of the MSBuildPlatform. I added unit tests and if the MSBuildPlatform is automatic the ARM uses same logic as MSIL for choosing the version of MSBuild. I added a test case specifically for the `MSBuildPlatform.x86 + PlatformTarget.Arm` combination

patriksvensson:
Ok, this looks good to me.

Can you rebase to one commit (and after that against develop) and I'll merge it :+1:

marcosnz:
Umm... I think I did it? lol. My GIt knowledge is weak...

patriksvensson:
:+1: Perfect! :smile:

@devlead @gep13 Anyone of you have time to do a final review and merge this if you're happy with it? Otherwise I can do it in a couple of hours.

patriksvensson:
Merged :+1: 

I rebases the PR and updated the commit message to describe the changess in more details, that's why it doesn't appear as merged. Sorry about this, but didn't want to bother you (again) for a rebase.

Thank you very much for this! :smile:

patriksvensson:
Commit can be seen here: https://github.com/cake-build/cake/commit/144f6aa14e24ab1eb11499da4404dfb48c73d93b

PR #536, opened 2015-11-26
Use new OutputDirectory Setting for Chocolatey Pack

gep13:
Relates to GH-505


gep13:
@patriksvensson does the build process not grab the latest version of Cake from MyGet?  #lazydeveloper

patriksvensson:
@gep13 No, it grabs the latest official version at the moment. Should add separate CI  for this.

gep13:
Related to this, do we want to add the environment variables for Chocolatey/Nuget into the AppVeyor yaml file, so that we can push the packages out automatically when the master branch is being built?

patriksvensson:
We can add the tasks, but I think it could be a manual step like NuGet for
the moment.

PR #539, opened 2015-11-27
Add ability to create ReleaseNotes using GitReleaseManager

gep13:


gep13:
@patriksvensson @devlead any chance either of you can take a look at this?

build.cake
```diff
@@ -311,6 +311,24 @@ Task("Publish-Chocolatey")
...
 });
 
+Task("Create-Release-Notes")
+  .Does(() =>
+{
+    var userName = EnvironmentVariable("CAKE_GITHUB_USERNAME");
+    var password = EnvironmentVariable("CAKE_GITHUB_PASSWORD");
+    var owner = EnvironmentVariable("CAKE_GITHUB_OWNER");
+    var repo = EnvironmentVariable("CAKE_GITHUB_REPO");
+    var targetBranch = EnvironmentVariable("CAKE_TARGET_BRANCH");
```
patriksvensson: What happens if this isn't set for some reason? Shouldn't this always be `main`?

gep13: In this case, I guess it would yes.  I think I was thinking with my generic build script hat on, i.e. I want to re-use the same build.cake file across multiple repositories, but in this case, that doesn't make sense.  I will fix up.

gep13: And I have a failing test :-D

gep13: @patriksvensson fixed up.

PR #550, opened 2015-12-03
Added fix for framework name on Mono.

### LINKED ISSUE #549: CakeEnvironment.GetTargetFramework throws on Mono
For some reason [AppDomainSetup.TargetFrameworkName](https://github.com/mono/mono/blob/b7a308f660de8174b64697a422abfc7315d07b8c/mcs/class/corlib/System/AppDomainSetup.cs#L300) returns null on Mono which causes [CakeEnvironment.GetTargetFramework](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/CakeEnvironment.cs#L130) to throw.

We should return a default framework, preferably .NET 4.5 if we can not resolve the current framework as we're supposed to. We might need to find a more permanent fix in the future though.

patriksvensson:
Closes #549

### DIFF
--- a/src/Cake.Core/CakeEnvironment.cs
+++ b/src/Cake.Core/CakeEnvironment.cs
@@ -116,9 +116,9 @@ namespace Cake.Core
             return Environment.GetEnvironmentVariables()
                 .Cast<System.Collections.DictionaryEntry>()
                 .ToDictionary(
-                key => (string)key.Key,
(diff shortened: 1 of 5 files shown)


patriksvensson:
The fix have been validated on LInux (Ubuntu Server 15.04) using Mono 4.0.5.

PR #560, opened 2015-12-07
Disable Quoting of process filename on unix

### LINKED ISSUE #559: Quoted process fails on unix
0.6.3 introduced issue on Unix systems.

devlead:
Fixes #559

### DIFF
--- a/src/Cake.Core/IO/ProcessRunner.cs
+++ b/src/Cake.Core/IO/ProcessRunner.cs
@@ -48,13 +48,16 @@ namespace Cake.Core.IO
                 throw new ArgumentNullException("settings");
             }
 
+            // Get the fileName
+            var fileName = _environment.IsUnix() ? filePath.FullPath : filePath.FullPath.Quote();
+
             // Get the arguments.
             var arguments = settings.Arguments ?? new ProcessArgumentBuilder();
 
             if (!settings.Silent)
             {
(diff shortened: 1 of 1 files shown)


patriksvensson:
Great! :+1: 

Can we also add unit tests for this as well [here](https://github.com/cake-build/cake/blob/develop/src/Cake.Core.Tests/Unit/IO/ProcessRunnerTests.cs)?

src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -48,13 +48,16 @@ public IProcess Start(FilePath filePath, ProcessSettings settings)
                 throw new ArgumentNullException("settings");
             }
 
+            // Get the fileName
+            var fileName = _environment.IsUnix() ? filePath.FullPath : filePath.FullPath.Quote();
```
patriksvensson: This variable name confused me a little since it's the whole path that this contains and not just the file name. Maybe we can change this to `fullFileName` or something that shows that this is the full path to the file?

devlead: Same name as ProcessStartInfo constructor parameter. Suggest we keep.

patriksvensson:
@devlead I read the code again and no tests are necessary for this one. It would require an integration test so let's just verify that it works for now.

PR #578, opened 2015-12-10
[HOLD] Forward all arguments to the cake.exe

eriklarko:
The way the build.sh works now does not allow arguments to be passed to your cake script. This is a simple fix for this problem. If the argument parsing in the script is important then this fix won't work :)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,27 +10,6 @@ TOOLS_DIR=$SCRIPT_DIR/tools
 NUGET_EXE=$TOOLS_DIR/nuget.exe
 CAKE_EXE=$TOOLS_DIR/Cake/Cake.exe
 
-# Define default arguments.
-SCRIPT="build.cake"
-TARGET="Travis"
-CONFIGURATION="Release"
-VERBOSITY="verbose"
-DRYRUN=false
-SHOW_VERSION=false
-
-# Parse arguments.
-for i in "$@"; do
-    case $1 in
-        -s|--script) SCRIPT="$2"; shift ;;
-        -t|--target) TARGET="$2"; shift ;;
-        -c|--configuration) CONFIGURATION="$2"; shift ;;
-        -v|--verbosity) VERBOSITY="$2"; shift ;;
-        -d|--dryrun) DRYRUN=true ;;
-        --version) SHOW_VERSION=true ;;
-    esac
-    shift
-done
-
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -10,42 +10,6 @@ TOOLS_DIR=$SCRIPT_DIR/tools
...
-        -c|--configuration) CONFIGURATION="$2"; shift ;;
-        -v|--verbosity) VERBOSITY="$2"; shift ;;
-        -d|--dryrun) DRYRUN=true ;;
-        --version) SHOW_VERSION=true ;;
-    esac
-    shift
-done
-
-# Make sure the tools folder exist.
-if [ ! -d $TOOLS_DIR ]; then
```
devlead: Why is this removed?

eriklarko: The tools things should not have been removed, I was over-eager on the dd :) I've uploaded a new patch

build.sh
```diff
@@ -10,27 +10,6 @@ TOOLS_DIR=$SCRIPT_DIR/tools
 NUGET_EXE=$TOOLS_DIR/nuget.exe
 CAKE_EXE=$TOOLS_DIR/Cake/Cake.exe
 
-# Define default arguments.
-SCRIPT="build.cake"
```
devlead: One side effect of this is that default values are lost.

eriklarko: Hmm, I thought that the cake.exe defaulted as well. If I run
```
$ mono tools/Cake/Cake.exe
```
it defaults to using build.cake


Are these defaults not the same as the ones in cake.exe?

patriksvensson: We use different defaults when building Cake. We use `verbose` verbosity for example, while the regular verbosity is `normal`.

eriklarko: Aaaah, I confused this file with the one in the example repo, https://github.com/cake-build/example/blob/master/build.sh. Sorry! Should I move the PR to https://github.com/cake-build/example?

patriksvensson:
Copied from [Gitter](https://gitter.im/cake-build/cake?at=5669e9e6cffd648a0554de5c):

> Yes, absolutely, you should be able to append $@ (like you did), but we will need to fix #474 and #579 to make sure we keep compatibility with the behavior for people who upgrade their existing scripts before we merge it.

devlead:
Closing due inactivity.

PR #584, opened 2015-12-12
Support arbitrary text when parsing AssemblyInformationalVersion.

MatthewKing:
AssemblyInformationalVersionAttribute itself can accept arbitrary text strings, unlike AssemblyVersionAttribute / AssemblyFileVersionAttribute. However, the existing AssemblyInfoParser will only parse values from AssemblyInformationalVersionAttribute that conform to the same restrictions that AssemblyVersionAttribute / AssemblyFileVersionAttribute are subject to.

A real-world situation where this will cause issues is when using semantic versioning with pre-release or build metadata, such as "1.0.0-beta1" or "1.2.3-rc1". These values should be perfectly valid in AssemblyInformationalVersion, and should be accepted by AssemblyInfoParser.

This pull request adds support for arbitrary text strings when parsing AssemblyInformationalVersionAttribute using AssemblyInfoParser.


patriksvensson:
Looks good to me :+1: 
Thanks for fixing this!

gep13:
@MatthewKing this was an oversight on my part!  Thanks for catching and fixing this, much appreciated! :+1:

PR #588, opened 2015-12-15
Added support for adding custom arguments to tools.

### LINKED ISSUE #500: Workarounds for incomplete tool settings
One of the difficulties I've run into with Cake is when using tools, often the tool author has not included support for all options/switches that are available for command line execution.

When I want to use such a switch, I either have to invoke the tool using a raw process, or I have to submit a PR to support the switch and wait until the next release to use it.

Perhaps we can build in a way to allow devs to include raw command line args (in addition to using the supported settings properties), so we can more easily work around missing settings until they get added.

I'm thinking of something along the lines of what @patriksvensson did in #470, where he added a base settings class with the tool path.  In this case, I'm thinking maybe we add a property to that base class for raw command line args, and somehow force the tool runner to include the raw args.  

I'm sure others can think of better solutions as well.

Ideas? Thoughts?

patriksvensson:
Closes #500

### DIFF
--- a/src/Cake.Core/Tooling/ToolSettings.cs
+++ b/src/Cake.Core/Tooling/ToolSettings.cs
@@ -1,3 +1,4 @@
+using System;
 using Cake.Core.IO;
 
 namespace Cake.Core.Tooling
@@ -12,5 +13,11 @@ namespace Cake.Core.Tooling
         /// </summary>
         /// <value>The tool path.</value>
         public FilePath ToolPath { get; set; }
+
+        /// <summary>
+        /// Gets or sets the argument customization.
+        /// </summary>
+        /// <value>The arguments.</value>
+        public Func<ProcessArgumentBuilder, ProcessArgumentBuilder> ArgumentCustomization { get; set; }
     }
 }
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -60,8 +60,6 @@
     <Compile Include="Fixtures\Build\AppVeyorInfoFixture.cs" />
     <Compile Include="Fixtures\AssemblyInfoFixture.cs" />
     <Compile Include="Fixtures\AssemblyInfoParserFixture.cs" />
-    <Compile Include="Fixtures\ToolFixture.cs" />
-    <Compile Include="Fixtures\ToolFixtureResult.cs" />
     <Compile Include="Fixtures\Tools\CakeRunnerFixture.cs" />
     <Compile Include="Fixtures\IO\FileCopyFixture.cs" />
     <Compile Include="Fixtures\IO\FileDeleteFixture.cs" />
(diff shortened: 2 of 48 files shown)


gep13:
@patriksvensson there seems to be something screwy going on here:

image

with the names of the files in this PR.

Any ideas?

patriksvensson:
@gep13 It's just a convention to name files containing generic classes using the CRef naming (i.e. `T:TestNamespace.GenericClass`1`). We hadn't had the need for it before since there are no classes with the same name that only differs by generic argument count. [CoreFX](https://github.com/dotnet/corefx/tree/master/src/System.Collections.Immutable/src/System/Collections/Immutable) for example uses the same conventions.

Can change it to something else if you want to, but I think we should have some convention for it.

gep13:
@patriksvensson Gotcha!  No, no need to change it.  I just hadn't seen this before.  My initial thought was that something had gone wrong with a merge/rebase etc, and the file name had got mangled :smile: Makes sense now that you explain it.  Good to know, thanks!

src/Cake.Core/Tooling/ToolSettings.cs
```diff
@@ -12,5 +13,11 @@ public class ToolSettings
         /// </summary>
         /// <value>The tool path.</value>
         public FilePath ToolPath { get; set; }
+
+        /// <summary>
+        /// Gets or sets the arguments.
+        /// </summary>
+        /// <value>The arguments.</value>
+        public Action<ProcessArgumentBuilder> ArgumentCustomization { get; set; }
```
devlead: Wonder if it would make sense to have it be a Func instead if Action instead 
```charp
public Func<ProcessArgumentBuilder, ProcessArgumentBuilder> ArgumentCustomization { get; set; }
```
so we could modify and return same instance, with the extension methods:
```csharp
Settings.ArgumentCustomization = arg=>arg.Append("-true");
```
,but also just as easily return a new instance/exsisting instance. 
```csharp
Settings.ArgumentCustomization = arg=>"my new arguments";
...
Settings.ArgumentCustomization = arg=>reUseExistingArguments;
...
Settings.ArgumentCustomization = arg=>GetCustomNewArgs(param1, param2);
```

patriksvensson: We could add an extension method for this, but just providing an `Action` is probably simpler and fits the most common use case where you want to add an argument that is missing.

If you completely want to rewrite the arguments, you could simply use the `builder.Clear()` method before adding your arguments.

devlead: I kinda disagree, but might be personal preference.
Personally don't think this is unclean, unclear or less fluent.
```csharp
Settings.ArgumentCustomization = builder=>"arg"
```
vs.
```csharp
Settings.ArgumentCustomization = builder => {  
    builder.Clear();
    builder.Append("Arg");
};
```
As the all the extension methods return a ProcessArgumentBuilder it would just work without any code change.

I would just find a `Func` more flexible with little added complexity, would support any scenario the Action does plus a few more.

patriksvensson:
I think this one is ready to be merged :+1:

PR #589, opened 2015-12-16
Add invariant IFormatterProvider to avoid issues based on user locale

devlead:
Code analytics raised CA1305, this PR fixes the potential user locale issue by:
1. Adds CultureInfo.InvariantCulture for locale insensitive "ToString"
2. string.Format handles null so removed that assignment

### DIFF
--- a/src/Cake/Arguments/VerbosityParser.cs
+++ b/src/Cake/Arguments/VerbosityParser.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using Cake.Core.Diagnostics;
 
 namespace Cake.Arguments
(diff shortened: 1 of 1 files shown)


gep13:
@devlead is there a reason to have the string Format broken out into a seperate const?

If not, if we are in this file making changes anyway, why not combine it into a single line?

devlead:
@gep13 nothing more than readability I guess, tried to change as little as possible in the existing code.

patriksvensson:
@gep13 Yeah, it's for readability. Don't like too long lines and it's a const so the compiler will merge it :smile:

gep13:
@patriksvensson @devlead gotcha.  no worries, just thought I would ask.  Having to mind switch between two ways of doing things, as another project, who shall rename nameless, works slightly differently :smile:

patriksvensson:
Looks good to me. Rebase against develop and I'll merge it :+1:

devlead:
@patriksvensson Rebased.

PR #605, opened 2015-12-25
Forward arguments to Cake from the bootstrapper scripts

jamesqo:
Hey guys! I came across your project while browsing the Reddit the other day, thought it was pretty cool.

This PR forwards the arguments from build.ps1 and build.sh to the Cake script, so it's possible to communicate with the script.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -8,7 +8,9 @@ Param(
     [switch]$Experimental,
     [switch]$WhatIf,
     [switch]$Mono,
-    [switch]$SkipToolPackageRestore
+    [switch]$SkipToolPackageRestore,
+    [Parameter(ValueFromRemainingArguments = $true)]
+    [string[]]$ScriptArgs
 )
 
 $PSScriptRoot = split-path -parent $MyInvocation.MyCommand.Definition;
@@ -63,5 +65,5 @@ if (!(Test-Path $CAKE_EXE)) {
 }
 
 # Start Cake
-Invoke-Expression "$CAKE_EXE `"$Script`" -target=`"$Target`" -configuration=`"$Configuration`" -verbosity=`"$Verbosity`" $UseMono $UseDryRun $UseExperimental"
+Invoke-Expression "$CAKE_EXE `"$Script`" -target=`"$Target`" -configuration=`"$Configuration`" -verbosity=`"$Verbosity`" $UseMono $UseDryRun $UseExperimental $ScriptArgs"
 exit $LASTEXITCODE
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -8,7 +8,9 @@ Param(
     [switch]$Experimental,
     [switch]$WhatIf,
     [switch]$Mono,
-    [switch]$SkipToolPackageRestore
+    [switch]$SkipToolPackageRestore,
+    [Parameter(ValueFromRemainingArguments = $true)]
```
patriksvensson: Cool, I didn't know about this. Great addition!

build.sh
```diff
@@ -25,8 +26,10 @@ for i in "$@"; do
         -t|--target) TARGET="$2"; shift ;;
         -c|--configuration) CONFIGURATION="$2"; shift ;;
         -v|--verbosity) VERBOSITY="$2"; shift ;;
-        -d|--dryrun) DRYRUN=true ;;
+        -d|--dryrun) DRYRUN="-dryrun" ;;
         --version) SHOW_VERSION=true ;;
+        --) shift; SCRIPT_ARGUMENTS+=("$@"); break ;;
```
patriksvensson: My bash script knowledge is not the best. Does this take all arguments after a `--` as they are?

patriksvensson:
@jamesqo Thanks! :smile:
As far as I can see this should not break any existing builds, right?

(Ping @gep13 @devlead)

build.sh
jamesqo: @patriksvensson Yep, `"$@"` should preserve the separation between the arguments. Something like `1 2 3` should get forwarded as `1 2 3` and not `"1 2 3"`.

jamesqo:
This is probably irrelevant to the original commit, but I've also added an `exec` statement at the end of the Bash script. It should increase performance since it avoids spawning a subshell and executes the program in-place.

jamesqo:
@patriksvensson No, it shouldn't break with the current bootstrapper.

hal-ler:
Any plans for this PR? It would be great as a means to offer alternative build options (such as custom install paths).

devlead:
@hal-ler we've just missed to review this.
@jamesqo Could you rebase against latest develop, squash commits into 1 and ping us when that's done and we'll review it for merging.

jamesqo:
@devlead Done!

devlead:
@jamesqo merged, thank for your contribution :+1:

devlead:
@jamesqo if you have the opportunity, please look at doing similar PR for the default bootstrappers
https://github.com/cake-build/bootstrapper/tree/master/res/scripts

jamesqo:
@devlead Sure, no problem! I'll submit one right now in fact, give me a minute.

PR #608, opened 2015-12-28
Move from WebClient to HttpClient

khellang:
I noticed in https://github.com/dotnet/cli/issues/270#issuecomment-167539366 that `WebClient` is used and that it's not part of the .NET Core surface area.

This moves Cake.Common from 96.39% to 97.50% :stuck_out_tongue_closed_eyes:


devlead:
Merged, thanks for your contribution :+1:

PR #609, opened 2015-12-28
Added Win32 platform target.

### LINKED ISSUE #604: PlatformTarget is missing Win32
When building C++ projects, the `x86` platform target is named `Win32`. This target is missing (or intentionally left out) from the `PlatformTarget` enum. The 64-bit target is still named `x64` though.

Quick fix is to use `WithProperty("Platform", "Win32")`.

patriksvensson:
Closes #604


gep13:
@patriksvensson do you need to include this additional PlatformTarget in some of the other tests?  For example `Should_Get_Correct_Path_To_MSBuild_Version_4`  In that one, you seem to be testing all the combination of targets, except the new one.

patriksvensson:
@gep13 Yes, I was thinking about it as well. I will add these tests for consistency asap, but in my opinion, these tests should be rewritten or removed since a lot of them are redudant.

gep13:
@patriksvensson that thought had occurred to me as well :+1:

patriksvensson:
Updated the PR with more tests :smile:

gep13:
Looks good to me :smile:

Issue #606, opened 2015-12-28
Consider porting cake to coreclr/corefx


shahid-pk:
One benefit will be that this will make cake portable without requiring global .net or mono. Another benefit is since ms uses mono as a prove of concept and not as a supported target , ms projects that would have used cake are not using it like https://github.com/dotnet/cli/issues/270

Even if we don't consider the above mention scenario still it will be useful to target coreclr which looks like to be the .net target of the future.

devlead:
@shahid-pk agree long term CoreCLR makes perfect sense, added some comments to 
[dotnet/cli#270](https://github.com/dotnet/cli/issues/270#issuecomment-167539366)

Main hurdle is CoreCLR isn't currently an supported platform by the _Microsoft .NET Compiler Platform ("Roslyn") CSharp scripting package_ , and that's kinda a deal breaker :wink:

Cake's already 96,54% CoreCLR compatible, so it's more _"refactor tweaks"_ than a rewrite once we get the dependencies in place.

If we were to remove some dependencies more code would likely have to be rewritten or replaced by other dependencies, but that might make perfect sense long term once we analyzed more deeply what's involved in a CoreCLR port.

I might be wrong but reading issues like NuGet/Home#1870 by @ferventcoder makes me think CoreCLR / vNext still is a bit of a moving target, needing some stabilization before 3rd party easily can take a dependency.
That said the _promise_ of CoreCLR really makes it an attractive target to port to and might totally justify some early adopter pain :wink:

ferventcoder:
How does one determine compat? I should go RTFM somewhere.

devlead:
@ferventcoder  The .Net portability analyzer Visual Studio extension is a good starting point.
https://visualstudiogallery.msdn.microsoft.com/1177943e-cfb7-4822-a8a6-e56c7905292b

gep13:
This issue has been superseded by this one: https://github.com/cake-build/cake/issues/1015

Going to close this one.

PR #614, opened 2015-12-30
Move Cake.Common towards a more .NET Core-friendly surface area

khellang:
This brings Cake.Common up from 97.50% to 98.68% :sparkles: The rest needs #ifdefs :disappointed:


devlead:
Relates to #606

devlead:
@khellang merged, thanks for contributing :+1:

PR #617, opened 2016-01-02
Make #tool and #addin package manager agnostic.

### LINKED ISSUE #616: Make #tool and #addin package manager agnostic
As we discussed before it would be nice if the `#tool` and `#addin` pre-processor directives where package manager agnostic.

We could use a URI for this and provide handlers per scheme for the actual implementation. If the package name isn't a valid URI, it would fall back to use NuGet for backwards compatibility.

```
scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
```

We will only have support for NuGet for the time being, but once we get the infrastructure in place we wouldn't be locked in to use NuGet packages only. We could also (in the future) provide a mechanism to register custom protocol handlers if user's have their own way of retrieving tools/addins.

```
// Old syntax (kept for backwards compatibility)
#tool "Cake.Foo"
#addin "Cake.Bar"

// New syntax
#tool "nuget://Cake.Foo/?version=1.2&source=https://myget.org/lol"
#addin "npm://Cake.Bar"
#tool "homebrew://Cake.Baz"
#addin "chocolatey://Cake.Qux/?version=0.1"
```

Using an URI scheme could in the future be supported for the `#r` and `#l` directives as well to support gists or similar services, but that is another issue.

patriksvensson:
This is a new way of referencing packages. This so we
can allow the usage of other package managers such as npm and
homebrew to retrieve packages in the future.

We still support the old #addin/#tool directive, but the new
syntax for it is more generic and easy to expand. We will
however need to update the documentation for this asap.

There are some breaking changes though. I've moved all the
NuGet stuff to it's own assembly (except INuGetToolResolver)
since those aren't really of interest for consumers of
Cake.Core. Keeping this isolated will potentially make it
easier when migrating to CoreCLR in the future as well.

nuget:?package=Cake.Foo
nuget:https://myget.org/f/Cake/?package=Cake.Foo&Version=1.2.3

Closes #616


patriksvensson:
After a lot of reading tonight, I've made some changes to the URI logic to use an URN for the package manager "scheme" (still a 100% valid URI). This is a little more explicit, but more flexible and will suit most usecases I can see. Done the initial work and will update PR tomorrow.

**We will expect information like this:**

```
packetmanager:[scheme:[//[user:password@]host[:port]][/]]?package=id[&query]
```

**Example:**

```
nuget:?package=Cake.Foo&version=1.2.3
nuget:https://myget.org/f/Cake/?package=Cake.Foo&version=1.2.3
nuget:https://user:pass@myget.org/f/Cake/?package=Cake.Foo&version=1.2.3
nuget:https://user:pass@myget.org/f/Cake:1234/?package=Cake.Foo&version=1.2.3
```

patriksvensson:
Still WIP. Just naming left and some minor changes.

src/Cake.Core/IO/NuGet/NuGetPackageInstaller.cs
```diff
@@ -31,8 +31,8 @@ public void InstallPackage(NuGetPackage package, DirectoryPath installationRoot)
             var process = _processRunner.Start(nugetPath, new ProcessSettings
             {
                 Arguments = GetNuGetToolInstallArguments(package, installationRoot),
-                RedirectStandardOutput = true,
-                Silent = true
+                RedirectStandardOutput = false,
+                Silent = false
```
gep13: @patriksvensson why did this change?

patriksvensson: @gep13 That change was unintentional. Was outputting NuGet information
while testing. Will change back. Thanks for seeing this :smile:

patriksvensson:
Ok, this is ready for merge if you don't have any objections.

devlead:
@patriksvensson I'll take i final look, and merge if ok.

gep13:
@patriksvensson just thinking about this one... Is there updates to the documentation on the website required so that people can start to use this when we pull it in?

devlead:
@gep13 yes preprocessor/directive part would need some love

patriksvensson:
Ok, I've added the suggestions and rebased against develop.

gep13:
Looks good to me :+1:

PR #629, opened 2016-01-09
Use gobbler to locate tools from PATH environment variable

SharpeRAD:
Allows you to set the PATH environment variable to a shared tools folder and still locate tools with non-standard locations (eg: "tools/JetBrains.ReSharper.CommandLineTools/tools/inspectcode.exe").

```powershell
$ENV:PATH = "C:/Tools/"
```

Also upper-cased any instances of the PATH variable for consistency :smile:

### DIFF
--- a/src/Cake.Common/Tools/Chocolatey/ChocolateyToolResolver.cs
+++ b/src/Cake.Common/Tools/Chocolatey/ChocolateyToolResolver.cs
@@ -59,7 +59,7 @@ namespace Cake.Common.Tools.Chocolatey
             }
 
             // Last resort try path
-            var envPath = _environment.GetEnvironmentVariable("path");
+            var envPath = _environment.GetEnvironmentVariable("PATH");
             if (!string.IsNullOrWhiteSpace(envPath))
             {
                 var pathFile = envPath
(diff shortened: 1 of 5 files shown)


patriksvensson:
@SharpeRAD Thanks for the contribution :+1:

I'm not sure this is a good idea though. Doing a recursive search through all directories in the path will "touch" more than 9000 files on my system and would slow down the tool resolution a lot. I think it's better that you add the `C:/tools/JetBrains.ReSharper.CommandLineTools/tools/` folder to your path directly to achieve the expected behavior, or that we add a CAKE_PATH environment variable (or similar) that does this.

What do you think?

(ping @devlead @gep13)

gep13:
I agree, I think a configurable option to allow inspection of for tools would be better.  If there is nothing assigned to this variable, Cake just ignores it.

@SharpeRAD how are you adding the ReSharper Tools on your path?  Are you using the Chocolatey installation?  If so, it automatically adds a shim for the tools into the System Path Variable, so Cake can already use them from there.

SharpeRAD:
A tools variable would be better, the path variable was just there already so I piggy backed that to achieve a quick solution. I'm currently installing them through "NugetInstall" and a tools variable but it would be better if it worked with the #tool directive.

```C#
NuGetInstall("xunit.runner.console", new NuGetInstallSettings 
	{
		ExcludeVersion  = true,
		OutputDirectory = tools
    });
```

I noticed your rewriting the package support so I'll wait until that's committed before resubmitting.

gep13:
@SharpeRAD ah, I see.  You could use ChocolateyInstall method instead, and then it should start working for you, without the need for this change.  That will required that Chocolatey is installed on the machine that you are running on though.

Issue #601, opened 2015-12-22
How to get ILRepack executable?


AnderssonPeter:
I just tried to run the ILRepack executable but its giving me the error `ILRepack: Could not locate executable.` my first thought was that it could not find the `primaryAssembly` so I tested different variations for about an hour, but my current guess is that it can't find the ILRepack under tools? but how do I get it there?

Note: I get the exact same error if I try `ILMerge` and I have installed ILMerge on my computer?

devlead:
You can fetch ILRepack via the [#tool](http://cakebuild.net/docs/fundamentals/preprocessor-directives) directive which fetches it from Nuget.

I.e.
```cake
#tool "ILRepack"

```

devlead:
You can read more about tool resolution here:
http://cakebuild.net/docs/tools/tool-resolution

Petoj87:
Thanks that solved my problem, maybe it would be a good idea to add this information to http://cakebuild.net/dsl/ilrepack and http://cakebuild.net/dsl/ilmerge ?

patriksvensson:
You can read more about tool resolution here: http://cakebuild.net/docs/tools/tool-resolution

devlead:
Fixed by #632

Issue #491, opened 2015-11-05
What is the best approach for Summary links


gep13:
When updating the GitVersion documentation, I did this:

![image](https://cloud.githubusercontent.com/assets/1271146/10969185/93a6d3fa-83be-11e5-8cd8-c4009fdf64e2.png)

Thinking it would me line breaks between them, but I think I need to be more explicit, as the end result is this:

![image](https://cloud.githubusercontent.com/assets/1271146/10969156/65b3b49a-83be-11e5-8cd1-3ee37627bf91.png)

I think we need to decide on a standard format for the summary sections of each alias, and apply them throughout.

Thoughts?

devlead:
As I think @patriksvensson has suggested before the one way would be to support the [list](https://msdn.microsoft.com/en-us/library/y3ww3c7e.aspx) tag,

The way I've seen in standard XML docs is using the [para](https://msdn.microsoft.com/en-us/library/x640hcd2.aspx) tag, don't know if we support that yet though.
```csharp
    /// <summary>DoWork is a method in the TestClass class.
    /// <para>Here's how you could make a second paragraph in a description. <see cref="System.Console.WriteLine(System.String)"/> for information about output statements.</para>
    /// <seealso cref="TestClass.Main"/>
    /// </summary>
    public static void DoWork(int Int1)
    {
    }
```

patriksvensson:
@devlead @gep13 The `para` tag is supported.

Here is a [list of what is supported by the documentation tool](https://github.com/cake-build/website/tree/develop/src/Cake.Web.Docs/Xml/Parsers) at the moment. I think the only thing that is missing is lists. I can make a stab tonight to implement that as well.

patriksvensson:
And there's also an implementation of `para` rendering: https://github.com/cake-build/website/blob/develop/src/Cake.Web.Core/Rendering/CommentRenderer.cs#L85

devlead:
@patriksvensson then `para` should work just fine for @gep13 ;)

About `lists` think [highlight.js](https://highlightjs.org/) is more important so we can get [contributing/documentation](http://cakebuild.net/docs/contributing/documentation) looking ok ;)
Added issue cake-build/website#23 for this.

I will add supported tags to [contributing/documentation](http://cakebuild.net/docs/contributing/documentation) so we have that documented too.
Added issue cake-build/website#24 for this.

gep13:
This also looks like a duplicate of: https://github.com/cake-build/cake/issues/150 so going to close it.

[...]

PR #632, opened 2016-01-10
Added information to summary tag

gep13:
- About how to include the necessary tool to get the aliases working correctly

Relates to #601

### DIFF
--- a/src/Cake.Common/Tools/ILMerge/ILMergeAliases.cs
+++ b/src/Cake.Common/Tools/ILMerge/ILMergeAliases.cs
@@ -7,7 +7,13 @@ using Cake.Core.IO;
 namespace Cake.Common.Tools.ILMerge
 {
     /// <summary>
-    /// Contains functionality related to ILMerge.
+    /// <para>Contains functionality related to ILMerge.</para>
+    /// <para>
+    /// In order to use the commands for this alias, include the following in your build.cake file to download and install from NuGet.org, or specify the ToolPath within the ILMergeSettings class:
+    /// <code>
+    /// #tool "nuget:?package=ILRepack"
+    /// </code>
+    /// </para>
     /// </summary>
     [CakeAliasCategory("ILMerge")]
     public static class ILMergeAliases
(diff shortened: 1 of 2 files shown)


gep13:
@devlead @patriksvensson what do you think about the following changes?  Good? Bad?

gep13:
This is related to this as well: https://github.com/cake-build/cake/issues/491

If we can decide what we want to achieve, then I can go back and update all the summary tags.

gep13:
@devlead @patriksvensson still looking for some thoughts on how to proceed with this.  Can I get your thought in this issue, and I can update PR. Thanks!

devlead:
I think we should change  `#tool "ILRepack"` to `#tool "nuget:?package=ILRepack"` to make it more clear.

Also wonder if it would be enough if that was added too sample code though? I.e. here.
https://github.com/gep13/cake/blob/Issue-601/src/Cake.Common/Tools/ILMerge/ILMergeAliases.cs#L30

gep13:
@devlead yip, that makes sense, and yes, I agree that is should **also** be added to sample code.

I think we should also have a `<see>` link as well, which takes you to the main site of the tool.  Thoughts?

devlead:
@gep13 for me a see href makes sense.
@patriksvensson Thoughts?

patriksvensson:
@gep13 A <see> tag to the tool would be very useful indeed.

patriksvensson:
@gep13 @devlead I *think* I added support for `<see href=...">` but I'm not sure. Will check.

patriksvensson:
@gep13 @devlead Yes, it's supported! :smile:

devlead:
@gep13 this PR is out-of-date with the develop branch, could you please review, rebase and ping us when you think it's ready for review again.

gep13:
@devlead I have just updated this.  Let me know what you think, and if you are happy with it, I can go ahead (at some point) and do the same thing for other aliases as well.

devlead:
LGTM 👍

Issue #619, opened 2016-01-03
Add GitLink Alias


gep13:
In the same way that we have GitReleaseManager, GitReleaseNotes and GitVersion aliases, we should really complete the set and add an alias for GitLink:

https://github.com/gittools/gitlink

All the command line arguments can be found here:

https://github.com/GitTools/GitLink/blob/develop/src/GitLink/HelpWriter.cs

ghuntley:
Can't wait - thanks for working on this.

gep13:
>@ghuntley said...
Can't wait - thanks for working on this.

Not a problem at all, adding this completes the set of GitTools i.e. GitVersion, GitReleaseNotes, GitReleaseManager and GitLink, so happy to see it getting added :+1:

PR #634, opened 2016-01-10
(GH-619) Added GitLink Support with Tests

gep13:
Relates to #619

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
@@ -1,0 +1,240 @@
+﻿using Cake.Common.Tests.Fixtures.Tools;
+using Cake.Core;
+using Cake.Core.IO;
+using Cake.Testing;
+using Cake.Testing.Xunit;
+using Xunit;
+
+namespace Cake.Common.Tests.Unit.Tools.GitLink
+{
+    public sealed class GitlinkRunnerTests
+    {
+        public sealed class TheRunMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Repository_Root_Path_Is_Null()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.RepositoryRootPath = null;
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentNullException(result, "repositoryRootPath");
+            }
+
+            [Fact]
+            public void Should_Find_GitLink_Runner()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("/Working/tools/gitlink.exe", result.Path.FullPath);
+            }
+
+            [Fact]
+            public void Should_Throw_If_Process_Was_Not_Started()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.GivenProcessCannotStart();
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsType<CakeException>(result);
+                Assert.Equal("GitLink: Process was not started.", result.Message);
+            }
+
+            [Fact]
+            public void Should_Throw_If_Has_A_Non_Zero_Exit_Code()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.GivenProcessExitsWithCode(1);
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsType<CakeException>(result);
+                Assert.Equal("GitLink: Process returned an error.", result.Message);
+            }
+
+            [Fact]
+            public void Should_Use_Provided_Repository_Root_Path_In_Process_Arguments()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.RepositoryRootPath = "source";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("\"/Working/source\"", result.Args);
+            }
+
+            [WindowsFact]
+            public void Should_Set_RepositoryUrl()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.RepositoryUrl = "http://mydomain.com";
+
+                // Then
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("\"c:/temp\" -u \"http://mydomain.com\"", result.Args);
+            }
+
+            [WindowsFact]
+            public void Should_Set_SolutionFileName()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.SolutionFileName = "solution.sln";
+
+                // Then
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("\"c:/temp\" -f \"solution.sln\"", result.Args);
+            }
+
+            [WindowsFact]
+            public void Should_Set_Configuration()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.Configuration = "Release";
+
+                // Then
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("\"c:/temp\" -c \"Release\"", result.Args);
+            }
+
+            [WindowsFact]
+            public void Should_Set_Platform()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.Platform = "AnyCPU";
+
+                // Then
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("\"c:/temp\" -p \"AnyCPU\"", result.Args);
+            }
+
+            [WindowsFact]
+            public void Should_Set_Branch()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.Branch = "master";
+
+                // Then
+                var result = fixture.Run();
+
+                // Then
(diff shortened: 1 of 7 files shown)


gep13:
@GeertvanHorrik when you have a minute, can you have a review of this PR? Thanks!

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
```diff
@@ -0,0 +1,241 @@
...
+                Assert.IsType<CakeException>(result);
+                Assert.Equal("GitLink: Process returned an error.", result.Message);
+            }
+
+            [Fact]
+            public void Should_Use_Provided_SolutionPath_In_Process_Arguments()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.SolutionPath = "source";
```
GeertvanHorrik: I might have chosen a wrong name in GitLink for this. It's actually the repository root on disk, has nothing to do with the solution.

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
```diff
@@ -0,0 +1,241 @@
...
+                // Then
+                Assert.Equal("\"c:/temp\" -d \"/Working/pdb\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Set_PowerShell_Switch()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.UsePowerShellCommand = true;
```
GeertvanHorrik: Maybe best to use UsePowerShell to stay equal to the terminology in GitLink, but I'll leave that up to you.

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
```diff
@@ -0,0 +1,241 @@
...
+                // Then
+                Assert.Equal("\"c:/temp\" -skipverify", result.Args);
+            }
+
+            [Fact]
+            public void Should_Set_Debug_Switch()
+            {
+                // Given
+                var fixture = new GitLinkFixture();
+                fixture.Settings.Debug = true;
```
GeertvanHorrik: Maybe rename to IsDebug?

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
```diff
@@ -0,0 +1,150 @@
...
+            Run(settings, GetArguments(solutionPath, settings));
+        }
+
+        private ProcessArgumentBuilder GetArguments(DirectoryPath solutionPath, GitLinkSettings settings)
+        {
+            var builder = new ProcessArgumentBuilder();
+
+            builder.AppendQuoted(solutionPath.MakeAbsolute(_environment).FullPath);
+
+            if (settings.RepositoryUrl != null)
```
GeertvanHorrik: Why not using string.IsNullOrWhitespace instead?

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
```diff
@@ -0,0 +1,150 @@
...
+                builder.AppendQuoted(settings.Platform);
+            }
+
+            if (!string.IsNullOrWhiteSpace(settings.Branch))
+            {
+                builder.Append("-b");
+                builder.AppendQuoted(settings.Branch);
+            }
+
+            if (settings.LogFilePath != null)
```
GeertvanHorrik: Use !string.IsNUllOrWhitespace instead?

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+            Run(settings, GetArguments(solutionPath, settings));
+        }
+
+        private ProcessArgumentBuilder GetArguments(DirectoryPath solutionPath, GitLinkSettings settings)
+        {
+            var builder = new ProcessArgumentBuilder();
+
+            builder.AppendQuoted(solutionPath.MakeAbsolute(_environment).FullPath);
```
GeertvanHorrik: I am assuming this is the "root" of the git repository in cake?

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
gep13: @GeertvanHorrik but it is still a path to the root folder, right?  I can fix up the name to make it clearer.  Perhaps `RepositoryRootPath`?

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
gep13: @GeertvanHorrik yip, I can do that.

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
gep13: @GeertvanHorrik yip, makes sense to me, I will correct.

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
GeertvanHorrik: That's a better name. It will check all subdirectories for .sln

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
gep13: @GeertvanHorrik this is a Uri object, so are you saying do:

`string.IsNullOrWhiteSpace(settings.RepositoryUrl.ToString())`

if so, I would still need to check for null object on settings.RepositoryUrl, right?

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
gep13: @GeertvanHorrik this is an object of Type FilePath, not just a string.  So either it has something, or it doesn't.

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
GeertvanHorrik: I wasn't aware it was an uri. I don't see the benefit of using uri above string here, but you know better about the internals of Cake.

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
gep13: @GeertvanHorrik not sure exactly what you mean.  The expectation is that the user will provide a path, either relative or not to the build.cake file in the repository, and this will be made to the absolute path.

src/Cake.Common.Tests/Unit/Tools/GitLink/GitlinkRunnerTests.cs
gep13: :+1:

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
gep13: Not really anything to do with Cake, just a personal preference.  If you have a variable called RepositoryUrl, then using a Uri object, in my mind, just makes sense.  :smile:

GeertvanHorrik: Agree when you need to code against it, but this is just a pass-through object anyway. You even have chance it will "mess" (change) your stuff this way.

gep13: @GeertvanHorrik that is a fair point. I will re-think this when I am fixing up your other suggestions.

src/Cake.Common/Tools/GitLink/GitLinkRunner.cs
gep13: @GeertvanHorrik did you see my question above?

GeertvanHorrik: Because I made the wrong naming decision in the beginning of GitLink, I want to make sure this is the root of the git repository, not the root of some subdirectory where the solution is in.

gep13: ah, right, yes, this one will get re-named as per the above.

gep13:
@GeertvanHorrik I have just updated this PR based on your feedback.  Can you have another look?

@devlead @patriksvensson I can squash this commit, once everyone is happy.

GeertvanHorrik:
Looks good!

patriksvensson:
@gep13 Tests are failing on Mono. If this is a Windows tool only, there's a `WindowsFact` and `WindowsTheory` you can use to make sure that the OS specific tests (paths) do not run on Linux.

gep13:
@GeertvanHorrik Am I right in saying that GitLink only runs on Windows?  If so, I will get the tests updated as per @patriksvensson suggestion.  Thanks!

patriksvensson:
@gep13 If GitLink runs on *nux as well, change the path to be platform agnostic, i.e. something like /Working/Stuff, which is a valid Windows path as well (default drive or working directory is normally appended in the file system).

GeertvanHorrik:
Yes, GitLink only runs on Windows (because internally it uses pdbstr.exe, which is a windows tool). But it might change in the future because the pdb format is going open source (from what I understood). For now, Windows only unfortunately.

gep13:
@GeertvanHorrik ok, that is great.  I will update this PR on that basis, and as/when GitLink moves to working xplat, we can update the support in Cake as well.

gep13:
@patriksvensson @devlead this one is ready for final review/merging.

PR #641, opened 2016-01-16
[WIP] Adding XmlPeek for reading XML values.

patridge:
## Goal

Allow a build script to extract values from XML files.

Example

```
Task("PrintServerValue")
    .Does(() =>
{
    var file = File("test.xml");
    var serverValue = XmlPeek("/configuration/appSettings/add[@key = 'server']/@value");
    Information(serverValue);
});
```

## Current State

I copied over XmlPoke code and its testing code as well. From there, I removed the writing logic, and set it to return a `Value` string.

## Work Needed

I am terrible with XPath stuff, and I haven't been able to confirm what I've written will actually work. Unfortunately, I am working on a Mac using Xamarin Studio for these changes. I ran into a number of issues building and testing. I feel bad even putting this in a pull request, but I've been changing so many things trying to test it locally with NUnit, I wouldn't be sure I was even testing the same thing if I get it to work. Here are the issues I ran into that may or may not be easy to fix.

* The current develop branch doesn't build at all, even before I made changes. Every instance of `using Cake.Common.Tests.Properties;` is a build error. I have these removed locally, but not committed.
* Running xUnit tests on Xamarin Studio is problematic. I haven't figured out how to get it to happen at all.
* Running `./build.sh` from the command line built it fine, but failed on every non-skipped unit tests with the following error.

> System.Resources.MissingManifestResourceException : Could not find any resources appropriate for the specified culture or the neutral culture.  Make sure "Cake.Common.Tests.Resources.resources" was correctly embedded or linked into assembly "Cake.Common.Tests" at compile time, or that all the satellite assemblies required are loadable and fully signed.

I don't know if these problems are all specific to my machine or to my environment in general (OS X, XS, etc.). The first item above could even be a preexisting issue with current develop branch.

### DIFF
--- a/src/Cake.Common/Xml/XmlPeekAliases.cs
+++ b/src/Cake.Common/Xml/XmlPeekAliases.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.Globalization;
+using System.IO;
+using System.Xml;
+using Cake.Core;
+using Cake.Core.Annotations;
+using Cake.Core.Diagnostics;
+using Cake.Core.IO;
+
+namespace Cake.Common.Xml
+{
+    /// <summary>
+    /// Contains functionality related to XML XPath queries.
+    /// </summary>
+    [CakeAliasCategory("XML")]
+    public static class XmlPeekAliases
+    {
+        /// <summary>
+        /// Gets the value of a target node.
+        /// </summary>
+        /// <returns>The value found at the given XPath query.</returns>
+        /// <param name="context">The context.</param>
+        /// <param name="filePath">The target file.</param>
+        /// <param name="xpath">The xpath of the node to get.</param>
+        [CakeMethodAlias]
+        public static string XmlPeek(this ICakeContext context, FilePath filePath, string xpath)
+        {
+            return context.XmlPeek(filePath, xpath, new XmlPeekSettings());
+        }
+        
+        /// <summary>
+        /// Get the value of a target node.
+        /// </summary>
+        /// <returns>The value found at the given XPath query.</returns>
+        /// <param name="context">The context.</param>
+        /// <param name="filePath">The target file.</param>
+        /// <param name="xpath">The xpath of the nodes to set.</param>
+        /// <param name="settings">Additional settings to tweak Xml Peek behavior.</param>
+        [CakeMethodAlias]
+        public static string XmlPeek(this ICakeContext context, FilePath filePath, string xpath, XmlPeekSettings settings)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+            
+            if (filePath == null)
+            {
+                throw new ArgumentNullException("filePath");
+            }
+            
+            var file = context.FileSystem.GetFile(filePath);
+            if (!file.Exists)
+            {
+                throw new FileNotFoundException("Source File not found.", file.Path.FullPath);
+            }
+            
+            using (var memoryStream = new MemoryStream())
+            {
+                using (var fileStream = file.Open(FileMode.Open, FileAccess.Read, FileShare.None))
+                using (var xmlReader = XmlReader.Create(fileStream))
+                using (var xmlWriter = XmlWriter.Create(memoryStream))
+                {
+                    var xmlValue = XmlPeek(xmlReader, xpath, settings);
+                    if (xmlValue == null)
+                    {
+                        context.Log.Warning("Warning: Failed to find node matching the XPath '{0}'", xpath);
+                    }
+                    return xmlValue;
+                }
+            }
+        }
+        
+        /// <summary>
+        /// Gets the value of a target node.
+        /// </summary>
+        /// <returns>The value found at the given XPath query (or the first, if multiple eligible nodes are found).</returns>
+        /// <param name="source">The source xml to transform.</param>
+        /// <param name="xpath">The xpath of the nodes to set.</param>
+        /// <param name="settings">Additional settings to tweak Xml Peek behavior.</param>
+        private static string XmlPeek(XmlReader source, string xpath, XmlPeekSettings settings)
+        {
+            if (source == null)
+            {
+                throw new ArgumentNullException("source");
+            }
+            
+            if (string.IsNullOrWhiteSpace(xpath))
+            {
+                throw new ArgumentNullException("xpath");
+            }
+            
+            if (settings == null)
+            {
+                throw new ArgumentNullException("settings");
+            }
+            
+            var document = new XmlDocument();
+            document.PreserveWhitespace = settings.PreserveWhitespace;
+            document.Load(source);
+            
+            var namespaceManager = new XmlNamespaceManager(document.NameTable);
+            foreach (var xmlNamespace in settings.Namespaces)
+            {
+                namespaceManager.AddNamespace(xmlNamespace.Key /* Prefix */, xmlNamespace.Value /* URI */);
+            }
+            
+            var node = document.SelectSingleNode(xpath, namespaceManager);
+            
+            return node != null ? node.Value : null;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


patridge:
It seems I've also got some StyleCop work to do. I'll do my best to fix things blindly from XS.

patriksvensson:
@patridge Thanks for your contribution :+1: 

First of all, let me say that I'm sorry for the bad Xamarin Studio experience you got when building this. I don't use Xamarin Studio to work on this project so it's kind of not supported at the moment, but I would absolutely want this to work. StyleCop is also an issue that we need to solve on *nix, but we haven't got around to do any work on it yet. We build and run tests via our CI server though (Linux and OS X), so it should build and run on Mono without problem.

Tell me when you feel happy with the PR and I'll review it.

Thanks!

patriksvensson:
@patridge I've created an issue for this: #642

patridge:
@patriksvensson No need to apologize. Sorry in advance for how noisy this PR might get as I push through the CI errors. I'll mention you again (after this) when I feel like I've done as much as I can.

While it is only a partial fix, Xamarin Studio can store style preferences in the solution file. While it isn't enforced anywhere, it could prevent a few potential issues (especially tabs/spaces stuff).

patridge:
@patriksvensson I guess I might be bugging you before I'm done. The [current issue](https://ci.appveyor.com/project/cakebuild/cake/build/0.7.0-build-1396) I'm seeing is what I was seeing before I even started adding code (first bullet in the original PR comment above regarding Properties not existing for `using Cake.Common.Tests.Properties;`).

I have the changes committed and will push them up, but I'm not sure how develop was building properly before I touched things.

patriksvensson:
@patridge Ah, it seems like Xamarin Studio is messing with `src/Cake.Common.Tests/Properties/Resources.Designer.cs` and rewriting paths to things.

Maybe you should add the XML directly to the fixture and revert the resources until we know how we can fix this.

patridge:
@patriksvensson Will do for now. I definitely don't have a lot of experience with embedded resources, so I won't be much help resolving the XS issue with things.

patriksvensson:
@patridge Perfect! Ping me if I can help out with anything.

patridge:
I figured out what Mono did to Resource.designer.cs and manually corrected it with a text editor. Now the new unit tests are actually failing in my `./build.sh` calls locally. Clearly, my blind coding isn't quite perfect, but at least it is something I can work with. :smiley:

patridge:
I had a major XML oops in there with multiple root elements so I could test a text node. Then, I had to relearn some XPath for how to get a node's text. But that is all worked out now.

@patriksvensson All seems to be well now, and tests are passing. I'm sure there are other project-specific things to address now (e.g., docs and such), so let me know if there are other things needed to make this acceptable.

src/Cake.Common/Xml/XmlPeekAliases.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+        
+        /// <summary>
+        /// Gets the value of a target node.
+        /// </summary>
+        /// <returns>The value found at the given XPath query (or the first, if multiple eligible nodes are found).</returns>
+        /// <param name="source">The source xml to transform.</param>
+        /// <param name="xpath">The xpath of the nodes to set.</param>
+        /// <param name="settings">Additional settings to tweak Xml Poke behavior.</param>
+        private static string XmlPeek(XmlReader source, string xpath, XmlPokeSettings settings)
```
patriksvensson: Would prefer if we added a `XmlPeekSettings` here.

src/Cake.Common/Xml/XmlPeekAliases.cs
```diff
@@ -0,0 +1,113 @@
...
+            var namespaceManager = new XmlNamespaceManager(document.NameTable);
+            foreach (var xmlNamespace in settings.Namespaces)
+            {
+                namespaceManager.AddNamespace(xmlNamespace.Key /* Prefix */, xmlNamespace.Value /* URI */);
+            }
+            
+            var node = document.SelectSingleNode(xpath, namespaceManager);
+            if (node == null)
+            {
+                const string format = "Failed to find node matching the XPath '{0}'";
```
patriksvensson: Since we're only querying for a value, wouldn't it be better to log a warning (Failed to find) and return null? This case the consumer don't have to wrap this in try/catch clause.

src/Cake.Common.Tests/Unit/XML/XmlPeekTests.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System.IO;
+using Cake.Common.Tests.Fixtures;
+using Xunit;
+
+namespace Cake.Common.Tests.Unit.XML
+{
+    public sealed class XmlPeekTests
```
patriksvensson: This should be called `XmlPeekAliasesTests`.

Would prefer it you use the same naming conventions as other tests:
  `XmlPeekAliasesTests` -> `TheXmlPeekMethod` -> `Should_Throw_If_FilePath_Is_Null`

src/Cake.Common.Tests/Fixtures/XmlPeekFixture.cs
```diff
@@ -0,0 +1,46 @@
...
+using Cake.Common.Tests.Properties;
+using Cake.Common.Xml;
+using Cake.Core;
+using Cake.Core.IO;
+using Cake.Testing;
+using NSubstitute;
+
+namespace Cake.Common.Tests.Fixtures
+{
+    internal sealed class XmlPeekFixture
```
patriksvensson: Should be called `XmlPeekAliasesFixture`.

patriksvensson:
@patridge Great work! I left some comments for you.

src/Cake.Common/Xml/XmlPeekAliases.cs
patridge: Addressed in a [new commit](https://github.com/cake-build/cake/commit/56a211eda0da0ffede7ba4edaaf4a309edefd287).

src/Cake.Common/Xml/XmlPeekAliases.cs
patridge: Good call. I definitely prefer that approach.

I see that I'll have access to an `ICakeLog` via `context.Log` in the extension methods, but I don't suppose you can point me to a good example to reference/duplicate for implementing this system.

src/Cake.Common.Tests/Unit/XML/XmlPeekTests.cs
patridge: Can do. Since I'd only been basing things on XmlPokeTests, I didn't realize it, and my derived code, diverged from the preferred conventions.

I made changes trying to match what you describe in [this commit](https://github.com/patridge/cake/commit/3c174c13e35ec06a0b0010c6e038f47fd6dd8b80).

src/Cake.Common.Tests/Fixtures/XmlPeekFixture.cs
patridge: Addressed in [this commit](https://github.com/patridge/cake/commit/3c174c13e35ec06a0b0010c6e038f47fd6dd8b80).

src/Cake.Common/Xml/XmlPeekAliases.cs
patridge: I found the various logging extension methods and put something in place in [this commit](https://github.com/cake-build/cake/commit/93cc424e865df77ba0db8bc4cbcbe31a2042d570).

patridge:
@patriksvensson I did my best to address the comments.

patriksvensson:
@patridge Found a xml doc comment that wasn't updated, but other than that it looks good to me!

After you've fixed the comment, could you rebase against develop and squash to one commit so I can merge this. Thanks! :+1:

gep13:
@patridge this is great work! Thanks for your contribution! :+1:

patridge:
I've only done a couple interactive rebases before, and only one successfully ending where I wanted things. I think things worked out again. If not, I saved things in a local branch if I need to try again.

@patriksvensson I fixed that doc comment before I did the rebase. Let me know if anything isn't ready to go.

gep13:
@patridge interactive rebases can be problematic, and a little scary, when you start to do them, but after a while, they become second nature.  I now routinely do interactive rebases as part of my development workflow.  It is a very powerful tool.  Seems like things have worked well here in this PR :+1:

patriksvensson:
@patridge Great! Will take a look at it as soon as I can. :+1:

Issue #278, opened 2015-07-07
Improve Tool Discovery


Redth:
By default right now most aliases that require external tools seem to look for the tool in `./tools/`, so the tools folder always relative to the executing script.  

While I understand the original intention the way to use cake is to install it in a given directory where you need it, I think it's desirable to have Cake (and its tools) optionally located in a central location on a system. For one, this would mean less duplication of executables (both cake and tools) and for complicated setups where there are cake scripts at various levels in a given directory tree, we wouldn't need to install cake in each level.

I think it's possible to allow discovery of tools beyond the `./tools/` folder without disrupting the behaviour when that folder actually exists.

I think the discovery process could go something like this:

1. Add a `-toolpath=` parameter to Cake for the user to specify the tool path
2. Look for tool in `./tools/` (exactly how it currently works)
3. Look in all the directories within the system `PATH` variable for the tool
4. On certain operating systems, some tools might be commonly placed in one location, look to see if it exists here (for example on Mac, nuget is available at `/usr/bin/nuget`)

Redth:
For the last part, where a tool may have a specific location on a given OS, there should be some sort of `Func` exposed by the tool that can do the tool specific work of looking in a tool specific path for the tool.  This would only be invoked if the tool wasn't found in the other locations first.

patriksvensson:
Note: PATH variable lookup implemented.

SharpeRAD:
I second this :+1:

You can achieve shared a shared tools location by passing in the directory to the build script but it means you can't use the "#tools" directive :disappointed: 

```c#
var tools = Argument("tools", "./tools");

Setup(() =>
{
	//Executed BEFORE the first task.
	Information("Building version {0} of {1}.", semVersion, appName);
    
	NuGetInstall("NUnit.Runners", new NuGetInstallSettings
	{
		ExcludeVersion  = true,
		OutputDirectory = tools
    });
});

Task("Run-Unit-Tests")
    .IsDependentOn("Build")
    .Does(() =>
{
    CreateDirectory(testResultsDir);

    NUnit3("./src/**/*.Tests.dll", new NUnit3Settings
	{
        ToolPath = tools + "NUnit.Console/tools/nunit3-console.exe",
        Framework = "4.0",

        OutputFile = testResultsDir + "tests.xml",
        ErrorOutputFile = testResultsDir + "error.xml",
        Results = testResultsDir + "results.xml"
    });
});
```

What I was proposing in #647 solves the problem with environment variables :smile:

[...]

PR #647, opened 2016-01-18
Tools and Addins environment variables

SharpeRAD:
Allows you to define shared folders in your powershell scripts for Cake to use when resolving / installing tools and addins:

```powershell
# Find tools
if(($Tools.IsPresent) -and (Test-Path $Tools))
{
    # Parameter
    Write-Host "Using tools parameter"
    $TOOLS_DIR = $Tools
}
elseif (Test-Path "C:/Tools/Cake/Cake.exe") 
{
    # Shared location
    Write-Host "Using shared tools"
    $TOOLS_DIR = "C:/Tools/"
}
else
{
    # Local path
    Write-Host "Using local tools"
    $PSScriptRoot = split-path -parent $MyInvocation.MyCommand.Definition
    $TOOLS_DIR = Join-Path $PSScriptRoot "tools"

    if (!(Test-Path $TOOLS_DIR)) 
    {
        Write-Host "Creating tools directory"
        New-Item $TOOLS_DIR -itemtype directory
    }
}

# Save paths to environment for use in child processes
$ENV:CAKE_TOOLS = $TOOLS_DIR
$ENV:CAKE_ADDINS = $TOOLS_DIR + "Addins/"
```

### DIFF
--- a/src/Cake.Core/Scripting/ScriptRunner.cs
+++ b/src/Cake.Core/Scripting/ScriptRunner.cs
@@ -110,13 +110,35 @@ namespace Cake.Core.Scripting
 
             // Install tools.
             _log.Verbose("Processing build script...");
-            var toolsPath = scriptPath.GetDirectory().Combine("tools");
+            var toolsEnv = _environment.GetEnvironmentVariable("CAKE_TOOLS");
+
+            DirectoryPath toolsPath;
+            if (!string.IsNullOrEmpty(toolsEnv))
+            {
+                toolsPath = new DirectoryPath(toolsEnv).MakeAbsolute(_environment);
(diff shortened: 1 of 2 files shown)


devlead:
Hi @SharpeRAD,
Thanks for contributing :+1: 
See what your trying to achieve here and it makes sense, though it can't be merged in the current state, think this will need to be discussed more in an issue so we can get more implementation details sorted first before we throw code at it :wink: 
This is a Core change, as such will need some more tests and make sure we cover all places where add-ins & tools are used, to avoid introducing regression.
I' will close this and we can continue discussion in #278 and reopen when we reached consensus.

Issue #638, opened 2016-01-14
Task Summary should include skipped tasks


gep13:
As part of a build, you might see something like this:

![image](https://cloud.githubusercontent.com/assets/1271146/12338366/26ab430a-bb07-11e5-9077-21b109b88e88.png)

But there is no mention of it in the task summary:

![image](https://cloud.githubusercontent.com/assets/1271146/12338386/38fd3f5e-bb07-11e5-9579-2f4fff5989cc.png)

This should be included in the summary, but in gray and with no time associated with it, to indicate that it was skipped.

Should this be an optional flag to include this information, or always enabled?

patriksvensson:
@gep13 I think it should always be enabled. Really useful to see what tasks that were skipped and good for debugging.

gep13:
:+1:

patriksvensson:
Closed via #648

gep13:
Woot! Thanks for the contribution @ErikSchierboom

ErikSchierboom:
@gep13 My pleasure!

PR #648, opened 2016-01-19
Display skipped tasks in task summary. #638

ErikSchierboom:
This PR ensures that skipped tasks are displayed in the task summary (issue #638). It does this by adding a `Skipped` property to the `CakeReport` class. 

When you run the updated CAKE version on a script that has two skipped tasks, the new output looks as follows:

image

I'm not sure the "(skipped)" text is the best option, but I preferred it over "-" and "- (skipped)".

Everything has been unit tested (save for the changing of the foreground color, which I don't know how to test). In the process, I have also added tests that verify that the `CakeReport` returned by the `RunTarget` method is correct.

### DIFF
--- a/src/Cake.Core.Tests/Cake.Core.Tests.csproj
+++ b/src/Cake.Core.Tests/Cake.Core.Tests.csproj
@@ -73,6 +73,7 @@
     <Compile Include="Unit\Annotations\CakePropertyAliasAttributeTests.cs" />
     <Compile Include="Unit\CakeContextTests.cs" />
     <Compile Include="Unit\CakeEngineTests.cs" />
+    <Compile Include="Unit\CakeReportTests.cs" />
     <Compile Include="Unit\CakeTaskBuilderExtensionsTests.cs" />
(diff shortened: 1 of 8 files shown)


patriksvensson:
@ErikSchierboom Awesome work :+1: I don't think testing the foreground color is super important to be honest. The important thing is that the logic around it all is tested and that you've done.

About the `(Skipped)` part, I was thinking that it might be better to just output `Skipped`.
What do you think?

(ping @devlead @gep13)

gep13:
@patriksvensson @ErikSchierboom if I had to choose, I would say drop the `( )`'s as well.  Also, I think it should be `Skipped` rather than `skipped`.  Small point, but we should get this right before pulling in.

Thanks for getting this PR done, it is something I am looking forward to running :smile:

ErikSchierboom:
@patriksvensson @gep13 I have changed the output to `"Skipped"`.

patriksvensson:
@ErikSchierboom Merged! Thank you very much for this contribution! :smile:

ErikSchierboom:
@patriksvensson My pleasure! You have an excellent code base.

Issue #621, opened 2016-01-07
Add method to get relative paths (for paths)


patriksvensson:
It would be nice (especially for logs) if `DirectoryPath` and `FilePath` had a `GetRelativePath` method that returned a new path that is relative to another path.

```csharp
public class FilePath {
  public FilePath GetRelativePath(FilePath to) {
    // Use file's directory as from path.
  }
  public DirectoryPath GetRelativePath(DirectoryPath to) {
    // Use file's directory as from path.
  }
}
```

```csharp
public class DirectoryPath {
  public FilePath GetRelativePath(FilePath to) {
    ...
  }
  public DirectoryPath GetRelativePath(DirectoryPath to) {
    ...
  }
}
```

**Requirements**

- Full unit test coverage proving that the functionality works as expected.
- No PInvoke calls to Shlwapi.dll to get the relative path.
- No URI tricks to get the relative path.
- Both from/to paths must be absolute.
- If no relative path can be calculated, the method should throw.

**Basic test cases**

```csharp
var path = new DirectoryPath("C:/A/D/E");
var result = path.GetRelativePath(new DirectoryPath("C:/A/B/C"));
Assert.Equal("../../B/C", result.FullPath);
```

```csharp
var path = new DirectoryPath("D:/A/B/C");
var result = Record.Exception(() => path.GetRelativePath(new DirectoryPath("C:/D/E/F")));
Assert.IsInstanceOf<InvalidOperationException>(result);
Assert.Equal("Paths must share a common prefix.", result.Message);
```

```csharp
var path = new DirectoryPath("C:/A/B/C/D/E/F");
var result = path.GetRelativePath(new DirectoryPath("C:/A/B/C"));
Assert.Equal("../../..", result.FullPath);
```

```csharp
var path = new DirectoryPath("C:/A/B/C");
var result = path.GetRelativePath(new DirectoryPath("C:/A/B/C/D/E/F"));
Assert.Equal("./D/E/F", result.FullPath);
```

ErikSchierboom:
@patriksvensson I am looking into this feature. What do you expect the returned FilePath to be for this code:

```csharp
var path = new FilePath("C:/hello.txt");
var result = path.GetRelativePath(new FilePath("C:/hello.txt"));
result = ???
```

Should it be `"."`, or `"./hello.txt"`, or just `"hello.txt"`?

ErikSchierboom:
[...]

PR #650, opened 2016-01-20
Add methods to get relative paths (#621)

ErikSchierboom:
This PR implements #621 and adds methods to get relative paths for both the `FilePath` and `DirectoryPath` classes.

This is a WIP as I was wondering how you'd want me to do the refactoring. At the moment, there is a lot of duplication, but I was not sure what would be the best way to refactor things. I could move the methods to the abstract base class and add some extract abstract methods, but I'm not sure you'd want to go that way. Any ideas?

### DIFF
--- a/src/Cake.Core/IO/DirectoryPath.cs
+++ b/src/Cake.Core/IO/DirectoryPath.cs
@@ -151,5 +151,30 @@ namespace Cake.Core.IO
         {
             return new DirectoryPath(path);
         }
+
+        /// <summary>
+        /// Get the relative path to another directory.
+        /// </summary>
+        /// <param name="to">The target directory path.</param>
+        /// <returns>A <see cref="DirectoryPath"/>.</returns>
+        public DirectoryPath GetRelativePath(DirectoryPath to)
+        {
+            return RelativePathResolver.Resolve(this, to);
+        }
+
+        /// <summary>
+        /// Get the relative path to another file.
+        /// </summary>
+        /// <param name="to">The target file path.</param>
+        /// <returns>A <see cref="FilePath"/>.</returns>
+        public FilePath GetRelativePath(FilePath to)
+        {
+            if (to == null)
+            {
+                throw new ArgumentNullException("to");
+            }
+
+            return GetRelativePath(to.GetDirectory()).GetFilePath(to.GetFilename());
+        }
     }
 }
--- a/src/Cake.Core.Tests/Unit/IO/DirectoryPathTests.cs
+++ b/src/Cake.Core.Tests/Unit/IO/DirectoryPathTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using Cake.Core.IO;
+using Cake.Testing.Xunit;
 using NSubstitute;
 using Xunit;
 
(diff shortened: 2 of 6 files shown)


patriksvensson:
Great job! :+1: 

I think that all permutations of this only uses things found in `Cake.Core.IO.Path`, so I guess you could move the code to something like `string GetRelativePath(Path from, Path to)` and return the path as a string. Then you can do the conversion to the right type in each method.

Not sure if this works since I see some minor differences in the code, but maybe that could be solved somehow.

The reason the tests fail is that Mono recognizes `C:/` as a relative path. You can fix this by either converting paths in the tests to Unix paths (without a drive) or mark the tests with a `WindowsTheory`/`WindowsFact` attribute. I think it would be useful to test both Windows and Unix paths here though.

ErikSchierboom:
The PR has been updated. I have now separated the tests into Windows path format tests and Unix format tests. That should fix the Travis build errors.

The code has also been refactored, to remove duplication.

src/Cake.Core/IO/DirectoryPath.cs
```diff
@@ -151,5 +151,70 @@ public static DirectoryPath FromString(string path)
         {
             return new DirectoryPath(path);
         }
+
+        /// <summary>
+        /// Get the relative path to another directory.
+        /// </summary>
+        /// <param name="to">The target directory path.</param>
+        /// <returns>A <see cref="DirectoryPath"/>.</returns>
+        public DirectoryPath GetRelativePath(DirectoryPath to)
```
patriksvensson: Could we break the logic of this method to it's own file. Mostly an aesthetic adjustment, but would follow how we've done with the PathCollapser stuff.

patriksvensson:
@ErikSchierboom Looks really good! :+1: Left a comment about a minor thing for you.

ErikSchierboom:
@patriksvensson Nice suggestion. I will get on it.

When looking at the `PathCollapserTests` file, I saw some `#if !UNIX` statements. Shouldn't those be converted to `[WindowsFact]` calls?

ErikSchierboom:
@patriksvensson The PR has been updated. I'm not sure about the name though.

patriksvensson:
@ErikSchierboom Maybe call it `RelativePathResolver` with a `Resolve` method. The current name is not bad per se, but might be confused with something that inherits from `Path`. What do you think?

Great finding with the `#if !UNIX` pre-processor. Should absolutely be a `WindowsTheory`, but I think that is better to fix that outside of this PR.

ErikSchierboom:
@patriksvensson Great name! I have updated the PR.

I will convert the `#if !UNIX` calls in a separate PR.

patriksvensson:
@ErikSchierboom Looks good to me. Will merge later tonight once I'm back home. :+1:

patriksvensson:
@ErikSchierboom Is this PR done? I saw the `[WIP]` tag in the title and just want to be sure I don't merge something that you're still working on.

ErikSchierboom:
@patriksvensson Whoops, I totally forgot to remove that. It is ready.

patriksvensson:
@ErikSchierboom Merged! Thanks! :+1:

Issue #649, opened 2016-01-19
FileExists alias in should make the provided path absolute


patriksvensson:
The `FileExists` alias in `Cake.Common` should make the provided path absolute before passing it to the file system abstraction. Not really a bug since it currently has no side effects, but if we ever plan on using Cake in another kind of host, this could pose a problem.

patriksvensson:
Merged via #651.

PR #651, opened 2016-01-20
FileExists alias uses absolute paths (#649)

ErikSchierboom:
This PR modifies the `FileExists` method to use an absolute path to determine if a file exists. It implements #649


patriksvensson:
Merged! Thank you for this! :+1:

PR #656, opened 2016-01-24
Allow double-dash arguments (#474)

ErikSchierboom:
This PR adds support for specifying the long-named version of parameters using the `"--"` prefix. It doesn't deprecate the existing `"-"` prefix, which also still works for long-named parameters.


patriksvensson:
Looks good to me! Will take a closer look at it tonight when I get home. :+1:

patriksvensson:
Merged! Awesome work :+1:

PR #657, opened 2016-01-25
Add DotNetBuild settings extension methods and tests

ErikSchierboom:
The `MSBuildSettings` and `XBuildSettings` classes have some convenient extensions methods. These were not present for the `DotNetBuildSettings` class. This PR adds those extension methods. It also adds tests for the `DotNetBuildSettings` class itself, which was untested.


patriksvensson:
Merged! Rebased this against develop for you, so GitHub does not recognize it as merged though so I have to close it.

Thanks :+1:

Issue #639, opened 2016-01-14
Don't show delegating tasks in summary


gep13:
i.e. Tasks that are only used to tie together other tasks.  If there is no action associated with the Task, don't show it in the summary.

patriksvensson:
:+1:

RichiCoder1:
Or show it in a non-tasky way to may it clear it's a summary or branch task.

ErikSchierboom:
How would you like the delegating task to be displayed? In grey, just like skipped tasks? And what should be the displayed text?

ErikSchierboom:
In a previous PR, I add a `Skipped` property to the `CakeReportEntry` class. However, I now think that should really have been an enumeration. That way, we can have an enumeration that specifies the executed task's status: `Executed`, `Skipped` or `Delegated`. Should I replace the boolean property with an enumeration, or create an additional boolean property?

patriksvensson:
@ErikSchierboom I agree with your thinking. An enum would be much cleaner. Not sure yet how the delegating task should be displayed. Not against graying it out, but the total won't match the sum of the tasks (since tasks that "delegate" could take some milliseconds). Not sure if this really matters. Maybe we can write the time for the delegated task but gray it out? 

What do you think?

(ping @gep13 @devlead)

devlead:
Gray is nice, but as build server logs are just text might not clear enough. Maybe just report on diagnostic / verbose?

patriksvensson:
@devlead Reports are not written to the log.

devlead:
Usually build servers display console output as log right?

patriksvensson:
@devlead Yes, they are. But we're writing the report using `IConsole` (and not `ICakeLog`) beacuse we want to do "pretty printing" of the summary (using colors).

devlead:
I understand that, colors rarely preserved on CI, so just an esthetic when running local then?

patriksvensson:
@devlead Yes, exactly. So the question is if we:

1. Exclude "delegating" tasks (so they won't show up in either log) 
2. Grey them out
3. Grey them out and add some description that tells the user that they just branched of to other tasks.

devlead:
Ok, I would vote hide by verbosity so they're shown if diagnostic.

patriksvensson:
@devlead That is a good idea. 

@ErikSchierboom What do you think about that?

ErikSchierboom:
[...]

PR #660, opened 2016-01-26
Don't show delegating tasks (#639)

ErikSchierboom:
This PR implements #639 to show delegating tasks based on verbosity level. At the moment, it only shows delegating tasks on the `Verbose` and `Diagnostic` levels. Furthermore, when a delegating task is displayed on the other levels, it is displayed in gray.

I've tagged it as a WIP due to the fact that I would like to discuss the design of the `CakeReport` class. At the moment, there are the following methods:

```csharp
public void Add(string task, TimeSpan span)
public void AddSkipped(string task)
public void AddDelegated(string task, TimeSpan span)
public void Add(string task, TimeSpan span, CakeTaskExecutionStatus executionStatus)
```

The first `Add` method actually calls the fourth method with the `CakeTaskExecutionStatus.Executed` value. I created the `AddSkipped` and `AddDelegated` methods to be more verbose, however they could also be removed and direct calls be made to the `Add` method that takes the `CakeTaskExecutionStatus`. Is that desirable? Or should I perhaps make the fourth `Add` method private?

If we decide to keep the `AddSkipped` and `AddDelegated` methods, the first `Add` task should be renamed to `AddExecuted`, right? However, that would break backwards-compatibility.

Secondly, the `CakeReportEntry` class has two constructors:

```csharp
public CakeReportEntry(string taskName, TimeSpan duration)
public CakeReportEntry(string taskName, TimeSpan duration, CakeTaskExecutionStatus executionStatus)
```

The first constructor just calls the second one. I've left it in there for backwards-compatibility. Should In leave it in or remove it?

### DIFF
--- a/src/Cake.Tests/Unit/CakeReportPrinterTests.cs
+++ b/src/Cake.Tests/Unit/CakeReportPrinterTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using Cake.Core;
+using Cake.Core.Diagnostics;
 using Cake.Testing;
 using NSubstitute;
 using Xunit;
@@ -14,8 +15,9 @@ namespace Cake.Tests.Unit
             public void Should_Throw_If_Report_Is_Null()
             {
                 // Given
+                var context = Substitute.For<ICakeContext>();
                 var console = Substitute.For<IConsole>();
-                var printer = new CakeReportPrinter(console);
+                var printer = new CakeReportPrinter(console, context);
 
                 // When
                 var result = Record.Exception(() => printer.Write(null));
@@ -28,13 +30,14 @@ namespace Cake.Tests.Unit
             public void Should_Default_To_30_Width()
             {
                 // Given
+                var context = Substitute.For<ICakeContext>();
                 var console = new FakeConsole();
                 var report = new CakeReport();
                 string taskName = "TaskName";
                 TimeSpan duration = TimeSpan.FromSeconds(10);
 
                 report.Add(taskName, duration);
-                var printer = new CakeReportPrinter(console);
+                var printer = new CakeReportPrinter(console, context);
 
                 // When
                 printer.Write(report);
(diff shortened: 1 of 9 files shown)


gep13:
@ErikSchierboom any chance you some screenshots of the output?  Yes, I am lazy :smile:

ErikSchierboom:
Here is an example that specifies the "Default" task as a delegating task:

```csharp
Task("Default")
	.IsDependentOn("Test");
```

This is the result when running the "Default" task with verbosity set to one of the three lowers levels:

image

In this mode, the delegating task is not shown.

With verbosity set to one of the two high levels:

image

Now, the delegating task is shown in grey.

patriksvensson:
Looks good to me. Could you rebase this against upstream/develop and i'll merge it. :+1:

ErikSchierboom:
@patriksvensson I have rebased it against the upstream/develop branch.

src/Cake.Tests/Unit/CakeReportPrinterTests.cs
```diff
@@ -6,6 +6,8 @@
 
 namespace Cake.Tests.Unit
 {
+    using Core.Diagnostics;
```
gep13: Can you put this using statement up beside the others?

src/Cake/CakeReportPrinter.cs
```diff
@@ -6,12 +6,16 @@
 
 namespace Cake
 {
+    using Core.Diagnostics;
```
gep13: Can you put this using statement up beside the others?

gep13:
@ErikSchierboom just had a quick look at this, and made another couple some comments.  Can you fix them up before we merge?

Thanks!

ErikSchierboom:
@gep13 Done!

gep13:
@ErikSchierboom thank you for working with us to get this PR fixed up before including into the codebase.  We really appreciate your patience!

ErikSchierboom:
@gep13 No problem at all, my pleasure!

Issue #512, opened 2015-11-11
Allow use of ICakeContext in WithCriteria 


jrnail23:
Overload `CakeTaskBuilder.WithCriteria` to accept `Func<ICakeContext,bool> `.
I'm factoring out some common behaviors in my scripts into reusable tasks, but not having the context available in `WithCriteria` is causing some friction.

Anyone have a reason that context shouldn't be allowed there?

patriksvensson:
Sounds like a reasonable addition. Will make the task API more consistent as well :+1:

ErikSchierboom:
I am looking into this, and was wondering how I should add it to the existing `CakeTask` class. This class has a property that contains all existing boolean criterias: `public IReadOnlyList<Func<bool>> Criterias`. In the `ShouldTaskExecute` method in the `CakeEngine` class, the aforementioned `Criterias` property is iterated over. Should I create an additional property which signature is like this: `public IReadOnlyList<Func<ICakeContext, bool>> ContextCriterias`? Or should `Func<bool>` instances be converted to `Func<ICakeContext, bool>` instances and the existing `Criterias` property thus return `IReadOnlyList<Func<ICakeContext, bool>>`?

patriksvensson:
@ErikSchierboom I would prefer if we changed `CakeTask.Criterias` to be a `IReadOnlyList<Func<ICakeContext, bool>>` and converted the `Func<bool>` to a `Func<ICakeContext, bool>`.

We could also make the `void AddCriteria(Func<bool> criteria)` an extension method (since it's more or less a convenience method) and change the implementation in `CakeTask` to `void AddCriteria(Func<ICakeContext, bool> criteria)`

patriksvensson:
Merged via #663.

PR #663, opened 2016-01-26
Allow use of ICakeContext in WithCriteria (#512)

ErikSchierboom:
This PR allows use of the `ICakeContext` in the `WithCriteria` extension method. As discussed in #512, the existing `Func<bool>` criteria still work, albeit now via an extension method.

### DIFF
--- a/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Cake.Core.Tests.Fixtures;
 using Xunit;
+using NSubstitute;
 
 namespace Cake.Core.Tests.Unit
 {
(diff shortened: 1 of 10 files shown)


src/Cake.Core.Tests/Unit/CakeEngineTests.cs
```diff
@@ -6,6 +6,9 @@
 
 namespace Cake.Core.Tests.Unit
 {
+    using Core.Diagnostics;
```
patriksvensson: This should be outside of the namespace.

patriksvensson:
@ErikSchierboom Sorry for not getting back sooner to you. You have a couple of pull request lined up so I thought I should mention that we haven't been neglecting anything on purpose. All of us have just been really busy the last couple of weeks.

Left some comments for you about this PR, but other than that I think it's ready to be merged :+1:

ErikSchierboom:
@patriksvensson No problem at all. I already suspected you were all busy. I have fixed the issue and did a rebase.

patriksvensson:
Merged! Thanks for this! :+1:

PR #665, opened 2016-01-27
GH1142: Option to keep the autogenerated NuSpec file

### LINKED ISSUE #1142: Add option to keep the autogenerated NuSpec file
For debugging purposes, it would be nice if the auto-generated NuSpec file could be kept so it can be inspected for errors that might occur during the NuGet pack phase.

asbjornu:
This PR writes the generated `.nuspec` file to debug output if the verbosity level is above Verbose, so it's easier to figure out problems (if any) in the generated file. Fixes #1142.


patriksvensson:
I would prefer if we added an option to not delete the temporary nuspec instead of dumping this to the log.

asbjornu:
@patriksvensson Sure. Would that option be a command line argument? What should happen to the saved file on next run? Delete it, or generate a new (uniquely named) one?

patriksvensson:
Should probably be an option to the pack settings. Maybe @gep13 or @devlead has some input on this.

gep13:
Agreed, I think this would best be served as a boolean property on the NuGetPackSettings class.  Which a user can set, if they want to keep the nuspec file around.  I would suggest that it should just replace the file each time it runs.

**NOTE:** Any implementation done for NuGetPack could equally be done on ChocolateyPack :+1:

devlead:
@asbjornu This pr is out-of-date with the develop branch and pending feedback. Could you please review, rebase and ping us when you think this PR is ready for review.

asbjornu:
@devlead I've rebased and done some poking around in the code now, and I can't seem to find the place the `.nuspec` file is actually deleted. So, is it? If not, perhaps we can just rework this PR to write out the path of the generated file and be done with it? No need to add an option for functionality that already exists, although unintentional. Or am I not looking in the right place, @gep13 & @patriksvensson?

gep13:
@asbjornu when you get back from vacation, can you take a look at this one again, and let's see if we can get this resolved/merged.  Thanks!

asbjornu:
@gep13 Back! :smile: I've rebased, but my previous question still stands: Where is the `.nuspec` file deleted? I can't find the code that performs the delete so it can be made optional.

patriksvensson:
@asbjornu You can find the deletion on line 130 in `NuGetPacker.cs`.

I don't want the PR to write the nuspec content  to the log, so if we can add a `KeepTempNuspec` (or similar) option to the settings class it would be awesome :smile:

asbjornu:
@patriksvensson Aha, there it is! PR rewritten, fixed, wrapped and should be ready to go, unless there's any comments to how I've implemented it.

src/Cake.Common/Tools/NuGet/Pack/NuGetPacker.cs
```diff
@@ -214,6 +214,12 @@ private ProcessArgumentBuilder GetArguments(FilePath filePath, NuGetPackSettings
                     settings.Properties.Select(property => string.Concat(property.Key, "=", property.Value))));
             }
 
+            // Keep temporarily autogenerated NuSpec file?
+            if (settings.KeepTemporaryNuSpecFile)
+            {
+                builder.Append("-KeepTemporaryNuSpecFile");
```
gep13: @asbjornu is there a reason that this is getting appended? It doesn't seem to be getting used anywhere.

asbjornu: Good question. I meant to add the parsing logic for it as well, but forgot since I didn't immediately figure out how it was done for `RequireLicenseAcceptance`. Pointers to how and where the argument parsing is performed?

gep13: @asbjornu ok, you lost me. What does that property have to do with this PR?  Bear in mind, any additions to the arguments builder will ultimately be passed through to the underlying tool, in this case NuGet.exe, and I don't think that `-KeepTemporaryNuSpecFile` is a command line argument for NuGet.exe, is it?  Therefore, the settings that you are passing in via the `NuGetPackSettings` is only used prior to calling NuGet.exe.  Or have I missed something?

asbjornu: Ah, I see. I understand why I lost you. I'm obviously in another planetary system. I'll get back to earth and remove this right away. :smile:

gep13: @asbjornu ha ha, someone had far too good a vacation I think :stuck_out_tongue:

gep13:
@asbjornu looks much better :smile: Was there an issue associated with this PR?  If not, can you create one, and reference it here? Thanks!

asbjornu:
@gep13 Nope, no issue. I just went right ahead with the PR.

gep13:
>@asbjornu said...
@gep13 Nope, no issue. I just went right ahead with the PR.

Could you create one?

asbjornu:
@gep13 #1142 created.

gep13:
@asbjornu perfect, thank you!

patriksvensson:
@asbjornu @gep13 Looks good to me!

gep13:
@asbjornu are you in a position to rebase this PR on the latest develop branch, and we can look to getting this pulled in? Thanks!

asbjornu:
@gep13: Done.

gep13:
@asbjornu I took the liberty of rebasing and merging this here: https://github.com/cake-build/cake/commit/c3be73d109d8500cb552f7775a3922cd8d5f904b

As a result, I will manually close this PR, however, your work has been correctly attributed, thanks!

Issue #567, opened 2015-12-08
Add GitVersion into build.cake


gep13:
Replacing the usage of the releasenotes.md

patriksvensson:
This should be closed now?

devlead:
Even if we're refactoring build.cake, GitVersion is present already, so I think we can close this issue right @gep13?

gep13:
Yes, I think we can close this one now.

PR #675, opened 2016-02-05
Add GitVersion to Cake Build Process

gep13:
This still needs some work, but the basics are there.

Thoughts?

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -22,9 +22,10 @@ var releaseNotes = ParseReleaseNotes("./ReleaseNotes.md");
 
 // Get version.
 var buildNumber = AppVeyor.Environment.Build.Number;
-var version = releaseNotes.Version.ToString();
-var semVersion = local ? version : (version + string.Concat("-build-", buildNumber));
-var milestone = string.Concat("v", version);
+GitVersion assertedVersions        = null;
+var version = string.Empty;
+var semVersion = string.Empty;
+var milestone = string.Empty;
 
 // Define directories.
 var buildDir = Directory("./src/Cake/bin") + Directory(configuration);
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -67,22 +68,59 @@ Task("Restore-NuGet-Packages")
...
 {
-    var file = "./src/SolutionInfo.cs";
-    CreateAssemblyInfo(file, new AssemblyInfoSettings {
-        Product = "Cake",
-        Version = version,
-        FileVersion = version,
-        InformationalVersion = semVersion,
-        Copyright = "Copyright (c) Patrik Svensson, Mattias Karlsson, Gary Ewan Park and contributors"
+    GitVersion(new GitVersionSettings {
+        UpdateAssemblyInfoFilePath = "./src/SolutionInfo.cs",
```
patriksvensson: This file is no longer created but but is instead updated by GitVersion?

build.cake
```diff
@@ -67,22 +68,59 @@ Task("Restore-NuGet-Packages")
...
+        UpdateAssemblyInfo = true,
+        OutputType = GitVersionOutput.BuildServer
     });
+    
+    version = EnvironmentVariable("GitVersion_MajorMinorPatch");
```
patriksvensson: Do we get the version from environment variables here?

patriksvensson:
@gep13 What I understand of it, it looks good to me code wise, but left a couple of questions just to make sure I understood it correctly :smile: 

GitVersion is not working on Mono yet?

build.cake
gep13: @patriksvensson that is correct.  Gitversion simply applies a regular expression update to the specified file, or it can loop through all files and look for AssemblyInfo.cs files.

build.cake
gep13: On some CI Servers, because of the way GitVersion is ran, the generated Environment Variables are available immediately after calling GitVersion. In the case of AppVeyor, which we are using, this is not the case.  In theory, we could remove this, and only ever call GitVersion twice, but I thought it best to leave this in, as it provides more information about the "intent".

gep13:
>@patriksvensson said...
GitVersion is not working on Mono yet?

Yes, this is correct.  I should probably add a WithCriteria to the GitVersion Tasks.

There are plans to make it xplat, but that won't be until the next major release, at the earliest.

gep13:
@patriksvensson this one is now ready for final review.  I kicked the Travis build, which seemed to have stalled.

patriksvensson:
Merged :+1:

gep13:
Relates to #567

PR #683, opened 2016-02-10
Ensure we also catch the possible FileNotFound exception that may be …

### LINKED ISSUE #682: ScriptAliasFinder fails on Mono when assembly isn't loaded in FindAliases
7dee2c6 changed the behavior of ScriptAliasFinder to log a warning when an assembly wasn't loaded. This works on Windows, but Mono throws a FileNotFound exception instead of a ReflectionTypeLoadException in this case.

I'll submit a pull request shortly.

nathansoz:
…thrown on mono

This should fix #682


devlead:
@nathansoz LGTM, waiting for CI to complete and I'll merge it's if successful .

devlead:
@nathansoz merged, thanks for your contribution :+1:

Issue #658, opened 2016-01-25
Add support for dotCover


Sam13:
Add support for coverage tool: dotCover
https://www.nuget.org/packages/JetBrains.dotCover.CommandLineTools/

Command line reference:
https://www.jetbrains.com/dotcover/help/dotCover__Console_Runner_Commands.html#analyse

Add in same way as OpenCover which enables to pass another tool (e.g. Nunit3) for execution

gep13:
@Sam13 thank you for the suggestion!  While we think that dotCover would be a great addition to Cake, we don't think that we are in a position to bring this directly into Cake.Common.  Instead, we suggest that this would be a great candidate as an Addin for Cake, similar to Cake.VsCode, Cake.Gitter, or Cake.Kudu.

All of the current Cake Addin's can be found on NuGet here:  https://www.nuget.org/packages?q=cake

Would creating an Addin for Cake be something that you are interested in?

patriksvensson:
@gep13 @Sam13 The reason for this is that dotCover is a proprietary product and testing this could be difficult for us.

Sam13:
@patriksvensson What do you mean by testing? 
dotCover is a commercial product but free of charge, isn't it?
See https://blog.jetbrains.com/dotnet/2015/10/14/details-on-dottrace-dotcover-dotmemory-licensing-changes/#clt

You can download the NuGet which contains the  command line tools and run dotCover console to create XML coverage reports which can be processed e.g. by ReportGenerator.
For this dotCover console I'd like to have Cake support as for OpenCover...

patriksvensson:
@Sam13 I stand corrected. I was under the impression that dotCover required a license to run from Jetbrains.

gep13:
@Sam13 oh! This is indeed very interesting!  I knew that you could use the ReSharper Command Line Tools, but I didn't know about dotCover being under the same license structure!

gep13:
@Sam13 thanks for the link to that blog post, this really is great news!

mholo65:
I could really see myself needing this in a not so distant future. I could start looking into this if no one else have already started this task?

Short-term goal: get **analyze** command working
Long-term goal: add support for other commands, eg: **cover**, **merge**, **report** and **delete**

mholo65:
Yes, analyze works in most scenarios while the others is needed for more complex setup, e.g. running coverage for multiple projects in separate steps.

phrusher:
[...]

PR #691, opened 2016-02-15
dotcover support

mholo65:
Still needs some work. Wanted to check if I'm on the right path here before doing more work. Plan is to also add support for other [commands](https://www.jetbrains.com/dotcover/help/dotCover__Console_Runner_Commands.html) than analyse, e.g. cover, merge, report and delete 

Fixes issue #658

### DIFF
--- a/src/Cake.Common/Tools/DotCover/Analyse/DotCoverAnalyseSettings.cs
+++ b/src/Cake.Common/Tools/DotCover/Analyse/DotCoverAnalyseSettings.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using Cake.Common.Tools.DotCover;
+using Cake.Core.IO;
+using Cake.Core.Tooling;
+
+namespace Cake.Common.Tools.DotCover.Analyse
+{
+    /// <summary>
+    /// Contains settings used by <see cref="DotCoverAnalyser" />.
+    /// </summary>
+    public sealed class DotCoverAnalyseSettings : ToolSettings
+    {
+        private readonly HashSet<string> _scope;
+        private readonly HashSet<string> _filters;
+        private readonly HashSet<string> _attributeFilters;
+
+        /// <summary>
+        /// Gets or sets the type of the report.
+        /// This represents the <c>/ReportType</c> option.
+        /// The Default value is <see cref="DotCoverReportType.XML"/>.
+        /// </summary>
+        public DotCoverReportType ReportType { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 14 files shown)


patriksvensson:
@mholo65 I think this looks great :+1: 

I think it would suffice with the Analyze command for this PR and add the others after that as separate PR:s to keep them focused.

What do you think?

mholo65:
Thanks, sounds good to me! I'll finalize this by adding support for more options and adding more unit tests. Expect an updated PR by the end of the week.

patriksvensson:
@mholo65 Sounds good to me :+1:

mholo65:
@patriksvensson I think this may be done for now, what do you think?

patriksvensson:
@mholo65 Will review it ASAP.

src/Cake.Common/Tools/DotCover/Analyse/DotCoverAnalyser.cs
```diff
@@ -0,0 +1,152 @@
+﻿using System;
+using Cake.Core;
+using Cake.Core.IO;
+
+namespace Cake.Common.Tools.DotCover.Analyse
+{
+    /// <summary>
+    /// DotCover Analyser builder.
+    /// </summary>
+    public class DotCoverAnalyser : DotCoverTool<DotCoverAnalyseSettings>
```
patriksvensson: Should be sealed.

src/Cake.Common/Tools/DotCover/Analyse/DotCoverAnalyseSettings.cs
```diff
@@ -0,0 +1,78 @@
...
+using Cake.Common.Tools.DotCover;
+using Cake.Core.IO;
+using Cake.Core.Tooling;
+
+namespace Cake.Common.Tools.DotCover.Analyse
+{
+    /// <summary>
+    /// Contains settings used by <see cref="DotCoverAnalyser" />.
+    /// </summary>
+    public class DotCoverAnalyseSettings : ToolSettings
```
patriksvensson: Should be sealed.

patriksvensson:
@mholo65 This looks good to me. Left some minor comments but otherwise this looks solid to me. Great job! :+1:

mholo65:
Thanks @patriksvensson, should be fixed now. Also fixed documentation on filters for alias DotCoverAnalyse since it was incorrect.

patriksvensson:
@mholo65 Merged!
Thank you for this! Awesome work! :+1:

PR #692, opened 2016-02-15
Add missing command line switch for ILRepack's targetplatform

dbent:
When building the command line arguments the `ILRepackRunner` was missing the `/targetplatform:` switch before the actual value, without it ILRepack thinks the value is an assembly it's trying to merge.

### DIFF
--- a/src/Cake.Common/Tools/ILRepack/ILRepackRunner.cs
+++ b/src/Cake.Common/Tools/ILRepack/ILRepackRunner.cs
@@ -127,7 +127,7 @@ namespace Cake.Common.Tools.ILRepack
 
             if (settings.TargetPlatform.HasValue)
             {
-                builder.Append(GetTargetPlatformString(settings.TargetPlatform.Value));
+                builder.AppendSwitch("/targetplatform", ":", GetTargetPlatformString(settings.TargetPlatform.Value));
             }
 
             if (settings.XmlDocs)


patriksvensson:
@dbent Will take a look at this later today :+1:

gep13:
@patriksvensson I was actually just looking at this.  My initial thinking was that this was going to be covered by a test, but looks like we have ILMergeRunnerTests, but we don't have ILRepackRunnerTests.  And where ILMergeRunner does this:

https://github.com/cake-build/cake/blob/ca0c3c35d63672d82cd3ee27c387522214854b91/src/Cake.Common/Tools/ILMerge/ILMergeRunner.cs#L131

There is no equivalent in the ILRepackRunner.  Looks like we need to add some more tests.

patriksvensson:
@gep13 Maybe we should add an issue to add test coverage to the affected class?

src/Cake.Common/Tools/ILRepack/ILRepackRunner.cs
```diff
@@ -127,7 +127,7 @@ protected override IEnumerable<string> GetToolExecutableNames()
 
             if (settings.TargetPlatform.HasValue)
             {
-                builder.Append(GetTargetPlatformString(settings.TargetPlatform.Value));
+                builder.Append("/targetplatform:" + GetTargetPlatformString(settings.TargetPlatform.Value));
```
patriksvensson: I would prefer if we used `builder.AppendSwitch("/targetplatform:", GetTargetPlatformString(settings.TargetPlatform.Value))` to be more in line with other tools.

gep13:
@patriksvensson agreed. I have created #694 so that we don't forget about this.

dbent:
@patriksvensson Updated to use `AppendSwitch()` but I had to use the overload with `seperator` as `string.Empty` otherwise ILRepack's argument parsing breaks if there's a space between the switch and value: `/targetplatform: v4`:
````
INFO: IL Repack - Version 2.0.10
System.ArgumentOutOfRangeException: startIndex cannot be larger than length of string.
Parameter name: startIndex
   at System.String.Substring(Int32 startIndex, Int32 length)
   at ILRepacking.ILRepack.ResolveTargetPlatformDirectory(String version)
   at ILRepacking.ILRepack.ResolveSearchDirectories()
   at ILRepacking.ILRepack.Repack()
   at ILRepacking.Application.Main(String[] args)
```

patriksvensson:
@dbent Ah I see, didn't think of that :smile:
You could also set the `:` as the separator.

dbent:
> You could also set the : as the separator.

Ah, why didn't I think of that? Done.

gep13:
@dbent would you be in a position to squash these commits into a single commit?  Thanks!

dbent:
@gep13 Done.

gep13:
@dbent perfect! Once the CI build passes, I will get this merged in.

gep13:
@dbent thank you again for the contribution, and for making the suggested changes to your PR, your help is very much appreciated!

Issue #709, opened 2016-02-26
TextTransformation.Save creates BOM on new file


andycmaj:
even though the input template didn't have one, the resulting file now does.
We should use `new System.Text.UTF8Encoding(false)` to make sure we don't add the BOM...

## Before

```csharp
            using (var writer = new StreamWriter(stream, Encoding.UTF8, 1024, true))
            {
                writer.Write(_template.Render());
            }
```

## After

```csharp
            using (var writer = new StreamWriter(stream, new System.Text.UTF8Encoding(false), 1024, true))
            {
                writer.Write(_template.Render());
            }
```

I'll submit a PR

andycmaj:
released with v0.10.0

PR #710, opened 2016-02-26
don't add BOM on text transformations (#709)

andycmaj:
fix for https://github.com/cake-build/cake/issues/709

`TextTransform.Save` inadvertently adds a BOM to the resulting file.
would be nice to prevent this as it can result in unexpected modifications of the input template.

change TextTransformation.Save to omit the BOM, and add a unit test to verify that.
Unit tests failed before fix ("Hello world" file length was returned as 14 instead of 11), and passed after fix.

### DIFF
--- a/src/Cake.Common/Text/TextTransformation.cs
+++ b/src/Cake.Common/Text/TextTransformation.cs
@@ -64,7 +64,7 @@ namespace Cake.Common.Text
             // Render the content to the file.
             var file = _fileSystem.GetFile(path);
             using (var stream = file.OpenWrite())
-            using (var writer = new StreamWriter(stream, Encoding.UTF8, 1024, true))
+            using (var writer = new StreamWriter(stream, new UTF8Encoding(false), 1024, true))
             {
                 writer.Write(_template.Render());
             }
--- a/src/Cake.Common.Tests/Unit/Text/TextTransformationTests.cs
+++ b/src/Cake.Common.Tests/Unit/Text/TextTransformationTests.cs
@@ -44,15 +44,18 @@ namespace Cake.Common.Tests.Unit.Text
             public void Should_Render_Content_To_File()
             {
                 // Given
+                var expectedContent = "Hello World";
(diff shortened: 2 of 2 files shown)


src/Cake.Common/Text/TextTransformation.cs
```diff
@@ -64,7 +64,7 @@ public void Save(FilePath path)
             // Render the content to the file.
             var file = _fileSystem.GetFile(path);
             using (var stream = file.OpenWrite())
-            using (var writer = new StreamWriter(stream, Encoding.UTF8, 1024, true))
+            using (var writer = new StreamWriter(stream, new UTF8Encoding(false), 1024, true))
```
patriksvensson: So the false parameter here just tells the StreamWriter to preserve BOM (if any) and not add one if none already exist in the stream?

andycmaj: not exactly... the `false` param tells `StreamWriter` NOT to emit the BOM when writing the stream.
A `Stream` itself never implicitly has a BOM (ie. you won't see the BOM in the stream's bytes no matter what). The BOM comes from the `Encoding`'s specific `GetPreamble` method.

Strangely, when you use `Encoding.UTF8`, it uses the constructor that explicitly ADDS the preamble (BOM): http://referencesource.microsoft.com/#mscorlib/system/text/encoding.cs,1547.

But the DEFAULT constructor for `UTF8Encoding` explicitly DOES NOT add the BOM: http://referencesource.microsoft.com/#mscorlib/system/text/utf8encoding.cs,65.

In my opinion, you'll have more success across various platforms if you don't add the BOM to the written file in this case.

devlead:
@andycmaj looks good to me :+1:  Could you rebase it against latest develop please.

andycmaj:
sure.

andycmaj:
Done. rebased on upstream/develop.

devlead:
@andycmaj Merged :+1:  Thanks for contributing!

PR #712, opened 2016-02-26
Add MSTest option for providing the tool (MSTest.exe) location

brettkc:
I was trying to use Cake to setup a build process and ran into issues running the MSTest tool. The issue is my Visual Studio is installed in an alternate location so Cake wouldn't find my MSTest.exe. I've altered it to allow a user defined path to be passed to MSTestSettings.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSTest/MSTestRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSTest/MSTestRunnerTests.cs
@@ -106,6 +106,23 @@ namespace Cake.Common.Tests.Unit.Tools.MSTest
             Assert.Equal(existingToolPath, result.Path.FullPath);
         }
 
+        [Theory]
+        [InlineData("U:/ProgramFilesX86/Microsoft Visual Studio 14.0/Common7/IDE/mstest.exe")]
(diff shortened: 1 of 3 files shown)


patriksvensson:
@brettkc Thanks for your contribution, but if you want to provide the location to `mstest.exe`, this is already possible via the `ToolPath` property on `ToolSettings` (which `MSBuildSettings` inherit from). 

```csharp
MSTest("./**/*.Tests.dll", new MSTestSettings
{
  ToolPath = "./tools/mstest.exe"
});
```

brettkc:
Ah, missed that -- thanks.

patriksvensson:
@brettkc No problem. We should make this more clear in the documentation. Updating it now :smile:

patriksvensson:
Added a fix for the homepage to show inherited methods and properties: http://cakebuild.net/api/cake.common.tools.msbuild/04799b3e

brettkc:
Thanks @patriksvensson. Slightly related issue, would it make sense to add a TestCategory setting to the MSTestSettings object? (see: https://msdn.microsoft.com/en-us/library/ms182489.aspx#category).

patriksvensson:
@brettkc Absolutely! Create a new issue for that so we're able to track it :smile:

PR #713, opened 2016-02-26
Added DotCover Cover command

mholo65:
Added DotCover Cover command. Needed since TeamCity only want snapshots, not HTML/XML reports.

If possible, to remove duplicate code, I would like to reuse **DotCoverCoverSettings** in **DotCoverAnalyser** by having **DotCoverAnalyseSettings** to inherit from **DotCoverCoverSettings**.  Is this allowed since it will make documentation unclear?

NOTE: Quickly hacked using my Linux box, need to verify functionality on Windows (dotCover.exe doesn't run under Mono). But this have to wait until monday when I'm back at work.

//cc @patriksvensson

PR #714, opened 2016-02-26
Added support for importing coverage to TeamCity

mholo65:
This is needed to import to import Coverage to TeamCity. See https://confluence.jetbrains.com/display/TCD9/Manually+Configuring+Reporting+Coverage

Note, quick hack, I would like to test functionality on monday when I have access to TeamCity.

Also, unit tests are non-existent for TeamCity, wan't to add these also prior to merging this PR.

related to #713 

//cc @patriksvensson

### DIFF
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -62,6 +62,7 @@
     <Compile Include="Fixtures\Build\BambooFixture.cs" />
     <Compile Include="Fixtures\Build\ContinuaCIInfoFixture.cs" />
     <Compile Include="Fixtures\Build\BambooInfoFixture.cs" />
+    <Compile Include="Fixtures\Build\TeamCityFixture.cs" />
     <Compile Include="Fixtures\AssemblyInfoFixture.cs" />
     <Compile Include="Fixtures\AssemblyInfoParserFixture.cs" />
     <Compile Include="Fixtures\Build\JenkinsFixture.cs" />
--- a/src/Cake.Common.Tests/Fixtures/Build/TeamCityFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Build/TeamCityFixture.cs
@@ -1,0 +1,28 @@
+using Cake.Common.Build.TeamCity;
+using Cake.Core;
+using NSubstitute;
+
+namespace Cake.Common.Tests.Fixtures.Build
+{
+    internal sealed class TeamCityFixture
+    {
+        public ICakeEnvironment Environment { get; set; }
+
+        public TeamCityFixture()
+        {
+            Environment = Substitute.For<ICakeEnvironment>();
+            Environment.WorkingDirectory.Returns("C:\\build\\CAKE-CAKE-JOB1");
+            Environment.GetEnvironmentVariable("TEAMCITY_VERSION").Returns((string)null);
+        }
+
+        public void IsRunningOnTeamCity()
+        {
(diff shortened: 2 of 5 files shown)


mholo65:
@patriksvensson I have given this some thought. TeamCity supports three different coverage tools (dotcover, partcover and ncover) of which one is supported in Cake (dotcover). Therefore this could be implemented less generic and maybe squash both commands into one, e.g:
```csharp
public void ImportDotCoverCoverage(FilePath resultFile, FilePath toolPath == null)
```
that would output the following service messages
```
##teamcity[dotNetCoverage dotcover_home='/path/to/dotCover.exe']
##teamcity[importData type='dotNetCoverage' tool='dotcover' path='/path/to/result.dcvr']
```
If toolPath is set to null, then TeamCity would use the bundled dotCover.

Or
```csharp
public void ImportDotCoverCoverage(FilePath resultFile, bool useBundled = false)
```
which would resolve the dotCover path if useBundled is false, otherwise TeamCity bundled would be used.

patriksvensson:
@mholo65 Good question. I lean towards the second alternative since that seems to be the nicest to the end user of the alias, but I'm not sure to be honest.

Does the rest of the @cake-build/team have an opinion about this?

mholo65:
It might also be wise to add the possibility to specify the ToolPath just like every Cake tool's path can be specified via ToolSettings.ToolPath.

Anyhow, for resolution of the tool's path, is there any common way to do this in Cake? Tools internally uses GetToolPath() method, but this is private. Could it be beneficial to have an alias for this?

patriksvensson:
@mholo65 There are plans to expand on the tool resolution strategy and make it available for scripts as well, but nothing yet I'm afraid. Maybe the first alternative is the best until we've done this?

mholo65:
Sounds OK! I'll update the PR in a couple of days.

mholo65:
@patriksvensson this is also done. Could you have a look?

patriksvensson:
@mholo65 From a code perspective this looks good to me :+1: 
Have you tested it on TeamCity?

Could yoe rebase this against develop and I'll merge it. Thanks!

mholo65:
@patriksvensson I have not tested on TeamCity. I have however verified that the TeamCity method outputs same service messages as I have manually built up using *Console.WriteLine()*

Rebased and all checks have passed!

patriksvensson:
@mholo65 It would be great if you could run this on TeamCity to verify that it works as expected before we merge it in.

mholo65:
@patriksvensson yes, I can have it tested, but it'll have to wait until Wednesday or Thursday when I've access to TeamCity again.

Until now I've only verified it locally, and it output the exact same service messages as we are currently using in our CI environment (i.e. *Console.WriteLine()* directly from Cake-script).

gep13:
@mholo65 is there any chance you can rebase this again against develop, so that we can get it in a position to merge in? Thanks!

mholo65:
@gep13 sure can do! But @patriksvensson wanted me to verify on TeamCity first, and I'm planning on doing that tomorrow.

gep13:
@mholo65 that sounds perfect. If you can do that verification, and let us know how that goes, we can than get this rebased, and merged in.

mholo65:
@gep13 @patriksvensson Rebased and verified with TeamCity, works as expected.

devlead:
@mholo65 merged, thanks for your contribution :+1:

gep13:
@mholo65 thanks for your help on this one, and for going the extra mile to help with verification :+1:

mholo65:
@gep13 np, only glad to help!

PR #719, opened 2016-02-29
Fix for PathTests Segements property test

### LINKED ISSUE #718: Path segment bug (or test bug, choose your own adventure!)
Found a bug in either the `Cake.Core.IO.Path` segments calculation *or* the `PathTests.TheSegmentsProperty.Should_Return_Segments_Of_Path()` test.

The test is a theory,  but the data isn't used in the test - instead it uses a hardcoded path:
```
[Theory]
[InlineData("Hello/World")]
[InlineData("/Hello/World")]
[InlineData("/Hello/World/")]
[InlineData("./Hello/World/")]
public void Should_Return_Segments_Of_Path(string pathName)
{
    // Given
    var path = new TestingPath("Hello/World");

    // When, Then
    Assert.Equal(2, path.Segments.Length);
    Assert.Equal("Hello", path.Segments[0]);
    Assert.Equal("World", path.Segments[1]);
}
```

So that's obviously wrong. When you change the test to use `pathName`, "/Hello/World" and "/Hello/World/" fail because the segment logic looks like this:
```
_segments = _path.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
if (_path.StartsWith("/") && _segments.Length > 0)
{
    _segments[0] = "/" + _segments[0];
}
```

I.e., if the path starts with "/", then "/" is prepended to the first segment.

So either the test is wrong or the logic is wrong and it just wasn't showing up because the test was wrong. I can take care of it - just need guidance on what the intent is here. Which is correct, the test or the logic?

daveaglick:
Resolves #718

### DIFF
--- a/src/Cake.Core.Tests/Unit/IO/PathTests.cs
+++ b/src/Cake.Core.Tests/Unit/IO/PathTests.cs
@@ -118,17 +118,29 @@ namespace Cake.Core.Tests.Unit.IO
         {
             [Theory]
             [InlineData("Hello/World")]
-            [InlineData("/Hello/World")]
-            [InlineData("/Hello/World/")]
             [InlineData("./Hello/World/")]
             public void Should_Return_Segments_Of_Path(string pathName)
             {
                 // Given
-                var path = new TestingPath("Hello/World");
+                var path = new TestingPath(pathName);
 
                 // When, Then
                 Assert.Equal(2, path.Segments.Length);
                 Assert.Equal("Hello", path.Segments[0]);
+                Assert.Equal("World", path.Segments[1]);
+            }
+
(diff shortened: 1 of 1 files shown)


patriksvensson:
@daveaglick Oh, thank you for this. Didn't expect of you to fix my bug when I wrote the reply to the issue :smile: Thank you very much!

patriksvensson:
@daveaglick Merged! :+1:

daveaglick:
It was very difficult. I mean, I almost forgot to hit that Ctrl-V after the Ctrl-C :smile:

patriksvensson:
Haha, yeah. Thanks again!

Issue #695, opened 2016-02-16
Add SpecFlow support


mholo65:
Add support for [SpecFlow](https://github.com/techtalk/SpecFlow). Commands should contains features like running tests (via NUnit/MSTest/xUnit) and generating reports.

I'm willing to implement and send a PR if it's suitable for Cake.Common.

gep13:
@mholo65 thanks for reaching out!

Yes, we agree that this would be a good addition for Cake.Common.  If you are in a position to get that working, we would gladly accept a PR for this.

devlead:
Fixed by #727

mholo65:
Thanks for merging. What about adding label and milestone to issue?

devlead:
@mholo65 added :+1:  It's merged so it'll be in the next one :wink:

PR #727, opened 2016-03-01
Added SpecFlow support

mholo65:
Added support for SpecFlow. Solves issue #695

At first I thought it would be nice if SpecFlow tool could intercept the Unit Test runner (NUnit/MSTest) just like OpenCover/DotCover, but SpecFlow sets some requirements on the arguments for the Unit Test runner. So it might be simpler to leave it as is.

Unit Tests still to be written.

### DIFF
--- a/src/Cake.Common/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporter.cs
+++ b/src/Cake.Common/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporter.cs
@@ -1,0 +1,114 @@
+using System;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Cake.Common.Diagnostics;
+using Cake.Core;
+using Cake.Core.IO;
+
+namespace Cake.Common.Tools.SpecFlow.TestExecutionReport
+{
+    /// <summary>
+    /// SpecFlow MSTest execution report runner.
+    /// </summary>
+    public sealed class SpecFlowTestExecutionReporter : SpecFlowTool<SpecFlowTestExecutionReportSettings>
+    {
+        private readonly ICakeEnvironment _environment;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SpecFlowTestExecutionReporter" /> class.
+        /// </summary>
+        /// <param name="fileSystem">The file system.</param>
+        /// <param name="environment">The environment.</param>
+        /// <param name="processRunner">The process runner.</param>
+        /// <param name="globber">The globber.</param>
+        public SpecFlowTestExecutionReporter(
+            IFileSystem fileSystem,
+            ICakeEnvironment environment,
+            IProcessRunner processRunner,
+            IGlobber globber)
+            : base(fileSystem, environment, processRunner, globber)
+        {
+            _environment = environment;
+        }
+
+        /// <summary>
+        /// Runs SpecFlow Test Execution Report with the specified settings.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="action">The action.</param>
+        /// <param name="projectFile">The project file path.</param>
+        /// <param name="settings">The settings.</param>
+        public void Run(ICakeContext context,
+            Action<ICakeContext> action,
+            FilePath projectFile,
+            SpecFlowTestExecutionReportSettings settings)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+            if (action == null)
+            {
+                throw new ArgumentNullException("action");
+            }
+            if (projectFile == null)
+            {
+                throw new ArgumentNullException("projectFile");
+            }
+            if (settings == null)
+            {
+                throw new ArgumentNullException("settings");
+            }
+
+            // Run the tool using the interceptor.
+            var interceptor = InterceptAction(context, action);
+
+            // Get / Verify Arguments
+            var builder = GetArguments(interceptor, settings, projectFile);
+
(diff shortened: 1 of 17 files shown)


patriksvensson:
@mholo65 I see.

Couldn't you use an interceptor but have different settings classes? You could then let the settings classes like `NUnitSpecFlowSettings`/`XUnitSpecFlowSettings` (which all derive from `SpecFlowSettings`) validate and render the arguments for the SpecFlow tool runner.

mholo65:
@patriksvensson That could actually work.

By intercepting the tool, there shouldn't be any need for separate MSBuild/NUnit/XUnit aliases nor separate Settings classes, one alias for [Test Execution reports](https://github.com/techtalk/SpecFlow/wiki/Reporting#test-execution-report) should be enough.

Example:
```csharp
public static void SpecFlowExecutionReport(
       Action<ICakeContext> action,
       FilePath projectFile,
       SpecFlowSettings settings
)
```
The SpecFlow runner should intercept the action (accepting only NUnit/MSTest/XUnit), examine the arguments (and throw exception/try to fix if invalid) and set SpecFlow parameters accordingly.

We'll still need an separate alias for the [Step Definition Report](https://github.com/techtalk/SpecFlow/wiki/Reporting#step-definition-report) though.

patriksvensson:
@mholo65 Sounds good to me. Implementing the step definition report should probably be done in a separate PR though?

mholo65:
@patriksvensson could be a separate PR. altough I've implemented it already in this PR :)

patriksvensson:
@mholo65 Ah, but then proceed as normal :smile:

mholo65:
@patriksvensson that should be it! Could You have a look? When ok, I'll rebase.

patriksvensson:
@mholo65 Sure, I'll take a look later tonight.

src/Cake.Common/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporter.cs
```diff
@@ -0,0 +1,220 @@
...
+                var testLabelsMatch = Regex.Match(arguments, "--labels=All", RegexOptions.IgnoreCase);
+
+                if (!testLabelsMatch.Success)
+                {
+                    throw new CakeException("NUnit3 must contain argument \"--labels=All\"");
+                }
+            }
+            else if (executable.IndexOf("xunit", StringComparison.OrdinalIgnoreCase) >= 0)
+            {
+                builder.Append("nunitexecutionreport");
```
patriksvensson: Is this correct?

src/Cake.Common/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporter.cs
```diff
@@ -0,0 +1,220 @@
...
+
+                // Set the project file
+                builder.AppendQuoted(projectFile.MakeAbsolute(_environment).FullPath);
+
+                // Set the test result
+                var arguments = GetToolArguments(context);
+
+                var xmlTestResultMatch = Regex.Match(arguments, "-nunit\\s+((?:\".+?\"|.*))", RegexOptions.IgnoreCase);
```
patriksvensson: Is this correct?

src/Cake.Common/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporter.cs
```diff
@@ -0,0 +1,220 @@
...
+
+                if (xmlTestResultMatch.Success && xmlTestResultMatch.Groups[1].Success)
+                {
+                    builder.AppendSwitch("/xmlTestResult", ":", xmlTestResultMatch.Groups[1].Value);
+                }
+                else
+                {
+                    throw new CakeException("XUnit2 must contain argument \"-nunit <filename>\"");
```
patriksvensson: Is this correct?

src/Cake.Common/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporter.cs
```diff
@@ -0,0 +1,220 @@
...
+        private ProcessArgumentBuilder GetArguments(
+            SpecFlowContext context,
+            SpecFlowTestExecutionReportSettings settings,
+            FilePath projectFile)
+        {
+            var builder = new ProcessArgumentBuilder();
+
+            var executable = context.FilePath.ToString();
+
+            if (executable.IndexOf("mstest", StringComparison.OrdinalIgnoreCase) >= 0)
```
patriksvensson: Maybe move the content of the three if-statements to their own methods in a separate static class. Would make it easier to follow the code. What do you think?

patriksvensson:
@mholo65 Left some minor comments, but other than that it looks good to me.

@devlead @gep13 Any other comments about this one?

mholo65:
@patriksvensson SpecFlow is able to parse test results from either MSTest or NUnit (v2), thus making it compatible with xUnit if the runner is configured to write output in NUnit (v2) format. See #729 

The three if-statements could be moved to a separate internal class to enhance readability/maintainability. I will do this + add code examples to documentation for aliases.

Any other comments from @devlead or @gep13?

mholo65:
@patriksvensson Rebased and moved the argument generation based on Unit Test provider to a separate class.

devlead:
LGTM :+1: I'll do some local testing and then merge if those go ok.

Issue #733, opened 2016-03-04
Use V3 Nuget in bootstrapper


devlead:
Currently installing Cake and it's dependencies fail if a machine only has V3 nuget sources. As Nuget V3 is backwards compatible we should change which version of nuget downloaded in the bootstrappers.

PR #734, opened 2016-03-04
Changes nuget version bootstrapper downloads

devlead:
This is a suggested fix for #733

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -15,7 +15,7 @@ $PSScriptRoot = split-path -parent $MyInvocation.MyCommand.Definition;
 $TOOLS_DIR = Join-Path $PSScriptRoot "tools"
 $NUGET_EXE = Join-Path $TOOLS_DIR "nuget.exe"
 $CAKE_EXE = Join-Path $TOOLS_DIR "Cake/Cake.exe"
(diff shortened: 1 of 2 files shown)


daveaglick:
:+1: I did this manually in my own bootstrapper scripts, but it'll be good to get the change in the official ones.

RichiCoder1:
Hmm. It might be ideal to lock it to a specific version rather than pulling latest, just in case.

gep13:
@devlead what are your thoughts on @RichiCoder1 comment above?

devlead:
Personally I prefer latest stable, it's potentially less maintenance but also with potentially greater risk. I'm fine with either, regardless we would need to test & keep up with new versions.

gep13:
>@devlead said...
regardless we would need to test & keep up with new versions.

Yes, I agree with that.  And at least pulling in the latest will make us proactive at fixes any issues with our own build.

@patriksvensson what are your thoughts?  If you are in agreement, I am happy to merge this in.

patriksvensson:
@gep13 I understand the reservations but this is fine by me. Easy to change if it wouldn't work for some reason.

gep13:
:+1:

daveaglick:
:+1:

PR #739, opened 2016-03-07
Adding dependencies to NuGetPackSettings

RossHammer:
This added the property Dependencies to NuGetPackSettings allowing dependencies to be specified in the cake file.

### DIFF
--- a/src/Cake.Common.Tests/Properties/Resources.Designer.cs
+++ b/src/Cake.Common.Tests/Properties/Resources.Designer.cs
@@ -250,6 +250,25 @@ namespace Cake.Common.Tests.Properties {
         
         /// <summary>
         ///   Looks up a localized string similar to &lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;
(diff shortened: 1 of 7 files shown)


devlead:
@RossHammer could you create/reference an issue & rebase PR against latest version and I'll review the PR.

RossHammer:
I created the issue #740  to reference and is should be up to date

devlead:
@RossHammer thanks for contributing and creating an issue. Please rebase against upstream develop so it's only one commit in PR and let me know when it's done or if you need assistance.

RossHammer:
I think I got it now. If it's not let me know because I'm not super familiar with git.

devlead:
@RossHammer LGTM :+1: I'll merge.Thanks for your contribution!

daveaglick:
Thanks @RossHammer! I was about to open an issue for this, but for the second time today I discovered that the community beat me to it.

fleed:
Is it possible to reference packages "in the same solution"?

Let's say that I have my solution with projects `A` and `B`. `B` depends on `A`. I'd like to use cake to build two distinct packages `A` and `B`, with `B` having a reference to package `A` and not to the assembly generated by `A`. A good example is any AspNet core repository like [MVC](https://github.com/aspnet/Mvc)
Can I achieve this with dependencies and a single cake build?

PR #767, opened 2016-03-17
Fix TeamCity tests on TeamCity

### LINKED ISSUE #766: TeamCity ImportDotCoverCoverage tests fail when running on TeamCity
TeamCityProvider [WriteServiceMessage](https://github.com/cake-build/cake/blob/77430f25a454424650240ea3522cf3814e4c8abd/src/Cake.Common/Build/TeamCity/TeamCityProvider.cs#L253) works directly with the Console, this isn't very _"mockable"_, because this the [TheImportDotCoverCoverageMethod](https://github.com/cake-build/cake/blob/77430f25a454424650240ea3522cf3814e4c8abd/src/Cake.Common.Tests/Unit/Build/TeamCity/TeamCityProviderTests.cs#L64) hi-jacks the console stream.

This works an all build systems except TeamCity were these service messages have a meaning.

I'll shorty submit a PR with a suggested fix for this.

devlead:
This fixes #766 by not using System.Console directly and instead using ICakeLog and mock that during tests.
* Change TeamCityProvider WriteServiceMessage to use ICakeLog instead of System.Console
* Change TeamCityProviderTests to not change System.Console stream so TeamCityFixture instead has a FakeLog/ICakeLog instance
* Add AggregateLogMessages method to FakeLog to get all current log entries message as a string

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Build/TeamCityFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Build/TeamCityFixture.cs
@@ -1,5 +1,7 @@
 using Cake.Common.Build.TeamCity;
 using Cake.Core;
+using Cake.Core.Diagnostics;
+using Cake.Testing;
 using NSubstitute;
 
 namespace Cake.Common.Tests.Fixtures.Build
@@ -7,12 +9,14 @@ namespace Cake.Common.Tests.Fixtures.Build
     internal sealed class TeamCityFixture
     {
         public ICakeEnvironment Environment { get; set; }
+        public FakeLog Log { get; set; }
 
         public TeamCityFixture()
         {
             Environment = Substitute.For<ICakeEnvironment>();
             Environment.WorkingDirectory.Returns("C:\\build\\CAKE-CAKE-JOB1");
             Environment.GetEnvironmentVariable("TEAMCITY_VERSION").Returns((string)null);
+            Log = new FakeLog();
         }
 
         public void IsRunningOnTeamCity()
@@ -22,7 +26,7 @@ namespace Cake.Common.Tests.Fixtures.Build
 
         public TeamCityProvider CreateTeamCityService()
         {
-            return new TeamCityProvider(Environment);
+            return new TeamCityProvider(Environment, Log);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


RichiCoder1:
I feel like there was a reason I didn't do this. Has this been tested against a teamcity instance?

devlead:
@RichiCoder1 I'm about to add so PRs are built in TeamCity too and will soon test against my branch. It it doesn't work, I'll look at refactoring some more.

gep13:
@devlead I haven't seen any failed build notifications from TeamCity, so I think this worked, right?  Or is TeamCity only running builds on master branch just now?

RichiCoder1:
The tell would be wether service messages are properly showing up in team city build log

devlead:
@gep13 currently only runs develop / master, I'll add  so it builds PRs tonight. 
@RichiCoder1 correct I'm running a few tests on my fork on a local TC then I'll test the PR.

gep13:
@devlead cool. Are you planning on adding that as a PR check as well?

devlead:
@gep13 no, but that would fit well with the integration tests though, so given pr when that's in place.

mholo65:
@gep13 @devlead It should be quite straightforward with [this](https://confluence.jetbrains.com/display/TW/Commit+Status+Publisher) plugin. I'm using it myself for Gerrit integration. It doesn't come with any documentation though :smile: *hint*, once installed, it can be found under *Build Features*

gep13:
@mholo65 we are using the TeamCity instance hosted on CodeBetter, so I think we are limited in what plugins we can install, but I am fairly sure that setting up integration with GitHub PR's is already there, as we do that on Chocolatey already.

devlead:
@gep13 yes it's there just a little configuration.

devlead:
@gep13: Test went well should be good to go.

src/Cake.Testing/FakeLog.cs
```diff
@@ -37,6 +39,18 @@ public Verbosity Verbosity
         }
 
         /// <summary>
+        /// Aggregates all current log entries message
+        /// </summary>
+        /// <returns>Log messages as <see cref="System.String"/></returns>
+        public string AggregateLogMessages()
```
patriksvensson: Can we change this to an extension method in the Cake.Common.Tests assembly instead since Cake.Testing is a public assembly?

devlead: Sure, I'll fix.

devlead: Adjusted @patriksvensson

devlead:
@gep13 Fixed @patriksvensson extension feedback and rebased, so should be ready for merge unless you find something else.

gep13:
Looks good to me! :+1:

PR #770, opened 2016-03-18
Added build servers to README.md

devlead:
I propose we add the build servers that the Cake repository currently continuously builds on.


gep13:
I approve of this proposal! :+1:

PR #775, opened 2016-03-23
Added StorePasswordInClearText to NuGetSourcesSettings

devlead:
* Enables possibility to add NuGet source with password stored in plain text
* Fixes GH774


patriksvensson:
Looks good to me! :+1:

Issue #791, opened 2016-03-26
Add FileSize alias


devlead:
Add convenient way to check the size of a file.

PR #792, opened 2016-03-26
Add FileSize alias

devlead:
Relates to #791

```csharp
FileSize("./build.cake");
```

### DIFF
--- a/src/Cake.Common.Tests/Unit/IO/FileAliasesTests.cs
+++ b/src/Cake.Common.Tests/Unit/IO/FileAliasesTests.cs
@@ -1129,5 +1129,66 @@ namespace Cake.Common.Tests.Unit.IO
                 Assert.Equal("/Working/build.txt", result.FullPath);
             }
         }
+
+        public sealed class TheFileSizeMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Context_Is_Null()
+            {
+                // Given, When
+                var result = Record.Exception(() => FileAliases.FileSize(null, "some file"));
+
+                // Then
+                Assert.IsArgumentNullException(result, "context");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Path_Is_Null()
+            {
+                // Given
+                var context = Substitute.For<ICakeContext>();
+
+                // When
+                var result = Record.Exception(() => FileAliases.FileSize(context, null));
+
+                // Then
+                Assert.IsArgumentNullException(result, "filePath");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Directory_Does_Not_Exist()
+            {
+                // Given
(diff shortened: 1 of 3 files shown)


src/Cake.Common.Tests/Unit/IO/FileAliasesTests.cs
```diff
@@ -1129,5 +1129,86 @@ public void Should_Return_Absolute_Directory_Path()
...
+                context.Environment.Returns(environment);
+
+                // When
+                var result = FileAliases.FileSize(context, "non-existent-file.txt");
+
+                // Then
+                Assert.Equal(result, -1);
+            }
+
+            [Fact]
```
patriksvensson: This test is identical to `Should_Return_Size_If_Absolute_Path_Exist()`. Can we make this a theory instead since the  only thing that differ is the path (absolute or relative)?

src/Cake.Common/IO/FileAliases.cs
```diff
@@ -331,5 +331,35 @@ public static FilePath MakeAbsolute(this ICakeContext context, FilePath path)
...
+            }
+
+            if (filePath == null)
+            {
+                throw new ArgumentNullException("filePath");
+            }
+
+            var file = context.FileSystem.GetFile(filePath.MakeAbsolute(context.Environment));
+
+            return file.Exists ? file.Length : -1;
```
patriksvensson: Personally I think that this should throw a FileNotFoundException if the file does not exist (just like [System.IO.File.Length](https://msdn.microsoft.com/en-us/library/system.io.fileinfo.length(v=vs.110).aspx)).

src/Cake.Testing/Extensions/FakeFileSystemExtensions.cs
```diff
@@ -42,6 +44,29 @@ public static FakeFile CreateFile(this FakeFileSystem fileSystem, FilePath path)
...
+        /// <param name="contentsBytes">The file contents.</param>
+        /// <returns>The same <see cref="FakeFile"/> instance so that multiple calls can be chained.</returns>
+        public static FakeFile CreateFile(this FakeFileSystem fileSystem, FilePath path, byte[] contentsBytes)
+        {
+            CreateDirectory(fileSystem, path.GetDirectory());
+
+            var file = fileSystem.GetFile(path);
+            if (!file.Exists)
+            {
+                using (Stream stream = file.OpenWrite(), ms = new MemoryStream(contentsBytes))
```
patriksvensson: Would prefer two distinct using statements here.

devlead:
@patriksvensson agreed with feedback and thing I've addressed it.

patriksvensson:
@devlead LGTM!
Could you rebase this against develop and I'll merge it.

devlead:
@patriksvensson rebased.

PR #797, opened 2016-03-27
actual type of RepositoryUrl is String not Uri

ghuntley:
In this case, I think the API doc is actually correct and the implementation of GitLink is at fault. This changes the API doc to match the actual API, feel free to close the PR and fix the API instead.


gep13:
@ghuntley good catch!  Initially, I had all the Url variables be as of type `Uri`, but then I changed my mind to make them all strings, as that is what would come from the command line anyway.  It is then up to GitLink to create this as a Uri if/when required.  Thanks for fixing this!

Issue #805, opened 2016-04-04
Exception running InspectCode and then directly after TeamCity.ImportData


laurentkempe:
Here is the exception 

> [17:00:11][Step 1/1] Importing data from 'artifacts/_ReSharperReports/inspectcode-report.xml' (45.71 KB) with 'ReSharperInspectCode' processor
> [17:00:11][Step 1/1] Caught exception while running service process importData
> [17:00:11]
> [Step 1/1] java.io.FileNotFoundException: C:\Temp\buildAgent1\work\ab9b99833ae569b1\artifacts\_ReSharperReports\inspectcode-report.xml (The process cannot access the file because it is being used by another process)
>     at java.io.FileInputStream.open0(Native Method)
>     at java.io.FileInputStream.open(FileInputStream.java:195)
>     at java.io.FileInputStream.<init>(FileInputStream.java:138)
>     at jetbrains.buildServer.dotnetTools.agent.inspectCode.InspectCodeFileProcessor.processReport(InspectCodeFileProcessor.java:54)
>     at jetbrains.buildServer.dotnetTools.agent.inspectCode.InspectCodeDataProcessor.processData(InspectCodeDataProcessor.java:38)
>     at jetbrains.buildServer.agent.impl.inspections.DataImportServiceMessage$1.run(DataImportServiceMessage.java:93)
>     at jetbrains.buildServer.agent.impl.serviceProcess.impl.ThreadedServiceProcessFactory$1.run(ThreadedServiceProcessFactory.java:31)
>     at java.lang.Thread.run(Thread.java:745)

This is due to [this line](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/InspectCode/InspectCodeRunner.cs#L81) which does `resultsFile.Open(FileMode.Open)`

Looks like there is the same problem here https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/DupFinder/DupFinderRunner.cs#L193 cc @gep13

devlead:
Yeah that stream should be wrapped ina using statement so it's disposed directly.

gep13:
Or, should the FileAccess be changed to FileAccess.Read, so it doesn't lock the file?

gep13:
i.e.

```
var xmlDoc = XDocument.Load(resultsFile.OpenRead());
```

gep13:
or it could even be `ReadLines()` which handles the disposing for us, assuming we know what the Encoding is.

devlead:
[...]

PR #807, opened 2016-04-04
(GH-805) Correcting issue with locked file

gep13:
- Rather than opening the File, thus locking it, use a Stream to dispose
of the lock on the file when working with it has completed.


gep13:
@devlead can you let me know what you think about this?

Relates to #805

devlead:
@gep13 LGTM :+1: I'll merge.

PR #831, opened 2016-04-18
Adds timeout to Tool/ToolSettings

### LINKED ISSUE #830: Add generic optional tool timeout
Currently some tools can hang with no chance of recovery, the ability to optionally set a timeout for would allow to take action in script and do retry or just abort on timeout.

devlead:
* Fixes #830

### DIFF
--- a/src/Cake.Core/Tooling/Tool.cs
+++ b/src/Cake.Core/Tooling/Tool.cs
@@ -82,7 +82,18 @@ namespace Cake.Core.Tooling
             var process = RunProcess(settings, arguments, processSettings);
 
             // Wait for the process to exit.
-            process.WaitForExit();
+            if (settings.ToolTimeout.HasValue)
+            {
+                if (!process.WaitForExit((int)settings.ToolTimeout.Value.TotalMilliseconds))
+                {
+                    const string message = "Tool timeout ({0}): {1}";
+                    throw new TimeoutException(string.Format(CultureInfo.InvariantCulture, message, settings.ToolTimeout.Value, GetToolName()));
+                }
+            }
+            else
+            {
+                process.WaitForExit();
+            }
 
             try
             {
--- a/src/Cake.Core/Tooling/ToolSettings.cs
+++ b/src/Cake.Core/Tooling/ToolSettings.cs
@@ -15,6 +15,11 @@ namespace Cake.Core.Tooling
         public FilePath ToolPath { get; set; }
 
         /// <summary>
+        /// Gets or sets optional timeout for tool execution.
+        /// </summary>
+        public TimeSpan? ToolTimeout { get; set; }
+
+        /// <summary>
         /// Gets or sets the argument customization.
         /// Argument customization is a way that lets you add, replace or reuse arguments passed to a tool.
         /// This allows you to support new tool arguments, customize arguments or address potential argument issues.


src/Cake.Core/Tooling/Tool.cs
```diff
@@ -82,7 +82,22 @@ protected void Run(TSettings settings, ProcessArgumentBuilder arguments)
...
             // Wait for the process to exit.
-            process.WaitForExit();
+            if (settings.ToolTimeout.HasValue)
+            {
+                if (!process.WaitForExit(settings.ToolTimeout.Value))
+                {
+                    throw new TimeoutException(
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            "Tool TimeOut ({0}): {1}",
```
patriksvensson: Should be spelled `timeout`.

src/Cake.Core/Tooling/Tool.cs
```diff
@@ -82,7 +82,22 @@ protected void Run(TSettings settings, ProcessArgumentBuilder arguments)
             var process = RunProcess(settings, arguments, processSettings);
 
             // Wait for the process to exit.
-            process.WaitForExit();
+            if (settings.ToolTimeout.HasValue)
+            {
+                if (!process.WaitForExit(settings.ToolTimeout.Value))
+                {
+                    throw new TimeoutException(
+                        string.Format(
```
patriksvensson: Can we separate formatting as we have done [here](https://github.com/devlead/cake/blob/feature/GH830/src/Cake.Core/Tooling/Tool.cs#L107)?

src/Cake.Core/Tooling/ToolSettings.cs
```diff
@@ -15,6 +15,11 @@ public class ToolSettings
         public FilePath ToolPath { get; set; }
 
         /// <summary>
+        /// Gets or sets optional timeout for tool execution
+        /// </summary>
+        public int? ToolTimeout { get; set; }
```
patriksvensson: What unit is this? Seconds? Milliseconds?

patriksvensson: Also, can we rename to only `Timeout`?

gep13: @patriksvensson said...
Also, can we rename to only Timeout?

Should that not be something like `TimeoutInSeconds` or `TimeoutInMilliseconds`?

gep13: Or were you thinking about putting the unit only in the comment?

patriksvensson: Would actually prefer this to be a `TimeSpan`.

src/Cake.Core/Tooling/ToolSettings.cs
```diff
@@ -15,6 +15,11 @@ public class ToolSettings
         public FilePath ToolPath { get; set; }
 
         /// <summary>
+        /// Gets or sets optional timeout for tool execution
```
patriksvensson: Missing period at the end of the sentence.

src/Cake.Core/Tooling/ToolSettings.cs
devlead: Hmm don't totally agree here. Happy do add InMilliseconds suffix.
Wouldn't just Timeout potentially collide with any setting that inherits? Think just Timeout is too generic and as it's two words I think we should case it ToolTimeoutInMilliseconds.

src/Cake.Core/Tooling/Tool.cs
devlead: Sure, I'll fix.

src/Cake.Core/Tooling/ToolSettings.cs
patriksvensson: I buy the prefix, but why not TimeSpan? I think we should go with a TimeSpan here. The script code becomes way more cleaner when you say `tool.Timeout = TimeSpan.FromSeconds(15)` instead of `tool.TimeoutInMilliseconds = 15000`.

devlead: sure TimeSpan work.

patriksvensson: I'm OK with TimeSpanInMilliseconds as well. Not going to push this if you think it's not a good idea :smile:

devlead:
@patriksvensson @gep13 
* Changed to TimeSpan
* Moved exception string format string to it's own line
* Added missing period in docs

patriksvensson:
@devlead Can you rebase this?

devlead:
@patriksvensson rebased.

patriksvensson:
Merged! :+1:

Issue #833, opened 2016-04-19
Add configuration file for Cake


patriksvensson:
Sometimes we want to be able to configure internals of Cake so I propose that we add support for this. My proposal is that we add `ICakeConfiguration` that will fetch values in the following order:

1) Environment
2) Configuration file
3) Arguments

```csharp
public interface ICakeConfiguration
{
  string GetValue(string key);
}
```

* Configuration keys are case insensitive.

#### Environment variables:

Environment variables are named like the following convention:

```
CAKE_[KEY]
```

The `CAKE_` part will be completely removed from the key name.

##### Example

```powershell
PS> $env:CAKE_RUNTESTS=true
PS> $env:CAKE_TOOLS_PATH=D:/tools
```

#### Configuration file:

```ini
; Cake configuration file

; Global settings
RunTests=true

[Tools]
Path=D:/tools
```

#### Arguments:

Arguments are passed in the following format:

```
--[key]=[value]
```

##### Example:

```
Cake.exe --tools_path=D:/tools --runtests=true
```

### Using the configuration

```csharp
public void DoMagic(ICakeConfiguration config)
{
  var runTests = config.Get("RunTests");
  var toolPath2 = config.Get("Tools_Path");
}
```

RLittlesII:
@patriksvensson would there be a way for me to explicitly choose one of the configuration mechanisms? I can see myself attempting to send in configuration from a file unaware that someone has setup environment variables.  Maybe on the `-Verbosity diagnostic` it can tell me which configuration is being respected.

patriksvensson:
@RLittlesII Different ways of getting configuration values will have different precedence.

1. Arguments
2. Configuration file
3. Environment

So an environment variable can be overridden by a configuration file which in turn can be overridden by an argument.

gitfool:
Nice. FWIW, sounds a lot like [aspnet/Configuration](https://github.com/aspnet/Configuration).

gep13:
> @gitfool said...
Nice. FWIW, sounds a lot like aspnet/Configuration.

Thanks, we are aware of the potential cross over.  However, since that package is still pre-release, and also due to the fact that we don't want to pull in another dependency for Cake at this time, we decided to roll our own for now.

bspinner:
Shouldn't the config-file also be respected by the `build.ps1` and `build.sh` scripts?

patriksvensson:
[...]

PR #840, opened 2016-04-20
Add configuration file support

patriksvensson:
*(Description copied from issue)*

> Sometimes we want to be able to configure internals of Cake so I propose that we add support for this. My proposal is that we add `ICakeConfiguration` that will fetch values in the following order:
> 
> 1) Environment
> 2) Configuration file
> 3) Arguments
> 
> ```csharp
> public interface ICakeConfiguration
> {
>   string GetValue(string key);
> }
> ```
> 
> * Configuration keys are case insensitive.
> 
> #### Environment variables:
> 
> Environment variables are named like the following convention:
> 
> ```
> CAKE_[KEY]
> ```
> 
> The `CAKE_` part will be completely removed from the key name.
> 
> ##### Example
> 
> ```powershell
> PS> $env:CAKE_RUNTESTS=true
> PS> $env:CAKE_TOOLS_PATH=D:/tools
> ```
> 
> #### Configuration file:
> 
> ```ini
> ; Cake configuration file
> 
> ; Global settings
> RunTests=true
> 
> [Tools]
> Path=D:/tools
> ```
> 
> #### Arguments:
> 
> Arguments are passed in the following format:
> 
> ```
> --[key]=[value]
> ```
> 
> ##### Example:
> 
> ```
> Cake.exe --tools_path=D:/tools --runtests=true
> ```
> 
> ### Using the configuration
> 
> ```csharp
> public void DoMagic(ICakeConfiguration config)
> {
>   var runTests = config.GetValue("RunTests");
>   var toolPath = config.GetValue("Tools_Path");
> }
> ```

Related to #833.


src/Cake/Arguments/ArgumentParser.cs
```diff
@@ -67,10 +66,11 @@ public CakeOptions Parse(IEnumerable<string> args)
                             // Make sure we parse the option
                             if (!ParseOption(value, options))
                             {
-                                return null;
+                                options.HasError = true;
+                                return options;
                             }
 
-                            options.Script = GetDefaultScript();
+                            options.Script = "./build.cake";
```
patriksvensson: @devlead @gep13 

Removed conventions we defined here since we only have one real convention for file names now. Also removes the responsibility of the tokenizer to determine whether or not files exist (which shouldn't be it's job).

Ok with you?

build.cake
```diff
@@ -56,6 +56,7 @@ Task("Restore-NuGet-Packages")
     .Does(() =>
 {
     NuGetRestore("./src/Cake.sln", new NuGetRestoreSettings {
+        ArgumentCustomization = arg => arg.Append("-MSBuildVersion 14"),
```
devlead: msbuild version '14' won't work on non Windows platforms.

src/Cake/Arguments/ArgumentParser.cs
devlead: @patriksvensson Yes, makes more sense to me.

src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
```diff
@@ -0,0 +1,77 @@
...
+            foreach (var token in tokens)
+            {
+                // Section?
+                if (token.IsSection)
+                {
+                    currentSection = token.Value;
+                    continue;
+                }
+
+                // Parse the key/value pair
```
patriksvensson: Add period.

RichiCoder1:
It's still technically pre-release, but I'm curious why not something like https://github.com/aspnet/Configuration?

patriksvensson:
@RichiCoder1: Prefer not to introduce dependencies unless necessary - and this was kind of trivial. Also didn't know about it and not sure that it supports the precedence rules we require.

RichiCoder1:
It came to mind precisely because it supports the kind of presendence rules you're talking about :).
This isn't the best article, but it covers all the details: https://msdn.microsoft.com/en-us/magazine/mt632279.aspx

I'm all for minimal dependencies, but configuration is one of those things I'd err towards a common and battle-tested solution, especially one you're doing more than just ini/xml configs.

patriksvensson:
This is just for internal configuration of Cake that won't be used much so I think it will be fine. If we want to change the implementation details later we can always do that.

src/Cake.Core/Cake.Core.csproj
```diff
@@ -210,8 +219,7 @@
   </ItemGroup>
   <ItemGroup />
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Import Project="..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets"
-          Condition="Exists('..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets') and $(Configuration)=='Release' and ($(Windows)=='True' or $(BuildingInsideVisualStudio)=='True')" />
+  <Import Project="..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets" Condition="Exists('..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets') and $(Configuration)=='Release' and ($(Windows)=='True' or $(BuildingInsideVisualStudio)=='True')" />
```
devlead: This change needed?

patriksvensson: Visual Studio does that when saving the project for some reason.

src/Cake.Core/ICakeArguments.cs
```diff
@@ -1,19 +1,11 @@
...
     /// <summary>
     /// Represents arguments passed to script.
     /// </summary>
     public interface ICakeArguments
     {
         /// <summary>
-        /// Initializes the argument list.
-        /// </summary>
-        /// <param name="arguments">The arguments.</param>
-        void SetArguments(IDictionary<string, string> arguments);
```
devlead: Just note to self, potential breaking change, don't think anyone has used this externally though.

patriksvensson: Yes, I agree. See no reason for people to use this one as well :smile:

src/Cake.Core/Scripting/ScriptRunner.cs
```diff
@@ -99,13 +99,11 @@ public void Run(IScriptHost host, FilePath scriptPath, IDictionary<string, strin
...
 
-            // Prepare the context.
-            host.Context.Arguments.SetArguments(arguments);
-            host.Context.Environment.WorkingDirectory = scriptPath.MakeAbsolute(host.Context.Environment).GetDirectory();
+            // Prepare the environment.
+            _environment.WorkingDirectory = scriptPath.MakeAbsolute(_environment).GetDirectory();
 
             // Analyze the script file.
             _log.Verbose("Analyzing build script...");
-            scriptPath = scriptPath.IsRelative ? scriptPath.MakeAbsolute(_environment) : scriptPath;
```
devlead: Note to self, correct, this is already done in:
https://github.com/cake-build/cake/blob/c3f1280b6edc6a4983ccea655cca3e5a8efab805/src/Cake.Core/Scripting/Analysis/ScriptAnalyzer.cs#L75

src/Cake/Program.cs
```diff
@@ -36,90 +23,27 @@ public static int Main()
...
-            else
-            {
-                // Roslyn related services.
-                builder.RegisterType<RoslynScriptEngine>().As<IScriptEngine>().SingleInstance();
-                builder.RegisterType<RoslynScriptSessionFactory>().SingleInstance();
-                builder.RegisterType<RoslynNightlyScriptSessionFactory>().SingleInstance();
-            }
+                // Parse the options.
+                var parser = container.Resolve<IArgumentParser>();
+                var options = parser.Parse(args) ?? new CakeOptions();
```
devlead: This should be within a try catch, i.e. an invalid verbosity will now generate an unhandled exception
```
Unhandled Exception: System.InvalidOperationException: The value 'Invalid' is not a valid verbosity.
   at Cake.Arguments.VerbosityParser.Parse(String value) in \src\Cake\Arguments\VerbosityParser.cs:line 49
   at Cake.Arguments.ArgumentParser.ParseOption(String name, String value, CakeOptions options) in \src\Cake\Arguments\ArgumentParser.cs:line 1
30
   at Cake.Arguments.ArgumentParser.Parse(IEnumerable`1 args) in \src\Cake\Arguments\ArgumentParser.cs:line 46
   at Cake.Program.Main() in \src\Cake\Program.cs:line 34
```

Where as it before would just generate an error:
```
Error: The value 'Invalid' is not a valid verbosity.
```

devlead:
@patriksvensson got a style cop error
https://ci.appveyor.com/project/cakebuild/cake/build/0.11.0-PullRequest.840+101.build.1886#L866
```
C:\projects\cake\src\Cake\Arguments\VerbosityParser.cs(19,1): error : SA1611 : CSharp.Documentation : The documentation header must contain param tags matching the element's parameter list. [C:\projects\cake\src\Cake\Cake.csproj]
  Cake.NuGet.Tests -> C:\projects\cake\src\Cake.NuGet.Tests\bin\Release\Cake.NuGet.Tests.dll
C:\projects\cake\tools\Cake\Cake.exe : An error occured when executing task 'Build'.
At line:1 char:1
```

patriksvensson:
@devlead Ok, will take a quick look.

devlead:
@patriksvensson LGTM, I'll take it for a final spin and merge if no issues arise.

Issue #860, opened 2016-05-02
During the publishing cycle, continue with each step, and error at end if there is a problem


gep13:
We want to attempt the publishing of each release of Cake.  If there is an issue publishing to one or more of the destinations, then we want to fail the overall build, however, we want to at least attempt all of them.

That way, manual publishing can happen for any that failed.

patriksvensson:
Merged via #863

PR #863, opened 2016-05-03
(GH-860) Attempt all publishing tasks

gep13:
Relates to #860

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -11,6 +11,7 @@
 //////////////////////////////////////////////////////////////////////
 
 BuildParameters parameters = BuildParameters.GetParameters(Context, BuildSystem);
+bool publishingError = false;
 
 ///////////////////////////////////////////////////////////////////////////////
 // SETUP / TEARDOWN
@@ -203,6 +204,11 @@ Task("Publish-MyGet")
             ApiKey = apiKey
         });
     }
+})
+.OnError(exception =>
+{
+    Information("Publish-MyGet Task failed, but continuing with next Task...");
+    publishingError = true;
 });
 
 Task("Publish-NuGet")
@@ -234,6 +240,11 @@ Task("Publish-NuGet")
           Source = apiUrl
         });
     }
+})
+.OnError(exception =>
+{
+    Information("Publish-NuGet Task failed, but continuing with next Task...");
+    publishingError = true;
 });
 
 Task("Publish-Chocolatey")
@@ -265,6 +276,11 @@ Task("Publish-Chocolatey")
           Source = apiUrl
         });
     }
+})
+.OnError(exception =>
+{
+    Information("Publish-Chocolatey Task failed, but continuing with next Task...");
+    publishingError = true;
 });
 
 Task("Publish-HomeBrew")
@@ -279,6 +295,11 @@ Task("Publish-HomeBrew")
     var hash = CalculateFileHash(parameters.Paths.Files.ZipArtifactPath).ToHex();
 
     Information("Hash for creating HomeBrew PullRequest: {0}", hash);
+})
+.OnError(exception =>
+{
+    Information("Publish-HomeBrew Task failed, but continuing with next Task...");
+    publishingError = true;
 });
 
 Task("Publish-GitHub-Release")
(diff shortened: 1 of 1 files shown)


gep13:
@patriksvensson @devlead just putting this out there for discussion as a possible workaround to the problem that we show during the 0.11.0 release.  Thoughts?

build.cake
```diff
@@ -207,6 +213,7 @@ Task("Publish-NuGet")
     .WithCriteria(() => parameters.IsMainCakeRepo)
     .WithCriteria(() => parameters.IsMainCakeBranch)
     .WithCriteria(() => parameters.IsTagged)
+    .ContinueOnError()
```
patriksvensson: I think that `ContinueOnError` will suppress the `OnError` handling but I'm not sure.

build.cake
```diff
@@ -317,7 +343,14 @@ Task("AppVeyor")
   .IsDependentOn("Publish-NuGet")
   .IsDependentOn("Publish-Chocolatey")
   .IsDependentOn("Publish-HomeBrew")
-  .IsDependentOn("Publish-GitHub-Release");
+  .IsDependentOn("Publish-GitHub-Release")
+  .Finally(() =>
```
patriksvensson: This is a really smart use of `Finally` :smile: :+1:

build.cake
gep13: @patriksvensson yeah, I wasn't sure what would happen there.  Will need to test it out to see what happens.  If it does, any suggestions on how best to handle this?

patriksvensson: @gep13 I think the easiest way to test this is a new Cake script with both `ContinueOnError` and `OnError` where you throw an exception in the task.

gep13: @patriksvensson sounds like a plan, will give it a try, and report back.

gep13:
@patriksvensson If I attempt the above in a stripped down version of the build script, I get the following:

image

Going to switch it out to use `ReportError` instead...

gep13:
@patriksvensson @devlead ok, I "think" I am onto a winner :smile: 

What do you think about the following:

https://gist.github.com/gep13/028b4d3f3a98c3cbdfb122da9a35fdf8

patriksvensson:
@gep13 ReportError won't work for this. Just use OnError and remove ContinueOnError. Read more here: http://cakebuild.net/docs/fundamentals/error-handling

gep13:
>@patriksvensson said...
@gep13 ReportError won't work for this. Just use OnError and remove ContinueOnError. Read more here: http://cakebuild.net/docs/fundamentals/error-handling

Based on the test that I just did, it seems to work?!?

gep13:
@patriksvensson if I remove ContinueOnError, how can I allow the build to continue onto the next task in the sequence?

patriksvensson:
ReportError should do a`throw` and is meant to be a way to log errors but still keep the original call stack. If it doesn't work likw this then somwthing must be wrong :(

gep13:
So my understanding of this is as follows...

I have a set of tasks which I want to run.  None of these tasks are dependent on each other, but I want to run them until, one after the other, and I want to control their execution order from a top level task.  To allow all the tasks to execute, regardless of whether each one succeeds or not, I use the `ContinueOnError` extension method for each task.  As a result, they all execute.

However, if one of these tasks does cause an error, I want to be able to Report, or rather, change a status flag that something has occurred.  I would have thought in this case that the ReportError extension method is doing exactly what it is supposed to.  Allowing an interception of the error that has occurred, without impacting on the overall flow.

What would you have expected to happen in the above scenario?

patriksvensson:
@gep13 Yes. ReportError should have changed the status flag and then (internally) rethown the original exception. The OnError method (should be renamed to `Catch`) should do what you want.

gep13:
Ah, ok, I see what you are saying...

So using this:

https://gist.github.com/gep13/cdf50d42aa7d4500bd43555013c72087

Also achieves the same thing.  Since we are not manually re-throwing the exception within OnError, it is as though the error never happened, and execution continues to the next task.

So, we might have an issue with ReportError, since it should have re-thrown.  However, I am wondering whether using ReportError in conjunction with ContinueOnError, actually results in the same functionality as OnError with no re-throw, i.e. silent continuation of workflow?

patriksvensson:
@gep13 Ah, that's a side effect I haven't thought about! If this is the desired effect we should probably document it (both in documentation and with a unit and integration test).

gep13:
>@patriksvensson said...
If this is the desired effect we should probably document it

Agreed.  If this shouldn't be allowed, then we should report an error, similar to when you use OnError and ContinueOnError on the same task.

gep13:
@patriksvensson @devlead ok, I think this is ready for a review.

patriksvensson:
Merged! :+1:

PR #868, opened 2016-05-05
Fix for tool installation path bug.

### LINKED ISSUE #867: Tool path is wrong when calling cake file in sub directory
### What You Are Seeing?

When passing a relative path to a script file not relative to the bootstrapper location, the paths to tools get messed up when using the `#tool` directive.

### What is Expected?

The tool path to be relative to the executing script.

0.11.0

64-bit

Windows

No

1) Create the following folder structure:

```
build.ps1
SubFolder
  - build.cake
```

2) Where `build.cake` looks like this:

```csharp
#tool "nuget:?package=XamarinComponent"
```

3) Then you should see that the tool will be downloaded to the wrong location when running the script with a relative path:

```
./build.ps1 -Script "./SubFolder/build.cake"
```

patriksvensson:
When passing a relative path to a script not in the same
directory as the bootstrapper, tools would be installed in
the wrong directory.

Closes #867


gep13:
**NOTE:** I am putting the failing Travis Build down to just being Travis :cry:

Issue #921, opened 2016-05-25
Do not publish Cake.Nuget on nuget.org


gep13:
This is an internal dependency for Cake, and should not be pushed to NuGet.org.

PR #924, opened 2016-05-26
Removed internal NuGet Package from Build Process

gep13:
Relates to #921 

@devlead does this make sense to you?


patriksvensson:
I think this makes sense. We could also remove the nuspec.

gep13:
@patriksvensson yes, that occurred to me as well.  I will get that removed too :+1:

devlead:
Agree with @patriksvensson othwise LGTM.

gep13:
@devlead @patriksvensson ok, I have made the suggested change :-D

devlead:
@gep13 LGTM :+1: I'll merge once CI:s done.

PR #926, opened 2016-05-26
Built in parameters now accept boolean values

### LINKED ISSUE #917: Allow setting boolean values for built in Cake arguments
Right now, if you want to dry run a Cake script you will have to write `--dryrun`. We should however support `--dryrun=true` and `--dryrun=false`. Not providing a value should default to `true`.

* `--dryrun` / `--noop` / `--whatif`
* `--mono`
* `--experimental`
* `--showdescription`
* `--debug`

A good place to start looking would be in [ArgumentParser.cs](https://github.com/cake-build/cake/blob/develop/src/Cake/Arguments/ArgumentParser.cs)

patriksvensson:
This PR enabled the built in parameters to Cake.exe to take the form as a boolean key/value pair instead of only as a flag.

This means that `--debug=true` will be the same thing as `--debug` and will give the user an opportunity to explicitly tell Cake to NOT debug the script by passing `--debug=false`. This is useful when writing a custom bootstrapper for Cake.

Closes #917

### DIFF
--- a/src/Cake.Tests/Unit/Arguments/ArgumentParserTests.cs
+++ b/src/Cake.Tests/Unit/Arguments/ArgumentParserTests.cs
@@ -218,9 +218,13 @@ namespace Cake.Tests.Unit.Arguments
                 }
 
                 [Theory]
-                [InlineData("-showdescription")]
-                [InlineData("-s")]
-                public void Can_Parse_ShowDescription(string input)
+                [InlineData("-showdescription", true)]
+                [InlineData("-showdescription=true", true)]
+                [InlineData("-showdescription=false", false)]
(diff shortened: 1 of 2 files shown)


gep13:
LGTM!

@patriksvensson is there any part of the documentation that should be updated to reflect that this is now possible?

patriksvensson:
@gep13 I'm not sure. Maybe we could be OK with this being a "hidden" feature since it mostly solves some not so common scenarios.

gep13:
@patriksvensson okay, that makes sense as well.  Was just if there was already some documentation around this, that it should be updated to match new functionality.

PR #838, opened 2016-04-20
DNU doesn't like semi-colons anymore.  It uses multiple args for the same flag.

adamhathcock:
Same as https://github.com/cake-build/cake/pull/834

Rebased on develop for release with 0.11.0

Thanks!

PR #935, opened 2016-05-31
GH945: Fix dotnet restore settings

adamhathcock:
source, fallbacksource and inferedruntimes all allow multiple arguments by specifying the same arg multiple times.

Same fix as this: https://github.com/cake-build/cake/pull/838

### DIFF
--- a/src/Cake.Common/Tools/DotNetCore/Build/DotNetCoreBuildSettings.cs
+++ b/src/Cake.Common/Tools/DotNetCore/Build/DotNetCoreBuildSettings.cs
@@ -29,9 +29,9 @@ namespace Cake.Common.Tools.DotNetCore.Build
         public string Configuration { get; set; }
 
         /// <summary>
-        /// Gets or sets specific frameworks to compile.
+        /// Gets or sets the specific framework to compile.
         /// </summary>
-        public ICollection<string> Frameworks { get; set; }
+        public string Framework { get; set; }
 
         /// <summary>
         /// Gets or sets the value that defines what `*` should be replaced with in version field in project.json.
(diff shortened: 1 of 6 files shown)


adamhathcock:
Seems like there is a lot wrong?

adamhathcock:
I'm thinking the usage of the dotnet cli isn't correct and some of this needs to be cleaned up.  That would involve breaking changes though.  I can limit what i'm doing to just additive changes but it's less pretty.

src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestoreSettings.cs
```diff
@@ -14,6 +14,11 @@ public sealed class DotNetCoreRestoreSettings : DotNetCoreSettings
         public string Source { get; set; }
```
devlead: @adamhathcock please in this PR remove `string Source { get; set; }` and it's usages in favor of the new `ICollection<string> Sources { get; set; }`

We've discussed internally with in @cake-build/team even though it's a breaking change, to keep API clean and consistent, and as we feel we catch this early.

src/Cake.Common/Tools/DotNetCore/Build/DotNetCoreBuildSettings.cs
```diff
@@ -29,6 +29,11 @@ public sealed class DotNetCoreBuildSettings : DotNetCoreSettings
         public string Configuration { get; set; }
```
devlead: @adamhathcock please in this PR remove `string Configuration { get; set; }` and it's usages in favor of the new `ICollection<string> Configurations { get; set; }`

We've discussed internally with in @cake-build/team even though it's a breaking change, to keep API clean and consistent, and as we feel we catch this early.

devlead:
@adamhathcock left some feedback, could you please address, rebase against latest develop, squash into 1 commit and when ping us when you PR is ready for review and we'll look at reviewing asap.

adamhathcock:
After looking at the dotnet cli repo code and some manual testing.  I think I've fixed the issues now. 
Dotnet Restore can have more than one source.  Fallbacksources need to be split out.

When building, you can only have one configuration and one framework at a time.

devlead:
@adamhathcock ah nice catch! Then we'll review this and ping you if we find anything. Meanwhile could you create an issue for this and reference it in this PR.

gep13:
Relates to #945

devlead:
@adamhathcock LGTM :+1: Merged, thanks for contributing!

PR #946, opened 2016-06-02
GH960: Add configuration for default NuGet source

wwwlicious:
Not too familiar with the code base to know where best to add the default or fallback check for a nuget source but the following seems to work for my purposes.

### DIFF
--- a/src/Cake.NuGet.Tests/Fixtures/NuGetPackageInstallerFixture.cs
+++ b/src/Cake.NuGet.Tests/Fixtures/NuGetPackageInstallerFixture.cs
@@ -25,6 +25,8 @@ namespace Cake.NuGet.Tests.Fixtures
         public PackageType PackageType { get; set; }
         public DirectoryPath InstallPath { get; set; }
 
+        public FakeConfiguration Config { get; set; }
+
         public NuGetPackageInstallerFixture()
         {
             Environment = FakeEnvironment.CreateUnixEnvironment();
@@ -32,6 +34,7 @@ namespace Cake.NuGet.Tests.Fixtures
             ProcessRunner = Substitute.For<IProcessRunner>();
             ContentResolver = Substitute.For<INuGetPackageContentResolver>();
             Log = Substitute.For<ICakeLog>();
+            Config = new FakeConfiguration();
 
             ToolResolver = Substitute.For<INuGetToolResolver>();
             ToolResolver.ResolvePath().Returns(new FilePath("/Working/tools/nuget.exe"));
@@ -40,10 +43,11 @@ namespace Cake.NuGet.Tests.Fixtures
             PackageType = PackageType.Addin;
             InstallPath = new DirectoryPath("./nuget");
         }
+        
 
         public NuGetPackageInstaller CreateInstaller()
         {
-            return new NuGetPackageInstaller(FileSystem, Environment, ProcessRunner, ToolResolver , ContentResolver, Log);
+            return new NuGetPackageInstaller(FileSystem, Environment, ProcessRunner, ToolResolver , ContentResolver, Log, Config);
         }
 
         public IReadOnlyCollection<IFile> Install()
--- a/src/Cake.NuGet.Tests/Fixtures/NuGetPackageInstallerFixture.cs
+++ b/src/Cake.NuGet.Tests/Fixtures/NuGetPackageInstallerFixture.cs
@@ -12,6 +12,8 @@ using NSubstitute;
 
 namespace Cake.NuGet.Tests.Fixtures
 {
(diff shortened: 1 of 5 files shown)


wwwlicious:
the appveyor build for this PR failed to run yesterday rather than failed when running, can it be retried.

patriksvensson:
@wwwlicious I don't think that the nuget package installer should use the Roslyn nuget source. I think that it should use a NUGET_SOURCE (that should be defined in Constants.cs) that could consist of a semicolon separated list of nuget sources to use.

wwwlicious:
ok, didn't notice that one, will see if I can refactor. Any pref for `Cake.Core.Constants` or `Cake.Constants`? The roslyn one is in Cake.Constants so I guess that one...

patriksvensson:
@wwwlicious Put it in it's own Constant.cs in Cake.NuGet project for the time being.

wwwlicious:
will do, any just to clarify, the purpose of the value is to be able to override the default nuget source via config, should a constant relate to a config key, a default nuget sources list or should I add two, a default source and a config key source and prefer the config over the default?

patriksvensson:
@wwwlicious It's just for the config key.

```csharp
namespace Cake.NuGet
{
    internal static class Constants
    {
        public static class NuGet
        {
            public const string Source = "NuGet_Source";
        }
    }
}
```

@devlead @gep13 You have any thoughts about this?

wwwlicious:
something like the following:

```csharp
// if an absolute uri is specified for source, use this
// otherwise check cake.config for customise package source/s
if (definition.Address != null)
{
    arguments.Append("-Source");
    arguments.AppendQuoted(definition.Address.AbsoluteUri);
}
else
{
    var nugetSource = config.GetValue(Constants.NuGet.Source);
    if (!string.IsNullOrWhiteSpace(nugetSource))
    {
        arguments.Append("-Source");
        arguments.AppendQuoted(nugetSource);
    }
}
```

patriksvensson:
@wwwlicious Yes, that looks good to me. We would also need test coverage for this.

wwwlicious:
ok, will do.

patriksvensson:
@wwwlicious Could you also create an issue so we can track this. Thanks!

wwwlicious:
It was actually referenced from an issue I raised against the resources repo as I've had to modify the build.ps1 script also but on further investigation I'm having to modify the nuget installer. 
https://github.com/cake-build/resources/issues/6

Happy to raise another issue in this repo for it but thought it is worth asking if the previous link is suitable first?

patriksvensson:
@wwwlicious We need an issue in this repository to track the changes in the milestone which to generate the release notes with GitReleaseManager. Otherwise the issue in the other repo would have been fine :smile:

wwwlicious:
No problem, will raise one now.

Am still figuring out how to best write tests for this following existing styles, will hopefully update PR tomorrow.

patriksvensson:
@wwwlicious No rush :smile: Feel free to ping me on Gitter och write something here if you have any questions or problems. Happy to help if I can.

wwwlicious:
I have added tests to check the config key is or isn't called depending on the uri type (absolute or releative) 
Can someone review and advise if this is ok to merge or I need to do more?

src/Cake.Common/Solution/Project/ProjectParser.cs
```diff
@@ -76,14 +76,18 @@ public ProjectParserResult Parse(FilePath projectPath)
                      .Elements(ProjectXElement.Configuration)
                      .Select(cfg => cfg.Value)
                      .FirstOrDefault()
+                 let platform = propertyGroup
+                     .Elements(ProjectXElement.Platform)
+                     .Select(cfg => cfg.Value)
+                     .FirstOrDefault()
+                let configPropertyGroups = project.Elements(ProjectXElement.PropertyGroup)
```
patriksvensson: I think this indentation is wrong.

patriksvensson:
@wwwlicious It looks like this PR contains more functionality than it should.

I'm thinking about the "	Adding outputPath to project properties based on configuration and platform"-commit (https://github.com/cake-build/cake/pull/946/commits/f3a201104e1d1368c57d64b9218d317ad3f4e162) that somehow managed to sneak into this PR.

wwwlicious:
Should be corrected now.

gep13:
@wwwlicious thank you for your contribution!

I have rebased, corrected, and merged this here: https://github.com/cake-build/cake/commit/e61ec08215107576c2dea8cb080db9d61f1f1780

As a result, I am going to close this PR, however, your work has been correctly attributed.  Thanks again!

gep13:
Relates to #960

PR #962, opened 2016-06-06
Added AddiationalParameters to DotNetCoreTestSettings

phenixdotnet:
Please find a small PR which allow custom parameters to be sent to the DotNetCore test command.
Its allow the scenario where you need to pass specifics arguments for the test runner like the ```-xml resultsfile.xml``` for xunit.


phenixdotnet:
Use ArgumentCustomization

Issue #902, opened 2016-05-19
Add IsDebug to ICakeEnvironment


mholo65:
Add runtime info to ICakeEnvironment to know if performing debug.

mholo65:
Discussed here #869 with @patriksvensson

mholo65:
Needed by #876 if we would like to use same directive for launch and break..

gep13:
Related to https://github.com/cake-build/cake/issues/502

gep13:
@mholo65 given the work that @patriksvensson has been doing on getting the CoreClr port of Cake to where it is, I am going to defer this work until after that is released.  Hope that makes sense.

mholo65:
@gep13 np, looking forward to CoreCLR support

mholo65:
Not needed

PR #965, opened 2016-06-06
[HOLD] GH876: Possibility to launch debugger

mholo65:
Still needs some work in Debug command. Currently it waits forever for the debugger to be attached, maybe the timespan [here](https://github.com/cake-build/cake/blob/f75166ccac5419b58a8347fe7afcbd469a7bbdde/src/Cake/Commands/DebugCommand.cs#L44) should be configurable via commandline?

### DIFF
--- a/src/Cake.Core/CakeEnvironment.cs
+++ b/src/Cake.Core/CakeEnvironment.cs
@@ -56,6 +56,27 @@ namespace Cake.Core
             return Machine.IsUnix();
         }
 
+        private static readonly string[] _debugArgs =
+        {
+            "-d", "--d", "-d=true", "--d=true",
+            "-debug", "--debug", "-debug=true", "--debug=true"
+        };
+
+        /// <summary>
+        /// Gets a value indicating whether the current execution is running in debug.
+        /// </summary>
+        /// <returns>
+        /// Whether or not the current execution is running in debug.
+        /// </returns>
+        public bool IsDebug
+        {
+            get
+            {
+                var args = Environment.GetCommandLineArgs();
+                return args.Intersect(_debugArgs).Any();
+            }
+        }
+
         /// <summary>
         /// Gets a special path.
         /// </summary>
--- a/src/Cake/Commands/DebugCommand.cs
+++ b/src/Cake/Commands/DebugCommand.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 using System;
 using System.Threading;
+using Cake.Core;
 using Cake.Core.Diagnostics;
 using Cake.Core.Scripting;
 using Cake.Diagnostics;
@@ -18,16 +19,18 @@ namespace Cake.Commands
         private readonly IScriptRunner _scriptRunner;
         private readonly IDebugger _debugger;
         private readonly ICakeLog _log;
+        private readonly IConsole _console;
         private readonly BuildScriptHost _host;
 
         // Delegate factory used by Autofac.
         public delegate DebugCommand Factory();
 
-        public DebugCommand(IScriptRunner scriptRunner, IDebugger debugger, BuildScriptHost host)
+        public DebugCommand(IScriptRunner scriptRunner, IDebugger debugger, IConsole console, BuildScriptHost host)
         {
             _scriptRunner = scriptRunner;
             _debugger = debugger;
             _host = host;
+            _console = console;
             _log = _host.Context.Log;
         }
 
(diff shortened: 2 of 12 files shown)


src/Cake.Core/CakeEnvironment.cs
```diff
@@ -53,6 +53,27 @@ public bool IsUnix()
...
+        /// Whether or not the current execution is running in debug.
+        /// </returns>
+        public bool IsDebug
+        {
+            get
+            {
+                var args = Environment.GetCommandLineArgs();
+                return args.Intersect(_debugArgs).Any();
+            }
+        }
```
mholo65: @patriksvensson is this good enough solution for #902?

src/Cake/Commands/DebugCommand.cs
```diff
@@ -38,15 +38,25 @@ public bool Execute(CakeOptions options)
...
 
             _log.Debug("Performing debug...");
             _log.Information(Verbosity.Quiet, message, pid);
 
-            _debugger.WaitForAttach(Timeout.InfiniteTimeSpan);
-
-            _log.Debug("Debugger attached");
+            while (true)
+            {
+                if (Console.KeyAvailable && Console.ReadKey(true).Key == ConsoleKey.Escape)
```
mholo65: @patriksvensson Should this maybe be abstracted? I mean add **KeyAvailable** + **ReadKey** to **IConsole**, then it would be possible to write unit test around this also...

mholo65: Hmm, at least it can't be tested on [TeamCity](http://teamcity.codebetter.com/viewLog.html?buildId=219414&buildTypeId=Cake_CakeMaster) right now...

mholo65:
@patriksvensson @devlead, when you have time, can I have some eyes on this? At least reading from console needs to be abstracted away to make **DebugCommand** testable on TeamCity...

mholo65:
I've removed [WIP] now and fixed reading key from console in tests..

Have a look when you have time :smile:

//cc @gep13 @patriksvensson @devlead

gep13:
@patriksvensson are you in a position to take a look at this, with regard to the changes that you are making in the PR for ICakeEnvironment.  Just want to make sure that they are going to work together.

patriksvensson:
@gep13 Good idea. The `IsDebug` property would make sense to add to the new `ICakeRuntime`.

src/Cake/Commands/DebugCommand.cs
```diff
@@ -38,15 +41,25 @@ public bool Execute(CakeOptions options)
...
 
             _log.Debug("Performing debug...");
             _log.Information(Verbosity.Quiet, message, pid);
 
-            _debugger.WaitForAttach(Timeout.InfiniteTimeSpan);
-
-            _log.Debug("Debugger attached");
+            while (true)
+            {
+                if (_console.KeyAvailable && _console.ReadKey(true).Key == ConsoleKey.Escape)
```
patriksvensson: I don't think aborting from this is necessary. If you wish to not attach the debugger, a Ctrl+C and relaunch without the `--debug` flag would suffice.

src/Cake.Core/CakeEnvironment.cs
patriksvensson: @mholo65 We've refactored ICakeEnviornment to include a `ICakeRuntime` component and I think this property would be right at home there. PR is not merged yet though, but hopefully soon. You could also inject `ICakeArguments` to the `ICakeRuntime` to get the command line arguments.

src/Cake/Commands/DebugCommand.cs
mholo65: @patriksvensson the ```--debug``` flag is required for debugging (i.e. instructs composition to hook up RoslynDebugScriptSession which generates DLL + PDB and loads them into memory). In case we don't like to merge the different Roslyn script sessions, aborting is needed.

patriksvensson: @mholo65 I'm not sure that I understand what you mean... :confused:

patriksvensson: @mholo65 Oh wait, I understand now.

patriksvensson: @mholo65 However, couldn't this be an argument to Cake instead? Like `--launchdebugger` or similar? Would prefer not adding support for user interaction in the core library.

mholo65: @patriksvensson yes, that's an option. The next question would then be: Should debugger break on first #break directive or on first line of cake script?

src/Cake.Core/CakeEnvironment.cs
mholo65: @patriksvensson yes, I could add this as an separate PR as it solves another issue..

gep13:
@mholo65 was away to look at this again, but noticed that there is now a conflict with the latest changes.  Would you be in a position to update this?  Thanks!

mholo65:
@gep13 sure thing, currently working on #902 (which is kind of a pre-req for this). But with current modules, it's not that straightforward to inject ICakeArguments to CakeRuntime as arguments are resolved much later than core... I'll open a PR where we can discuss that.

...and this PR probably needs some changes as @patriksvensson didn't want user interaction in the core..

gep13:
@mholo65 ok, perfect, thanks for getting back to me.  I will mark this on hold for now, and we can update it when ready.  Thanks!

mholo65:
No need for this anymore.

Issue #966, opened 2016-06-07
Signtool: Add support for description URL (/du)


dbertram:
In addition to the recently added `/d` option (#524), our team also uses the `/du` option to specify an extended description URL for the signed content.

From https://msdn.microsoft.com/en-us/library/8s9b9yaz(v=vs.110).aspx#Anchor_3:
> /du _URL_ - Specifies a Uniform Resource Locator (URL) for the expanded description of the signed content.

PR #967, opened 2016-06-07
GH966: Signtool: Add support for description URL (/du)

dbertram:
From https://msdn.microsoft.com/en-us/library/8s9b9yaz(v=vs.110).aspx#Anchor_3:
>/du _URL_ - Specifies a Uniform Resource Locator (URL) for the expanded description of the signed content.

Adds `SignToolSignSettings.DescriptionUri` and corresponding changes in `SignToolSignRunner`.

### DIFF
--- a/src/Cake.Common/Tools/SignTool/SignToolSignSettings.cs
+++ b/src/Cake.Common/Tools/SignTool/SignToolSignSettings.cs
@@ -33,5 +33,10 @@ namespace Cake.Common.Tools.SignTool
         /// Gets or sets the signed content's description.
         /// </summary>
         public string Description { get; set; }
+
+        /// <summary>
+        /// Gets or sets the signed content's expanded description URL.
+        /// </summary>
+        public Uri DescriptionUri { get; set; }
     }
 }
--- a/src/Cake.Common.Tests/Unit/Tools/SignTool/SignToolSignRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/SignTool/SignToolSignRunnerTests.cs
@@ -1,4 +1,5 @@
-﻿using Cake.Common.Tests.Fixtures.Tools;
+﻿using System;
+using Cake.Common.Tests.Fixtures.Tools;
 using Cake.Core;
 using Cake.Testing;
 using Xunit;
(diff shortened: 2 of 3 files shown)


gep13:
Relates to #966

src/Cake.Common/Tools/SignTool/SignToolSignSettings.cs
```diff
@@ -33,5 +33,10 @@ public sealed class SignToolSignSettings : ToolSettings
         /// Gets or sets the signed content's description.
         /// </summary>
         public string Description { get; set; }
+
+        /// <summary>
+        /// Gets or sets the signed content's expanded description URL.
+        /// </summary>
+        public Uri DescriptionUri { get; set; }
```
gep13: @dbertram @patriksvensson @devlead should this be a Uri?  Arguably, yes, it is a URI, but at the time that Settings class knows about it, it is a string, taken from the command line.  In the GitVersion and GitLink aliases, any URI's were accepted as a string, and then passed to the underlying tool.  I don't think we need to complicate the Cake wrapper, unless we are doing some sort of validation of the URI, in terms of scheme, domain, etc, in which case, converting to a URI would make sense.

Thoughts?

dbertram: I only went with `Uri` here (rather than `string`) to match what was already being done with `SignToolSignSettings.TimestampUri` (which is passed through as `/t` in the same manner).

I'm happy to change this to use string if folks think that's more appropriate.

gep13: @dbertram Gotcha!  I definitely prefer consistency over anything else.  For now, I would suggest that you leave it as Uri, and we need to ensure consistency across the various Settings classes within the Cake.Common project.  For example here:

https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/GitLink/GitLinkSettings.cs#L14

gep13:
@dbertram I have restarted the Travis Build, as it looks like there was a timeout when it was attempting to do the nuget package restore.

dbertram:
@gep13 Looks like nuget restore failed again on that Travis CI build. :/

devlead:
@dbertram kicked the build again.

gep13:
@dbertram this is honestly unlikely anything to do with your PR.  It is just that sometimes Travis has a hard time restoring nuget packages.  We have put in some retry logic using Polly, which normally gets it to work, but looks like there are more serious problems today than normal.

devlead:
@dbertram looks good to me, PR needs rebasing against latest develop though.

dbertram:
@devlead Rebased.

PR #984, opened 2016-06-11
GH970: Adds a default source to NuGet push command

### LINKED ISSUE #970: Add an explicit NuGet source for NuGetPush
As of NuGet client 3.4.something, performing a NuGet push requires a source to be specified (it used to use a default source if none was provided). See NuGet/Home#2504 for more information.

At first glance, I would assume this should go in `NuGetPusher` (https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/NuGet/Push/NuGetPusher.cs#L74) as an alternate branch if `NuGetPushSettings.Source` isn't specified. The bigger question is probably what the default should be...

daveaglick:
Resolves #970

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NuGet/Push/NuGetPusherTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NuGet/Push/NuGetPusherTests.cs
@@ -4,6 +4,7 @@
 using System;
 using Cake.Common.Tests.Fixtures.Tools.NuGet.Pusher;
 using Cake.Common.Tools.NuGet;
+using Cake.Common.Tools.NuGet.Push;
 using Cake.Testing;
 using Cake.Testing.Xunit;
 using Xunit;
@@ -140,7 +141,7 @@ namespace Cake.Common.Tests.Unit.Tools.NuGet.Push
                 var result = fixture.Run();
 
                 // Then
-                Assert.Equal("push \"/Working/existing.nupkg\" -NonInteractive", result.Args);
+                Assert.Equal("push \"/Working/existing.nupkg\" -NonInteractive -Source \"" + NuGetPusher.DefaultPushSource + "\"", result.Args);
             }
 
             [Fact]
@@ -154,7 +155,7 @@ namespace Cake.Common.Tests.Unit.Tools.NuGet.Push
                 var result = fixture.Run();
 
                 // Then
-                Assert.Equal("push \"/Working/existing.nupkg\" 1234 -NonInteractive", result.Args);
+                Assert.Equal("push \"/Working/existing.nupkg\" 1234 -NonInteractive -Source \"" + NuGetPusher.DefaultPushSource + "\"", result.Args);
             }
 
             [Fact]
@@ -169,7 +170,7 @@ namespace Cake.Common.Tests.Unit.Tools.NuGet.Push
 
                 // Then
                 Assert.Equal("push \"/Working/existing.nupkg\" -NonInteractive " +
-                             "-ConfigFile \"/Working/NuGet.config\"", result.Args);
+                             "-ConfigFile \"/Working/NuGet.config\" -Source \"" + NuGetPusher.DefaultPushSource + "\"", result.Args);
             }
 
             [Fact]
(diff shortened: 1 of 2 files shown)


patriksvensson:
@daveaglick I realized that we probably want the default source to respect the default nuget source in the configuration. There is a PR in the pipeline for this (#946) that will add this to the configuration. Would it be possible to update this PR with that functionality once it's merged?

daveaglick:
Good idea - no problem, I'll keep an eye out for the merge.

danielwertheim:
@patriksvensson @daveaglick just saw this and had a quick glance, so I might misunderstand it, but having a default source that anything is pushed to, be it NuGet or Octo, is something I'm a bit... concerned about. Sure you need to have an API-key and/credentials, and when used within companies you could lock down traffic etc., but I'm thinking of the possibility of accidentally pushing something somewhere where it really does not belong.

gep13:
>@danielwertheim said...
just saw this and had a quick glance, so I might misunderstand it, but having a default source that anything is pushed to, be it NuGet or Octo, is something I'm a bit... concerned about.

Just to be clear, this was exactly how NuGet v2 and earlier worked.  Do you have the same concern when using that version of the client?

danielwertheim:
@gep13 when doing open source stuff etc I would not care. When inside a corporate environment I would care for anything that could push by default somewhere. Again, just sharing my thoughts and yes I understand that we are talking about lots if factor that should get combined in the correct way for something to go wrong.

gep13:
@daveaglick would you be in a position to update this PR?  I have just merged the work that @patriksvensson mentioned.

daveaglick:
@gep13 No problem - I can probably get to it tomorrow.

daveaglick:
After looking through the changes in #946, I'm wondering how best to integrate that with this...

That PR adds a default NuGet source for pulling packages from, and if one isn't configured (and wasn't otherwise specified), it doesn't add a `-Source` argument to the NuGet install command.

By contrast, this PR and issue are about a default source for pushing packages to (since NuGet now requires a source to be explicitly specified for push - not supplying one is an error). Should we assume that the default push source will be the same as the default install source? If so, then I can just read from the configuration ala #946. However, as far as I can tell, no default is specified in the configuration automatically. I.e., we'd just change where the user has to specify the source from the push command to the configuration - the user still has to put in an explicit source somewhere.

Does this line of questions make sense? I see these options:
* Use the same configuration option from #946, but give it an automatic default value of the public NuGet gallery. This would also have an impact on #946 by essentially nullifying this check (since it would always have a value and never be null unless the user explicitly unset it): https://github.com/cake-build/cake/pull/946/files#diff-edc69f2c4b33f3b5fe7a59ffcda857a5R187
* Use a different configuration option (say, `PushSource`) and give *that* configuration option an automatic default value of the public NuGet gallery.

Thoughts (it's likely I'm overthinking this...)?

gep13:
@daveaglick I think you are onto something here...

I thought that each configuration value had a default in code, which would be used in the place of a configuration value being passed.  However, this doesn't seem to be the case.  @patriksvensson can you confirm?

With regard to whether the push source should be the same as the pull source, my gut says that 9 times out of ten, they would be the same, but that we should still create a configuration option for them to be different.  Although, that would mean changing the name of the one that has just been created.

@cake-build/team thoughts?

gep13:
@patriksvensson if you are still around, can you provide your thoughts on this one?

patriksvensson:
@gep13 Not sure to be honest. Was there a reason they removed the default push source from nuget?

gep13:
>@patriksvensson said...
@gep13 Not sure to be honest. Was there a reason they removed the default push source from nuget?

No idea :smile: I am sure that they had a great reason for it.

gep13:
@daveaglick been giving this a bit of thought....

If the coomand had indeed changed, and source is now a mandatory parameter, should the alias now change to reflect that by adding a source parameter, and we can obsolete the existing one?

gep13:
@daveaglick @patriksvensson @devlead actually, is this still required?

According to this:

image

It would suggest that it still defaults.

gep13:
Actually, scratch that...

The docs in my local client are out of date:

https://github.com/NuGet/NuGetDocs/pull/459/files#diff-065bfcc9e70f29b5f101907e6268d6ccR624

image

devlead:
If it's now mandatory I'm for an overload taking source as the new preferred alias. But as you can also set DefaultPushSource in nuget.config should we really obsolete the old one?

gep13:
>@devlead said...
But as you can also set DefaultPushSource in nuget.config should we really obsolete the old one?

Hmm, that is a good point!

If we don't Obsolete it, then we need to make crystal clear in the documentation which alias to use, and when.

gep13:
As discussed here: https://github.com/cake-build/cake/issues/970#issuecomment-234931413 I am going to go ahead and close this PR.

PR #990, opened 2016-06-16
GH810: Support for octo push command, common Octopus base classes

### LINKED ISSUE #810: Support Octopus Deploy Push (octo.exe push)
I worked on a project that had me use the `octo.exe push` to deploy a .zip file to an Octopus Deploy server.
We used a handcrafted solution, but I'm wondering if it's worth it to implement with the rest of the Cake tools.

Cake already has support for the Octopus Deploy command line tool `octo.exe` in [src/Cake.Common/Tools/OctopusDeploy](https://github.com/cake-build/cake/tree/develop/src/Cake.Common/Tools/OctopusDeploy), but it's currently limited to `octo.exe create-release` through method [OctoCreateRelease(context, projectName, settings)](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/OctopusDeploy/OctopusDeployAliases.cs#L61).

Related to #472, which is about `pack` instead of `push`, but mentions an `octo.exe` wrapper, which is the current implementation of `create-release`, and was basically my initial working solution.
I'm looking on how to implement a Cake Addin that fits the bill.

For those wondering, the reason I'm not using NuGet push instead of `octo.exe push` is basically because I'm uploading a .zip file that isn't a NuGet package (it has no nuspec). NuGet push will not push a file that is not a NuGet package.
It leverages [a feature from the Octopus Deploy internal feed](http://docs.octopusdeploy.com/display/OD/Supported+Packages) where you can push to push a package file using the filename to specify the package ID and version (`<id>.<version>.zip`).

agc93:
Resolves #810 

This PR integrates support for the octo.exe push command alongside the existing release creation. In order to play nice with the existing alias, I have created the `OctopusDeployPusher` as a separate `Tool<T>`.

Tests have all been added as well.

I have refactored the new `Tool` and `Settings` classes to use a *new* common `OctopusDeployArgumentBuilder` tied to the *existing* `OctopusDeploySettings` class. I have *not* however refactored the existing release creation alias to use this structure. It should be basically drop-in, but I will leave that as a call for someone more familiar with the code base.

This is a much bigger contribution than my usual, so feedback is welcome as always :)

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
@@ -1,0 +1,36 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System.Collections.Generic;
+using Cake.Common.Tools.OctopusDeploy;
+using Cake.Core.IO;
+using Cake.Testing.Fixtures;
+
+namespace Cake.Common.Tests.Fixtures.Tools
+{
+    internal sealed class OctopusDeployPusherFixture : ToolFixture<OctopusPushSettings>
+    {
+        internal string Server { get; set; }
+        internal string ApiKey { get; set; }
+        public List<FilePath> Packages { get; set; }
+
+        public OctopusDeployPusherFixture()
+            : base("Octo.exe")
+        {
+            Packages = new List<FilePath>
+            {
+                "MyPackage.1.0.0.zip",
+                "MyOtherPackage.1.0.1.nupkg"
+            };
+
+            Server = "http://octopus";
+            ApiKey = "API-12345";
+        }
+
+        protected override void RunTool()
+        {
+            var tool = new OctopusDeployPusher(FileSystem, Environment, ProcessRunner, Tools);
+            tool.PushPackage(Server, ApiKey, Packages == null ? null : Packages.ToArray(), Settings);
+        }
+    }
+}
--- a/src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPushTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPushTests.cs
@@ -1,0 +1,328 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System.Collections.Generic;
+using Cake.Testing;
+using Cake.Common.Tests.Fixtures.Tools;
+using Cake.Core.IO;
+using Cake.Testing.Xunit;
+using Xunit;
+
+namespace Cake.Common.Tests.Unit.Tools.OctopusDeploy
+{
+    public sealed class OctoPushTests
+    {
+        public sealed class TheBaseArgumentBuilder
+        {
+            [Fact]
+            public void Should_Throw_If_Server_Is_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Server = null;
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentException(result, "settings", "No server specified.");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Api_Key_Is_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.ApiKey = null;
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentException(result, "settings", "No API key specified.");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Settings_Is_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings = null;
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentNullException(result, "settings");
+            }
+
+            [Fact]
+            public void Should_Add_Username_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.Username = "mike123";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" --package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345 " +
+                             "--user \"mike123\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Password_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.Password = "secret";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" --package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345 " +
+                             "--pass \"secret\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Configuration_File_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.ConfigurationFile = "configFile.txt";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" --package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345 " +
+                             "--configFile \"/Working/configFile.txt\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Debug_Flag_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.EnableDebugLogging = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" --package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345 " +
+                             "--debug", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Ignore_Ssl_Errors_Flag_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.IgnoreSslErrors = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" --package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345 " +
+                             "--ignoreSslErrors", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Enable_Service_Messages_Flag_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.EnableServiceMessages = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" --package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345 " +
+                             "--enableServiceMessages", result.Args);
+            }
+        }
+        public sealed class ThePushPackageMethod
+        {
+            [Fact]
+            public void Should_Throw_If_No_Packages_Provided()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Packages = new List<FilePath>();
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentNullException(result, "packagePaths");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Package_List_Is_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Packages = null;
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentNullException(result, "packagePaths");
+            }
+
+            [Fact]
+            public void Should_Add_Single_Argument_For_Single_Package()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Packages = new List<FilePath> { "MyPackage.1.0.0.zip" };
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" " +
+                             "--server http://octopus --apiKey API-12345", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Multiple_Arguments_For_Multiple_Packages()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Packages = new List<FilePath>
+                {
+                    "MyPackage.1.0.0.zip",
+                    "MyOtherPackage.1.0.1.nupkg",
+                    "MyExtraPackage.1.0.1.nupkg"
+                };
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" " +
+                             "--package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--package \"/Working/MyExtraPackage.1.0.1.nupkg\" " +
+                             "--server http://octopus --apiKey API-12345", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Replace_Existing_Flag_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
+                fixture.Settings.ReplaceExisting = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("push --package \"/Working/MyPackage.1.0.0.zip\" " +
+                             "--package \"/Working/MyOtherPackage.1.0.1.nupkg\" " +
+                             "--replace-existing " +
+                             "--server http://octopus --apiKey API-12345", result.Args);
+            }
+
+            [Fact]
+            public void Should_Throw_If_Octo_Executable_Was_Not_Found()
+            {
+                // Given
+                var fixture = new OctopusDeployPusherFixture();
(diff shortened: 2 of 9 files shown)


src/Cake.Common/Tools/OctopusDeploy/PushPackageSettings.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
gep13: @agc93 from a consistency stand point, this extra line should be removed here.  Personally, I think we do need a linebreak here, but let's keep things the same for now.

src/Cake.Common/Tools/OctopusDeploy/PushPackageSettings.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
```
gep13: @agc93 none of these using statements are actually required, and can be removed.

src/Cake.Common/Tools/OctopusDeploy/PushPackageSettings.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    /// <summary>
+    /// Contains settings used by <see cref="OctopusDeployPusher.PushPackage"/>.
+    /// </summary>
+    public sealed class PushPackageSettings : OctopusDeploySettings
```
gep13: @agc93 I think this class should be called `OctopusPushSettings` to be consistent with other classes in this section of the code.

gep13: @agc93 deriving from `OctopusDeploySettings` doesn't feel right to me.  I think I would prefer to see something like `OctopusBaseSettings` which both classes derive from.  @patriksvensson @devlead what are your thoughts here?

src/Cake.Common/Tools/OctopusDeploy/PushPackageArgumentBuilder.cs
```diff
@@ -0,0 +1,42 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System.Collections.Generic;
+using System.Linq;
+using Cake.Core;
+using Cake.Core.IO;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    internal class PushPackageArgumentBuilder : OctopusDeployArgumentBuilder<PushPackageSettings>
```
gep13: @agc93 again, from a naming point of view, I think this should be `OctopusPushArgumentBuilder`

gep13: @agc93 and also, the same comment as before, deriving from the deploy class feels "wrong" to me.  Perhaps a common class, which both derive from?  I think there are examples of using this approach in the code base.  @devlead @patriksvensson thoughts?

src/Cake.Common/Tools/OctopusDeploy/PushPackageArgumentBuilder.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private readonly List<FilePath> _packagePaths;
+
+        public PushPackageArgumentBuilder(IEnumerable<FilePath> packagePaths, PushPackageSettings settings, ICakeEnvironment environment) : base(environment, settings)
```
gep13: @agc93 small nitpick.... can you put the `: base(...` on a new line?

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPusher.cs
```diff
@@ -0,0 +1,82 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
gep13: @agc93 there are some trailing whitespaces in this file.  Can you remove?

image

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPusher.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <param name="fileSystem">The file system.</param>
+        /// <param name="environment">The environment.</param>
+        /// <param name="processRunner">The process runner.</param>
+        /// <param name="tools">The tool locator.</param>
+        public OctopusDeployPusher(
+            IFileSystem fileSystem, 
+            ICakeEnvironment environment, 
+            IProcessRunner processRunner, 
+            IToolLocator tools) : base(fileSystem, environment, processRunner, tools)
```
gep13: @agc93 `: base(...` on a new line please

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPusher.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                throw new ArgumentNullException("packagePaths");
+            }
+            if (settings == null)
+            {
+                throw new ArgumentNullException("settings");
+            }
+            if (string.IsNullOrEmpty(settings.Server))
+            {
+                throw new ArgumentException("No server specified.", "settings");
```
gep13: @agc93 typically, all required settings are passed into the Alias as individual parameters, rather than including them in the Settings class.  That way it is explicit that they are expected to be there.  I would suggest making the alias the following:

```
 public void PushPackage(string server, string apiKey, FilePath[] packagePaths, PushPackageSettings settings)
```

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployArgumentBuilder.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
```
gep13: @agc93 can we get the license header in this file?

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployArgumentBuilder.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
```
gep13: @agc93 can you remove the unrequired usings?

image

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployArgumentBuilder.cs
```diff
@@ -0,0 +1,76 @@
...
+                Builder.Append("--ignoreSslErrors");
+            }
+
+            if (Settings.EnableServiceMessages)
+            {
+                Builder.Append("--enableServiceMessages");
+            }
+        }
+    }
+}
```
gep13: @agc93 there is an extra line at the end of this file which can be removed :smile:

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployAliases.cs
```diff
@@ -2,8 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http.Headers;
```
gep13: @agc93 this using statement is not required.

src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPushTests.cs
```diff
@@ -0,0 +1,331 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
```
gep13: @agc93 we can lose some usings here:

image

gep13: @agc93 also, Cake.Testing should be after Cake.Common...

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+using Cake.Common.Tools.OctopusDeploy;
+using Cake.Core.IO;
+using Cake.Testing.Fixtures;
+
+namespace Cake.Common.Tests.Fixtures.Tools
+{
+    internal sealed class OctopusDeployPusherFixture : ToolFixture<PushPackageSettings>
+    {
+        public List<FilePath> Packages { get; set; }
+        public OctopusDeployPusherFixture() : base("Octo.exe")
```
gep13: @agc93 can we get the `: base(...` on a new line?

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using Cake.Common.Tools.OctopusDeploy;
+using Cake.Core.IO;
+using Cake.Testing.Fixtures;
+
+namespace Cake.Common.Tests.Fixtures.Tools
+{
+    internal sealed class OctopusDeployPusherFixture : ToolFixture<PushPackageSettings>
+    {
+        public List<FilePath> Packages { get; set; }
```
gep13: @agc93 can we get a new line at the end of this one?

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+        public OctopusDeployPusherFixture() : base("Octo.exe")
+        {
+            Packages = new List<FilePath>
+            {
+                "MyPackage.1.0.0.zip",
+                "MyOtherPackage.1.0.1.nupkg"
+            };
```
gep13: @agc93 can we get a new line after this?

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+            Settings.Server = "http://octopus";
+            Settings.ApiKey = "API-12345";
+        }
+
+        protected override void RunTool()
+        {
+            var tool = new OctopusDeployPusher(FileSystem, Environment, ProcessRunner, Tools);
+            tool.PushPackage(Settings, Packages == null ? null : Packages.ToArray());
```
gep13: @agc93 is there a reason that the PushPackage method can't simply accept the List<FilePath>

gep13:
@agc93 thank you very much for your continued contributions to Cake, and thanks for tackling this larger issue.  I have left some comments in the code.  Nothing major, just a few small things.  If you can fix them up, adn then rebase on the most recent develop, I can take another look.  Thanks!

src/Cake.Common/Tools/OctopusDeploy/PushPackageArgumentBuilder.cs
agc93: I might be missing something, but isn't that exactly what `OctopusDeployArgumentBuilder` is? A common base class? Is this just a naming thing?

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
agc93: Not really. I ummed and ahhed about using `params` vs the `List` and was originally planning on doing the actual alias with a `params` but then changed my mind. I'll switch it to a `List` and simplify it.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPusher.cs
agc93: Should this be cleaned up in the existing `ReleaseCreator` code, or leave it for now?

src/Cake.Common/Tools/OctopusDeploy/PushPackageSettings.cs
agc93: Same as with `OctopusDeployArgumentBuilder`, isn't the existing `OctopusDeploySettings` `abstract` class exactly that? I think the naming (the product actually being called Octopus **Deploy**) is a lot of the confusion here.

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
agc93: Okay, not sure what needs the new line here?

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
agc93: Also not sure here. Are there some documented rules on whitespacing and newlining? Or maybe some ReSharper inspections?

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
gep13: @agc93 sounds like a plan to me :+1:

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
gep13: @agc93 I was referring to this:

image

I don't think we have a fixed rule for this particular situation.

src/Cake.Common.Tests/Fixtures/Tools/OctopusDeployPusherFixture.cs
gep13: @agc93 I was referring to this:

image

i.e. new line between property and constructor.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPusher.cs
gep13: @agc93 ah, I hadn't realised that they were already being used in the CreateRelease alias.  I think we should put the duplication in place just now, and then create a new issue to correct the status of the other alias.  This will mean that we will have to obsolete that alias across a couple versions of Cake though, so an issue to track this would be very useful.  Can you create that issue?  @devlead @patriksvensson what are your throughts on this?

src/Cake.Common/Tools/OctopusDeploy/PushPackageArgumentBuilder.cs
gep13: @agc93 yes, sorry, you are right.  I was misreading the class name.  I thought there was a deploy method being wrapped up as well.  In which case, ignore this comment, nothing to see here :smile:

src/Cake.Common/Tools/OctopusDeploy/PushPackageSettings.cs
gep13: @agc93 yes, it is the naming (and me not paying attention).  You are spot on here though, so leave this as is.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPusher.cs
agc93: Okay, so just to be clear I will include `server` and `apiKey` in the constructor parameters, but which do we trust: the raw params or the values from the settings object?

Also, just so I am aware, what is the reasoning behind not putting these in a settings class since they are going to be common to all Octopus Deploy aliases (including future ones, if the case arises)?

agc93:
Alright so I have made all the whitespace corrections and a couple of the refactorings (keeping it local since last time has made me scared of `squash`). 

So the only outstanding point is the common arguments one above of moving `server` and `apiKey` out of the constructor (and potential future removal) in favour of alias parameters.

I must admit I'm not 100% on the rationale for that as it seems like the sort of thing that should be in a common class (as per existing design). Admittedly, I'm not as familiar with the existing conventions so happy to defer to you ( @gep13 ), or @devlead or @patriksvensson on this one.

gep13:
@agc93 ha ha, I know exactly what you mean about squash/rebase.  It was something that I was very much wary of before I got used to it.  If you want, when it gets to that stage, we can maybe have a skype call, or something, and we can walk through it.

In terms of the rationale behind my comments for the parameters, it basically comes down to this.  We always want things to "just" work.  By that I mean, we want all aliases to be explicit about what parameters are required in order for it to function.  In the case of this alias, the server and apikey are required, otherwise it will not function.  While we could simply have these in the settings class, the user won't know from just looking at the alias definition, that those things have to be set.  By making them direct input parameters, we are making it as clear as possible that they are required.  All other things that "can" be set, as in the settings class.

The is the same thought process behind things like the MSBuild alias.  The solution file path is an input parameter, rather than a property in the MSBuildSettings class.

You while i agree, we want to make sure that we re-use as much of the common settings class as possible, I don't want to do that whilst losing clarity on the intention of the alias. and how it operates.

Does that make sense?

wwwlicious:
While I understand the rationale, this approach works fine until you need to change what is 'required' then you break the contract (method signature) and is IMO always painful. A single settings param is more flexible and will outlive the 'required' fields....always.

If docs and decent error messages are available when calling something without its required properties then it really isn't much of a hassle to call it a few times, figure out what's missing and add it. 

Just my two cents.

patriksvensson:
@wwwlicious Removing a parameter from the settings would also be breaking. Not removing the required setting (but obsoleting it) would not break it per se but would introduce a semantic change, which is even worse in my opinion. Showing required parameters in the alias method signature will give a clear picture to the consumer what is required and not.

wwwlicious:
@patriksvensson I am not advocating making a breaking change (or obsoleting) existing api's here or elsewhere, my comment is simply to say that a 'required fields' pattern is more fragile to a consumer than a message based approach where you pass a single settings param to the method for the long term and therefore not one I would personally advocate. 

Doc's, sensible defaults and guard clauses can easily identify what is required from what isn't. Contract stability is more important to me over the lifespan of a project when consuming 3rd party code than the short amount of time spent when initially wiring it up.

As I said, just my opinion so I'm not jumping up and down about it or anything! :cake:

agc93:
@gep13 I will do a commit now of what I have and I might take you up on that offer once this is sorted :smile:

agc93:
Okay, so taking the "where should required fields" part aside, still leaves the challenge of the server address and API key arguments which now a) have to be separately implemented by each Octopus command implementation despite being common arguments and b) will now be in both the Settings class and the method call, meaning one has to be trusted over the other. 

For a) this seems a little counter-intuitive to me since we are deliberately taking known common functionality (usually a base class in a nutshell) and moving it out of an existing base class, with all the deprecation problems that might introduce.. as per earlier, happy to make the change here, but that will mean that the create release still needs to be updated (with all that comes with that).

For b) I assume judging by the rationale so far, we would simply ignore any server address or API keys passed in the settings object since we think they will be removed in a future release. Does that sound correct?

patriksvensson:
@agc93 Yes, it might seem counter constructove, but I see the greater benefit of a consistent API. This benefit outweights the implementation cost in my opinion.

gep13:
@agc93 let me know when you are in a position to look at this, and to fix up the remaining issues, and we can have a chat about how to do the merge/squash/rebase.

agc93:
@gep13 I had a go at being self-sufficient, how does that look? :smile:

gep13:
@agc93 after a very brief look, looks like the rebase/squashing has worked :+1:   I will review it properly again, and I will get back to you if I have any comments.

agc93:
Is there anything outstanding on this? Should I rebase again now or is it likely to take a while?

gep13:
@agc93 I took the liberty of doing the rebase and merging into develop as there were a couple of small corrections that I thought were needed.  I have merged it here: https://github.com/cake-build/cake/commit/e864c2287817c32d402fa56779d39743641603eb As a result, I will close this PR, but the work has been included in the main develop branch, it is just that GitHub can't follow the flow.  Thanks again!

gep13:
@agc93 can I get you to raise an issue about removing the duplicated properties in the Settings files, that are now mandatory properties in the aliases? Thanks!

PR #994, opened 2016-06-21
resolved error in documentation example

ghuntley:


gep13:
@ghuntley good catch, thank you very much for correcting!

PR #1008, opened 2016-06-24
GH502: Refactored ICakeEnvironment

### LINKED ISSUE #502: Refactor ICakeEnvironment
Right now, `ICakeEnvironment` is kind of a mess, mixing properties and methods for things.

I'm also not a fan of IsUnix-property which I think could be more fine grained i.e. we could add a OperatingSystem enum that gives more information such as Windows, OSX, Linux etc. We could even make this a flags enum and add additional more finegrained information such as what operating system, but not sure if that would be useful.

I would suggest that we make some changes.

It would make sense to introduce something like `IPlatform` that could keep information about OS and hardware. The ICakeEnvironment have a tendency to become a dumping ground for everything environment related and this would mean that we would keep platform specific stuff isolated.

* Move `bool Is64BitOperativeSystem()` to `IPlatform`
* Move `bool IsUnix()` to `IPlatform`
* Change `DirectoryPath GetApplicationRoot();` to `DirectoryPath ApplicationRoot { get; }`
* Add `IPlatform Platform { get; }`

```csharp
public interface IPlatform
{
  OperatingSystem OperatingSystem { get; }
  bool Is64Bit { get; }
  bool IsUnix { get; } // Could be an extension method for IPlatform
}
```

```csharp
public enum OperatingSystem
{
  Unknown = 0,
  Windows = 1,
  Linux = 2,
  OSX = 3
}
```

Not sure if we should expose the enum directly, or expose a class that 
contains information about the operating system (type/flavor, name, version) etc. 
Not sure if this information would be of interest to Cake users. 

**NOTE: Since this is a breaking change, we wouldn't actually remove the methods but mark them as obsolete and remove them sometime in the future.**

I will glady do this changes if you let me :smile:

patriksvensson:
This included obsoleting a lot of members and also
introducing two concepts, `ICakePlatform` and `ICakeRuntime`.

**ICakePlatform** is an abstraction around the actual platform
that Cake is running on. This will give you information about
things such the platform family and architecture.

```
public interface ICakePlatform
{
    PlatformFamily Family { get; }
    bool Is64Bit { get; }
}
```

```
public enum PlatformFamily
{
    Unknown = 0,
    Windows = 1,
    Linux = 2,
    OSX = 3
}
```

**ICakeRuntime** is an abstraction around the runtime that Cake
is running in such as the .NET target platform.

```
public interface ICakeRuntime
{
    FrameworkName TargetFramework { get; }
}
```

Closes #502

### DIFF
--- a/src/Cake.Core/CakeEnvironment.cs
+++ b/src/Cake.Core/CakeEnvironment.cs
@@ -16,6 +16,10 @@ namespace Cake.Core
     /// </summary>
     public sealed class CakeEnvironment : ICakeEnvironment
     {
+        private readonly ICakePlatform _platform;
+        private readonly ICakeRuntime _runtime;
+        private readonly DirectoryPath _applicationRoot;
+
         /// <summary>
         /// Gets or sets the working directory.
         /// </summary>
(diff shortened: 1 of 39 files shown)


src/Cake.Testing/FakeRuntime.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// An implementation of a fake <see cref="ICakeRuntime"/>
+    /// </summary>
+    public sealed class FakeRuntime : ICakeRuntime
+    {
+        /// <summary>
+        /// Gets or sets the target .NET framework version that the current AppDomain is targeting.
+        /// </summary>
+        public FrameworkName TargetFramework { get; set; }
```
patriksvensson: Add default value here. Probably `.NETFramework,Version=v4.5`.

src/Cake.Core/ICakeEnvironment.cs
```diff
@@ -57,9 +46,43 @@ public interface ICakeEnvironment
...
+        /// </summary>
+        /// <returns>Whether or not the current operative system is 64 bit.</returns>
+        [Obsolete("Please use ICakeEnvironment.Platform.Is64Bit instead.")]
+        bool Is64BitOperativeSystem();
+
+        /// <summary>
+        /// Determines whether the current machine is running Unix.
+        /// </summary>
+        /// <returns>Whether or not the current machine is running Unix.</returns>
+        [Obsolete("Please use ICakeEnvironment.Platform.IsPosix instead.")]
```
patriksvensson: Wrong name of method.

src/Cake.Core/CakeEnvironment.cs
```diff
@@ -125,16 +128,49 @@ public string GetEnvironmentVariable(string variable)
...
+            return _platform.Is64Bit;
+        }
+
+        /// <summary>
+        /// Determines whether the current machine is running Unix.
+        /// </summary>
+        /// <returns>
+        /// Whether or not the current machine is running Unix.
+        /// </returns>
+        [Obsolete("Please use CakeEnvironment.Platform.IsPosix instead.")]
```
patriksvensson: Wrong name of method.

src/Cake.Testing/FakeEnvironment.cs
```diff
@@ -117,76 +149,64 @@ public void SetSpecialPath(SpecialPath kind, DirectoryPath path)
...
 
         /// <summary>
-        /// Sets an environment variable.
+        /// Determines whether the current machine is running Unix.
         /// </summary>
-        /// <param name="variable">The variable.</param>
-        /// <param name="value">The value.</param>
-        public void SetEnvironmentVariable(string variable, string value)
+        /// <returns>Whether or not the current machine is running Unix.</returns>
+        [Obsolete("Please use FakeEnvironment.Platform.IsPosix instead.")]
```
patriksvensson: Wrong name of method.

patriksvensson:
Should we add an extension method to `ICakeEnvironment` called `IsUnix` for convenience and backwards compatability? Not sure how we going to solve it since the actual method needs to be obsoleted though.

patriksvensson:
@gep13 @devlead This one is ready for review.

devlead:
@patriksvensson LGTM tested on several projects. I've rebased and merged into develop.

PR #1016, opened 2016-06-28
dotnet CLI preview 2 doesn't have the quiet option anymore

adamhathcock:
Just noticed this.

### DIFF
--- a/src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestoreSettings.cs
+++ b/src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestoreSettings.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
+
+using System;
 using System.Collections.Generic;
 using Cake.Core.IO;
 
@@ -39,6 +41,7 @@ namespace Cake.Common.Tools.DotNetCore.Restore
         /// <summary>
         /// Gets or sets a value indicating whether to display any output.
         /// </summary>
+        [Obsolete(".NET CLI does not support this option anymore. Please use DotNetCoreRestoreSettings.Verbosity instead.")]
         public bool Quiet { get; set; }
 
         /// <summary>
--- a/src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestorer.cs
+++ b/src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestorer.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 using System;
 using Cake.Core;
+using Cake.Core.Diagnostics;
 using Cake.Core.IO;
 using Cake.Core.Tooling;
 
@@ -14,6 +15,7 @@ namespace Cake.Common.Tools.DotNetCore.Restore
     public sealed class DotNetCoreRestorer : DotNetCoreTool<DotNetCoreRestoreSettings>
     {
         private readonly ICakeEnvironment _environment;
+        private readonly ICakeLog _log;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DotNetCoreRestorer" /> class.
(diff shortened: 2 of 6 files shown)


adamhathcock:
Looks like the builds are failing because your build script on the dotnet cli also uses the Quiet option.

patriksvensson:
@adamhathcock Great catching this!

The reason for the builds failing is that the tests for DotNetRestore has not been updated.
I'm not sure if we want to remove the property directly due to backwards compatibility. I think we would want to obsolete it with a warning for now and refer to using the Verbosity parameter and remove it in a couple releases.

adamhathcock:
I'll update

adamhathcock:
Now it's failing because of Warning As Error.  Should I just remove the Quiet attribute?

patriksvensson:
@adamhathcock Yes. I think you can remove that attribute for now and just write to the log in the Tool (as a warning) that the flag has been removed from dotnet cli RTM and they should set verbosity instead.

devlead:
Isn't this just that the tests should pragma ignore the obsolete attribute?

patriksvensson:
@devlead @adamhathcock Yes, that would be preferred.

src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestoreSettings.cs
```diff
@@ -38,7 +40,8 @@ public sealed class DotNetCoreRestoreSettings : DotNetCoreSettings
 
         /// <summary>
         /// Gets or sets a value indicating whether to display any output.
-        /// </summary>
+        /// </summary>
+        [Obsolete("dotnet cli does not have this option")]
```
patriksvensson: Can we change this to `.NET CLI does not support this option anymore. Please use DotNetCoreRestoreSettings.Verbosity instead.`

adamhathcock:
finally did the pragmas in all the correct places.

src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestorer.cs
```diff
@@ -101,12 +101,13 @@ private ProcessArgumentBuilder GetArguments(string root, DotNetCoreRestoreSettin
...
                 }
             }
-
+#pragma warning disable 0618
             // Quiet
-            if (settings.Quiet)
+            if (settings.Quiet && !settings.Verbosity.HasValue)
             {
-                builder.Append("--quiet");
+                settings.Verbosity = DotNetCoreRestoreVerbosity.Minimal;
```
patriksvensson: I don't think we should set this propery automatically. Better to write to the log as a warning giving the same hint as the obsolete message.

gep13:
@adamhathcock are you in a position to update with the comment that was made by @patriksvensson

adamhathcock:
I've been on vacation for the past two weeks.  I can fix it on Wednesday

adamhathcock:
Looking at this now: I need a `ICakeContext` to do a log item but the infrastructure doesn't look well supported to that from a tool.

Making the tool take the context involves changing the tests a lot at least.

adamhathcock:
Nevermind, I was being dumb.  I think I've done everything requested....but now I just noticed I have to update the branch again

adamhathcock:
@gep13 @patriksvensson finally updated and passed.  Needs a small review again

adamhathcock:
@gep13 should I create an issue?

gep13:
@adamhathcock yes, if you could.  Ideally, an issue is created, which gives us a place to discuss the problem/feature before any work is undertaken, and then we use the issue title to automatically generate the release notes for each release.

adamhathcock:
I thought it was a simple fix so I didn't...next time I promise I will :)

gep13:
>@adamhathcock said...
I thought it was a simple fix so I didn't...next time I promise I will :)

Not a problem at all.  For anything that is a documentation change, we typically don't want/need issues created, however, anything that is a code change, we would like to see an issue getting raised/discussed first, just in case something is already being worked on, or not required, etc.  The last thing we want is people doing work that isn't required, and then having to close the PR.  Obviously, that won't be the case here, but we like to follow the same process, so that there is some consistency, and people know what to expect.

gep13:
Relates to #1073

gep13:
LGTM! Thanks again for the contribution.

PR #1017, opened 2016-06-28
Ignore things

adamhathcock:
Just ignoring some files when looking through.

I would do more but I don't want to overlap.


gep13:
@patriksvensson can you confirm if this is still required?  Or has this already been taken care of?  If so, can you go ahead and close this PR?

@adamhathcock thank you for your contribution, however, in this case, it might not be required.

PR #1018, opened 2016-06-28
Bootstrapping without mono in mingw/cygwin.

### LINKED ISSUE #603: Check for Git Bash/Cygwin in build.sh bootstrapper
When using Git Bash or Cygwin on Windows and using `build.sh`, `cake.exe` is executed under mono by default whereas mono probably isn't installed.

The script should probably check for the environment and only execute `cake.exe` under mono if necessary, otherwise invoke directly.

qhris:
- Fixes #603

Running `./build.sh` under mingw/msys or cygwin on windows complains about not finding the `mono` command which stops the build. This PR fixes that issue.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,6 +9,7 @@ SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 TOOLS_DIR=$SCRIPT_DIR/tools
 NUGET_EXE=$TOOLS_DIR/nuget.exe
 CAKE_EXE=$TOOLS_DIR/Cake/Cake.exe
+MONO_CMD=mono
 
 # Define default arguments.
 SCRIPT="build.cake"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -19,8 +20,15 @@ DRYRUN=
...
+# Run without mono on windows platforms.
+case "$(uname -s)" in
+    CYGWIN*) unset MONO_CMD ;;
+    MINGW32*) unset MONO_CMD ;;
+    MSYS*) unset MONO_CMD ;;
+esac
+
 # Parse arguments.
-for i in "$@"; do
+while [[ $# -gt 0 ]]; do
```
qhris: Changing the evaluation here because on at least mingw the command `./build.sh --target Run-Unit-Tests` produces an empty string to be evaluated for `SCRIPT_ARGUMENTS`; ultimately causing `cake.exe` to think we are passing multiple scripts and fails.

qhris:
Notice: Cygwin still fails because `build.sh` uses CRLF-style line endings. Running the script through a command such as `dos2unix` fixes that issue. I did not include that change here, but can do upon request.

qhris:
Didn't think of it earlier but won't this become unnecessary once we're completely ported to .NET Core?

gep13:
@qhris thank you very much for your contribution to Cake, however, in this instance, I don't think we are going to pull this in.  There is active work on porting to CoreClr.  Once that is complete, we can look into this again, if it is still required.

Issue #1023, opened 2016-06-29
Tool should support other exit codes than 0


Sam13:
Currently a ```CakeException``` is thrown when a tool has other exit code than zero.

Since they are tools around (e.g. RoboCopy) which have different exit codes I suggest to add a list of valid exit codes to the ```ToolSettings``` class.

Or add a custom validation callback (```Func<int, bool>```) to validate the exit code on client side (similar to ```postAction```) of ```Tool.Run()```

devlead:
Fixed by  #1031

PR #1031, opened 2016-07-06
GH1023: Support custom exit code validation for tool.

Sam13:
Added custom exit code validation for `Tool` by adding a `Func<int, bool>` lamda expression to the `Tool.Run` method. Similar to `postAction`.
This can be used when developing custom tool support e.g. for Robocopy which runs sucessful but provides other exit codes than zero.

### DIFF
--- a/src/Cake.Core.Tests/Fixtures/DummyToolFixture.cs
+++ b/src/Cake.Core.Tests/Fixtures/DummyToolFixture.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
+
+using System;
 using Cake.Core.Tests.Stubs;
 using Cake.Testing.Fixtures;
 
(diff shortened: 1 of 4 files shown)


patriksvensson:
@Sam13 Thank you for this. I would prefer to have a virtual method in the Tool base class for this though that every tool can override.

```csharp
public virtual void ProcessExitCode(int exitCode)
{
   if(exitCode != 0)
   {
      // The current throw logic.
   }
}
```

Sam13:
@patriksvensson Ok, I can change. I thought I make it consistent to the `postAction`.

By the way: It would be nice if StyleCop runs also inside VS and activate `WarningsAsErrors ` in the project files...

patriksvensson:
@Sam13 It's disabled for Debug, but if you switch to Release configuration it will behave as expected.

Sam13:
@patriksvensson Got it, thanks

devlead:
@Sam13 LGTM :+1:, could you rebase into 1 commit and we'll do a final review. ( this references #1023 )

Sam13:
Done

devlead:
@Sam13 Merged, thanks for your contribution :+1:

Sam13:
Glad you like, more to come (if time allows ;-)

PR #1048, opened 2016-07-12
Return custom message for specific error codes in NUnit/NUnit3 tools

### LINKED ISSUE #517: NUnit runner: Handle specific non-zero exit codes
Right now, any time the NUnit console tool returns a non-zero exit code, the Cake NUnit runner throws a `CakeException` with the less-than-helpful error message, `NUnit: Process returned an error.`

I think this behavior is too broad, as it treats failed tests the same as an unexpected problem.
I'd like the flexibility to decide whether to blow up my build based on test failures, errors, etc.

NUnit does have meaningful exit codes ([discussed here](http://stackoverflow.com/questions/3889577/does-anyone-know-where-to-find-nunit-console-exit-codes-meanings)), so we should throw more specific exceptions as appropriate.

ChrisMaddock:
Fixes #517.

Uses the new custom error code functionality in #1031 for the NUnit tools. Have been eyeing this up for a while, however now the frameworks already been discussed and put in, it seemed too simple not to do.

I've never used XUnit before - is that `Theory/InlineData` thing the correct approach to doing this sort of test?

For reference:
- [NUnit2 exit codes](https://github.com/nunit/nunitv2/blob/master/src/ConsoleRunner/nunit-console/ConsoleUi.cs#L26)
- [NUnit3 exit codes](https://github.com/nunit/nunit/blob/master/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L44)

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
@@ -148,19 +148,25 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
                 Assert.Equal("NUnit3: Process was not started.", result.Message);
             }
 
-            [Fact]
-            public void Should_Throw_If_Process_Has_A_Non_Zero_Exit_Code()
+            [Theory]
+            [InlineData(10, "NUnit3: 10 test(s) failed (exit code 10).")]
+            [InlineData(-1, "NUnit3: Invalid argument (exit code -1).")]
+            [InlineData(-2, "NUnit3: Invalid assembly (exit code -2).")]
+            [InlineData(-4, "NUnit3: Invalid test fixture (exit code -4).")]
+            [InlineData(-100, "NUnit3: Unexpected error (exit code -100).")]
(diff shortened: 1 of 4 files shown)


gep13:
@ChrisMaddock thanks for making this contribution!

As you will have noticed, our CLA bot has identified that you need to sign our CLA, but it helpfully hasn't told you where that is.  You can find it here:

https://cla2.dotnetfoundation.org/

Are you in a position to sign it for us? Thanks!

ChrisMaddock:
Thanks @gep13 - was wondering what to do about that! Done!

gep13:
@ChrisMaddock perfect!  We are looking into what is causing the issue.  Now that you have signed it though, you won't need to sign it again :smile:

gep13:
@ChrisMaddock LGTM! Thank you for your contribution!

I have merged the changes here https://github.com/cake-build/cake/commit/a509518b0dcd2a2553b262d7e8f1db98ae244395 to save you doing a rebase.

GitHub won't follow what I did, so I will close this PR, but your commit is still intact.  Thanks again!

ChrisMaddock:
Great, thanks!

PR #1051, opened 2016-07-13
Improved documentation for ProcessSettings Timeout property

ghuntley:


gep13:
@ghuntley this is a great addition, thanks for taking the time!

Issue #876, opened 2016-05-12
Add possiblity to launch debugger


mholo65:
Once PR #858 is merged, it might be useful to have the possibility to let Cake.exe launch a debugger using **System.Diagnostics.Debugger.Launch()** instead of waiting for debugger to be attached.

Maybe waiting for debugger to be attached could be canceled by user, and then on **#debug** pre-processor directive (as mentioned in issue #869) we could do Debugger.Launch() if debugger isn't attached...

gep13:
@mholo65 is this something that you will be in a position to look at, or would you like one of us to take it?  Thanks!

mholo65:
@gep13 yes, I can look at this. But to make this is as elegant as possible, then at least some bit of Cake.Core (e.g. ICakeEnvironment) needs to know if we are running debug or not. Hence the reference to #902

By doing this we could use same pre-processor directive and the change these [BreakDirectiveProcessor.cs](https://github.com/cake-build/cake/blob/d35a5f697aaf53cf9997058df35dd4d1adb3be83/src/Cake.Core/Scripting/Processors/BreakDirectiveProcessor.cs#L24-32) lines to:
```csharp
replacement = null;

if (!line.Trim().Equals("#break", StringComparison.Ordinal))
{
    return false;
}

if(_environment.IsDebug)
{
    replacement = @"if (System.Diagnostics.Debugger.IsAttached) { System.Diagnostics.Debugger.Break(); } else { System.Diagnostics.Debugger.Launch(); }";
}

return true;
```

Result will be that ```#break``` directive will be commented out if ```--debug``` flag was not supplied to Cake.exe, otherwise debugger will break or launch depending on if it's attached or not.

What's your opinion on this?

gep13:
@mholo65 I agree, I think we need that before attempting this.  @patriksvensson were you planning on attempting #502 in the 0.13.0 release, or should we bump this particular issue onto the next milestone?

patriksvensson:
@mholo65 What would the non-elegant solution look like until we've fixed the environment refactoring?

mholo65:
@patriksvensson well, it's probably by adding a new directive ```#debug-launch```, or similar

[...]

PR #1074, opened 2016-07-18
[WIP] DO NOT MERGE! Added IsDebug to CakeRuntime.

### LINKED ISSUE #902: Add IsDebug to ICakeEnvironment

mholo65:
Resolves #902

### DIFF
--- a/src/Cake.Core/CakeRuntime.cs
+++ b/src/Cake.Core/CakeRuntime.cs
@@ -17,15 +17,62 @@ namespace Cake.Core
         public FrameworkName TargetFramework { get; private set; }
 
         /// <summary>
+        /// Gets a value indicating whether the current execution is running in debug mode.
+        /// </summary>
+        public bool IsDebug { get; private set; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="CakeRuntime"/> class.
         /// </summary>
-        public CakeRuntime()
+        /// <param name="arguments">The arguments.</param>
+        public CakeRuntime(ICakeArguments arguments)
         {
+            if (arguments == null)
+            {
+                throw new ArgumentNullException("arguments");
+            }
+
             // Try to get the current framework name from the current application domain,
             // but if that is null, we default to .NET 4.5. The reason for doing this is
             // that this actually is what happens on Mono.
             var frameworkName = AppDomain.CurrentDomain.SetupInformation.TargetFrameworkName;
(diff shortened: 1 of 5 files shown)


src/Cake.Core/CakeRuntime.cs
```diff
@@ -17,15 +17,62 @@ public sealed class CakeRuntime : ICakeRuntime
...
+        /// Gets a value indicating whether the current execution is running in debug mode.
+        /// </summary>
+        public bool IsDebug { get; private set; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="CakeRuntime"/> class.
         /// </summary>
-        public CakeRuntime()
+        /// <param name="arguments">The arguments.</param>
+        public CakeRuntime(ICakeArguments arguments)
```
mholo65: This does not work as ICakeArguments is resolved much later on... One solution could be to wrap it around Lazy<T> or Func<T> which is both supported by Autofac for these "chicken-egg" issues..

Another way could be to extract some of the Core classes to a separate module which could be built after the Arguments module..

Thought? @gep13 @patriksvensson

gep13:
@mholo65 just wanted to touch base on this one again...

Where are we with it?  Thanks!

mholo65:
@gep13 as mentioned above, this is a "chiclen-egg" issue with some registrations in Autofac. Need to think of an solution which doesn't require major refactoring.

mholo65:
@patriksvensson 

It's not possible to inject **ICakeArguments** to **CakeRuntime**... For that to work, we'll need to register the **ArgumentsModule** prior to **CoreModule**. Don't even know if that's possible, haven't dug that deep..

Should we resort to implementation as  presented [here](https://github.com/mholo65/cake/blob/2c873ac7845531455646801fbce3f9e83efd6d74/src/Cake.Core/CakeEnvironment.cs#L71)

mholo65:
@patriksvensson ok! I'll fix it. In the meantime, you might want to move #902 and #876 to v0.18.0 milestone

gep13:
>@mholo65 said...
you might want to move #902 and #876 to v0.18.0 milestone

I have cleared those issues from having a milestone assigned.  We can assign them once we know when they are going to be ready.

mholo65:
@patriksvensson still want this and #965?

I could rebase and fix, otherwise close.

mholo65:
Closing. No need for this.

PR #1075, opened 2016-07-18
Reverted build.sh from changes in commit eb1c4e9

mholo65:
build.sh is broken, can be seen here https://travis-ci.org/cake-build/cake/jobs/145653224

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -26,7 +26,7 @@ SCRIPT="build.cake"
 TARGET="Travis"
 CONFIGURATION="Release"
 VERBOSITY="verbose"
-DRYRUN=false
+DRYRUN=
 SHOW_VERSION=false
 SCRIPT_ARGUMENTS=()
 


gep13:
Relates to #1077

gep13:
@mholo65 thank you for doing this.  I went through the process of rationalising the various bootstrappers that we have across the cake-build organisation last night, and this was contained within one of them, so I thought I should also add it here.  I didn't realise that it would have this result.

mholo65:
@gep13 np, but there is still something fishy with Linux / osx builds... Travis seems to use v0.13.0 of Cake while building + some unit tests fail. Bitrise, I can't acces...

gep13:
Yip, I "think" I have just fixed the problems with the Unit Tests.

gep13:
@mholo65 yip, Travis CI just passed, and will now wait to see if Bitrise does as well.

gep13:
@mholo65 and now the Bitrise CI is passing again.  Thanks again for all your help!

Issue #1078, opened 2016-07-19
PowerShell bootstrapper MD5 check of packages config not supported by earlier versions of PowerShell


devlead:
See cake-build/resources#8

gep13:
Also the website, cake-vso and example repos :-)

devlead:
@gep13 :+1: :wink:

PR #1079, opened 2016-07-19
Fix so PS bootstrapper MD5 check works on PS V2

devlead:
This PR addresses issue #1078 by replacing the missing cmdlet with corresponding .Net classes.

Tested with `PowerShell -version 2.0 -NoProfile .\build.ps1 -Verbose` which failed previous this fix and now works.


gep13:
LGTM! :+1:

Issue #1089, opened 2016-07-20
Add ITeardownContext to Teardown method


gep13:
With information about overall success of the build.

patriksvensson:
Closed via #1090

PR #1090, opened 2016-07-20
GH1089: Add additional context information to Teardown

gep13:
Relates to #1089

### DIFF
--- a/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
@@ -683,7 +683,7 @@ namespace Cake.Core.Tests.Unit
                 var result = new List<string>();
                 var fixture = new CakeEngineFixture();
                 var engine = fixture.CreateEngine();
-                engine.RegisterTaskSetupAction((cc, sc) => result.Add("TASK_SETUP:" + sc.Task.Name));
+                engine.RegisterTaskSetupAction(context => result.Add("TASK_SETUP:" + context.Task.Name));
                 engine.RegisterTask("A").Does(() => result.Add("Executing A"));
                 engine.RegisterTask("B").Does(() => result.Add("Executing B")).IsDependentOn("A");
 
@@ -701,7 +701,7 @@ namespace Cake.Core.Tests.Unit
                 var result = new List<string>();
                 var fixture = new CakeEngineFixture();
                 var engine = fixture.CreateEngine();
-                engine.RegisterTaskSetupAction((cc, sc) => { throw new Exception("fake exception"); });
+                engine.RegisterTaskSetupAction(context => { throw new Exception("fake exception"); });
                 engine.RegisterTask("A").Does(() => result.Add("Executing A"));
                 engine.RegisterTask("B").Does(() => result.Add("Executing B")).IsDependentOn("A");
 
@@ -722,8 +722,8 @@ namespace Cake.Core.Tests.Unit
                 var result = new List<string>();
                 var fixture = new CakeEngineFixture();
                 var engine = fixture.CreateEngine();
-                engine.RegisterTaskSetupAction((cc, sc) => result.Add("TASK_SETUP:" + sc.Task.Name));
-                engine.RegisterTaskTeardownAction((cc, sc) => result.Add("TASK_TEARDOWN:" + sc.Task.Name));
+                engine.RegisterTaskSetupAction(context => result.Add("TASK_SETUP:" + context.Task.Name));
+                engine.RegisterTaskTeardownAction(context => result.Add("TASK_TEARDOWN:" + context.Task.Name));
                 engine.RegisterTask("A").Does(() => result.Add("Executing A"));
                 engine.RegisterTask("B").Does(() => result.Add("Executing B")).IsDependentOn("A");
 
@@ -741,8 +741,8 @@ namespace Cake.Core.Tests.Unit
                 var result = new List<string>();
                 var fixture = new CakeEngineFixture();
                 var engine = fixture.CreateEngine();
-                engine.RegisterTaskSetupAction((cc, sc) => result.Add("TASK_SETUP:" + sc.Task.Name));
-                engine.RegisterTaskTeardownAction((cc, sc) => result.Add("TASK_TEARDOWN:" + sc.Task.Name));
+                engine.RegisterTaskSetupAction(context => result.Add("TASK_SETUP:" + context.Task.Name));
+                engine.RegisterTaskTeardownAction(context => result.Add("TASK_TEARDOWN:" + context.Task.Name));
                 engine.RegisterTask("A").Does(() => result.Add("Executing A"));
                 engine.RegisterTask("B").Does(() => result.Add("Executing B")).WithCriteria(() => false).IsDependentOn("A");
                 engine.RegisterTask("C").Does(() => result.Add("Executing C")).IsDependentOn("B");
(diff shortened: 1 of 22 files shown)


gep13:
@cake-build/team Apologies for creating a branch on this repository.  I was already completed the work locally before I noticed :cry:

gep13:
@cake-build/team what are your thoughts on this?

This would allow the following:

```
Teardown((context, build) =>
{
    if(build.Successful)
    {
        Information("Build successful");
    }
    else
    {
        Information("Build failed");
    }
});
```

Tested locally using Cake build as test, and seems to work as I would expect it to:

image

I basically threw an exception in the package task for testing it out.

patriksvensson:
@gep13 If `ITearDownContext` would inherit `ICakeContext` then we would be backwards compatible without adding a new overload for `Teardown`.

gep13:
@patriksvensson I did think about that, but then I thought it would be "different" from others in the same file, so thought consistency was better.  Unless we also go through and change the others as well?

gep13:
@patriksvensson or, should we do this one in the "new" way, and then create another issue to go back and change the existing ones?

patriksvensson:
Maybe a good idea to change them all. What do you think?

gep13:
I can try to take a stab at that :-)

devlead:
An idea, maybe
https://github.com/cake-build/cake/blob/1f2c8b9d677cdc55df33c6d9c3c036fa056c3c45/src/Cake.Core/CakeEngine.cs#L149-L157
```csharp
            catch
            {
                exceptionWasThrown = true;
                throw;
            }
            finally
            {
                PerformTeardown(strategy, context, exceptionWasThrown);
            }
```
should be something like
```csharp
            catch(Exception ex)
            {
                exceptionWasThrown = true;
                thrownException = ex;
                throw;
            }
            finally
            {
                PerformTeardown(strategy, context, exceptionWasThrown, ex);
            }
```
so we could report exception back to on teardown context.

Thoughts?

gep13:
>@devlead said...
Should be added to ITeardownContext interface too right?

Yip, last night's commit was an interim one :-D

src/Cake.Core.Tests/Unit/CakeEngineTests.cs
```diff
@@ -683,7 +683,7 @@ public void Should_Run_Task_Setup_Before_Each_Task()
                 var result = new List<string>();
                 var fixture = new CakeEngineFixture();
                 var engine = fixture.CreateEngine();
-                engine.RegisterTaskSetupAction((cc, sc) => result.Add("TASK_SETUP:" + sc.Task.Name));
+                engine.RegisterTaskSetupAction(context => result.Add("TASK_SETUP:" + ((ITaskSetupContext)context).Task.Name));
```
patriksvensson: Should we have to cast the context here?

src/Cake.Core.Tests/Unit/TaskSetupContextTests.cs
```diff
@@ -13,8 +14,11 @@ public sealed class TheConstructor
             [Fact]
             public void Should_Throw_If_Task_Is_Null()
```
patriksvensson: Maybe we should add a test for checking that it throws if the context is null as well.

src/Cake.Core.Tests/Unit/CakeEngineTests.cs
gep13: No, not any more. Originally I did, then I changed it based on your feedback. Will correct.

src/Cake.Core/TeardownContext.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TeardownContext"/> class.
+        /// </summary>
+        /// <param name="cakeContext">The Cake Context</param>
+        /// <param name="successful">If set to <c>true</c>, the build was not successful.</param>
+        /// <param name="throwException">The exception that was thrown by the target.</param>
+        public TeardownContext(ICakeContext cakeContext, bool successful, Exception throwException)
+            : base(cakeContext)
+        {
+            _successful = successful;
```
patriksvensson: Would a successful build be the same as a build with no exceptions? In that case, maybe we can just check `thrownException` for null instead of providing success. Or do we want this to be explicit? Either one is fine by me, just asking the question loud :smile:

src/Cake.Core.Tests/Unit/TaskTeardownContextTests.cs
```diff
@@ -10,10 +12,23 @@ public sealed class TaskTeardownContextTests
         public sealed class TheConstructor
         {
             [Fact]
-            public void Should_Throw_If_Task_Is_Null()
+            public void Should_Throw_If_Context_Is_Null()
             {
                 // Given, When
-                var result = Record.Exception(() => new TaskSetupContext(null));
+                var result = Record.Exception(() => new TaskSetupContext(null, null));
```
patriksvensson: Should provide a task here.

src/Cake.Core.Tests/Unit/TeardownContextTests.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public sealed class TeardownContextTests
+    {
+        public sealed class TheConstructor
+        {
+            [Fact]
+            public void Should_Throw_If_Context_Is_Null()
+            {
+                // Given, When
+                var result = Record.Exception(() => new TeardownContext(null, null));
```
patriksvensson: Should provide a task here.

src/Cake.Core/CakeContextAdapter.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// </summary>
+    public abstract class CakeContextAdapter
+    {
+        private readonly ICakeContext _cakeContext;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CakeContextAdapter"/> class.
+        /// </summary>
+        /// <param name="cakeContext">The Cake Context</param>
+        protected CakeContextAdapter(ICakeContext cakeContext)
```
patriksvensson: Can we rename this to `context`?

src/Cake.Core/TaskSetupContext.cs
```diff
@@ -8,15 +8,17 @@ namespace Cake.Core
...
     {
         private readonly ICakeTaskInfo _task;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TaskSetupContext"/> class.
         /// </summary>
+        /// <param name="cakeContext">The Cake Context.</param>
         /// <param name="task">The task.</param>
-        public TaskSetupContext(ICakeTaskInfo task)
+        public TaskSetupContext(ICakeContext cakeContext, ICakeTaskInfo task)
```
patriksvensson: Can we rename this to `context`?

src/Cake.Core/TaskTeardownContext.cs
```diff
@@ -17,10 +17,12 @@ public sealed class TaskTeardownContext : ITaskTeardownContext
         /// <summary>
         /// Initializes a new instance of the <see cref="TaskTeardownContext"/> class.
         /// </summary>
+        /// <param name="cakeContext">The Cake Context.</param>
         /// <param name="task">The task.</param>
         /// <param name="duration">The duration of the task.</param>
         /// <param name="skipped">if set to <c>true</c>, the task was not executed.</param>
-        public TaskTeardownContext(ICakeTaskInfo task, TimeSpan duration, bool skipped)
+        public TaskTeardownContext(ICakeContext cakeContext, ICakeTaskInfo task, TimeSpan duration, bool skipped)
```
patriksvensson: Can we rename this to `context`?

src/Cake.Core/TeardownContext.cs
```diff
@@ -0,0 +1,48 @@
...
+    public sealed class TeardownContext : CakeContextAdapter, ITeardownContext
+    {
+        private readonly Exception _thrownException;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TeardownContext"/> class.
+        /// </summary>
+        /// <param name="cakeContext">The Cake Context</param>
+        /// <param name="throwException">The exception that was thrown by the target.</param>
+        public TeardownContext(ICakeContext cakeContext, Exception throwException)
```
patriksvensson: Can we rename this to `context`?

gep13:
@patriksvensson I think I have addressed all your comments :smile:

patriksvensson:
@gep13 Looks good to me! Do you feel ready with this? Should I merge it?

PR #1099, opened 2016-07-22
GH1098: Updates AppVeyor UploadTestResults

### LINKED ISSUE #1097: AppVeyor.UploadTestResults failing silently
### What You Are Seeing?
When uploading test results using the AppVeyor.UploadTestResults method, no test results appear in the appveyor build, but there is no error produced.

After some debugging, it appears the appveyor api is returning 415 Unsupported Media Type

### What is Expected?
The tests should upload correctly, or some sort of error should be produced

0.14.0.0

x64

Windows

AppVeyor

Cake build with NUnit tests, then upload to AppVeyor

```
Task("Run-Tests")
    .IsDependentOn("Build")
    .Does(() =>
{  
    NUnit(testAssemblies);

    if(AppVeyor.IsRunningOnAppVeyor)
    {
        AppVeyor.UploadTestResults(testOutput, AppVeyorTestResultsType.NUnit);
    }
});
```

### Output Log
There is no output produced on failure

### Possible Resolution

The appveyor example for this uses powershell: https://www.appveyor.com/docs/running-tests#uploading-xml-test-results

I have tried adding content-type, changing to multipart content upload and a few other changes, but nothing worked.

However, the powershell example uses System.Net.WebClient rather than System.Net.Http.HttpClient.  Changing to this worked for me.  I can submit a pull request with this change, but wondered if there were any other factors that I hadn't considered?

Philo:
The updates changes from using HttpClient to WebClient.

The WebClient.UploadFile packages and sends a payload in the format expected by AppVeyor.  WebClient is also supported on Mono (https://searchcode.com/codesearch/view/35714320/).

Resolves #1097

### DIFF
--- a/src/Cake.Common/Build/AppVeyor/AppVeyorProvider.cs
+++ b/src/Cake.Common/Build/AppVeyor/AppVeyorProvider.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Globalization;
 using System.IO;
+using System.Net;
 using System.Net.Http;
 using Cake.Common.Build.AppVeyor.Data;
+using Cake.Common.Net;
 using Cake.Core;
 using Cake.Core.IO;
 
@@ -117,12 +119,11 @@ namespace Cake.Common.Build.AppVeyor
                 throw new CakeException("Failed to get AppVeyor API url.");
             }
 
-            var url = string.Format(CultureInfo.InvariantCulture, "{0}/api/testresults/{1}/{2}", baseUri, resultsType, Environment.JobId);
+            var url = new Uri(string.Format(CultureInfo.InvariantCulture, "{0}/api/testresults/{1}/{2}", baseUri, resultsType, Environment.JobId));
 
-            using (var stream = File.OpenRead(path.FullPath))
             using (var client = new HttpClient())
             {
-                client.PostAsync(url, new StreamContent(stream)).Wait();
+                client.UploadFileAsync(url, path.FullPath).Wait();
             }
         }
 
--- a/src/Cake.Common/Net/HttpAliases.cs
+++ b/src/Cake.Common/Net/HttpAliases.cs
@@ -130,5 +130,119 @@ namespace Cake.Common.Net
 
             context.Log.Verbose("Download complete, saved to: {0}", outputPath.FullPath);
         }
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


gep13:
@cake-build/team what are your thoughts on this one?  Looks good to me, just wanted to check with you both before merging. Thanks!

devlead:
What's the reason for removing HttpClient?  Webaclient would break CoreCLR compatibly.

gep13:
```
Possible Resolution

The appveyor example for this uses powershell: https://www.appveyor.com/docs/running-tests#uploading-xml-test-results

I have tried adding content-type, changing to multipart content upload and a few other changes, but nothing worked.

However, the powershell example uses System.Net.WebClient rather than System.Net.Http.HttpClient. Changing to this worked for me. I can submit a pull request with this change, but wondered if there were any other factors that I hadn't considered?
```

gep13:
@devlead check the linked issue.

Philo:
The AppVeyor result need to be uploaded as multipart form-data and webclient had support for this via UploadFile, to support coreclr I guess the UploadFile code would need to be copied and adapted to work assuming there isn't a coreclr equivalent.

devlead:
@Philo you could try something like
```csharp
            using (var client = new HttpClient())
            {
                using (var content = new MultipartFormDataContent(string.Format(CultureInfo.InvariantCulture, "---------------------{0:X}", DateTime.UtcNow.Ticks)))
                {
                    using (var stream = _fileSystem.GetFile(path).OpenRead())
                    {
                        var streamContent = new StreamContent(stream) {
                            Headers = {ContentType = MediaTypeHeaderValue.Parse("application/octet-stream")}
                        };
                        content.Add(streamContent, "file", path.GetFilename().ToString());
                        client.PostAsync(url, content).Wait();
                    }
                }
            }
```

you would have to add an `FileSystem _fileSystem` to provider, if it works we should probably add an `UploadFile` alias to [HttpAliases](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Net/HttpAliases.cs) so it could be reused as this felt harder than it should be :wink:

patriksvensson:
@Philo @devlead Yeah, we can't change to WebClient since that would break CoreClr compatibility.

gep13:
LGTM! Thanks for fixing this this up!

PR #1120, opened 2016-07-25
GH-1119: Add CakeExecuteScript parent cake.exe ToolPath fallback

### LINKED ISSUE #1119: CakeExecuteScript tool resolution fails if script in parent path
If i run `.\src\Cake\bin\Release\Cake.exe \temp\test\pipelinesenv.cake` tool resolution will fail not finding cake.exe
`Error: Cake.Core.CakeException: Cake: Could not locate executable.`

I propose we fallback to the cake.exe path of the process running the parent script, seems reasonable to want to use same version of Cake unless ToolPath explicitly specified, to not break anything I'll keep hombrew unix logic and fallback to entry assembly.

devlead:
* Fixes #1119


gep13:
LGTM! Rebased and merged here: https://github.com/cake-build/cake/commit/e9ae9e927dcddd245fe549c613e6641d56726d9e

PR #1124, opened 2016-07-26
GH1123: Made TargetFramework constant.

### LINKED ISSUE #1123: Set ICakeRuntime.TargetFramework to constant
We only target a specific framework so probing the AppDomain is not necessary.

patriksvensson:
Should preserve compatibility with older Mono versions.

Closes #1123


devlead:
LGTM 👍 merged.

PR #1137, opened 2016-08-08
[HOLD] GH1111: Fixed DotNetCoreRestore: dotnet restore no longer supports globbing

### LINKED ISSUE #1111: DotNetCoreRestore: dotnet restore no longer supports globbing
As pointed [here](https://gitter.im/cake-build/cake?at=5793a37d67c448730eee57ef) by @mmisztal1980, `dotnet restore` no longer supports globbing on the `root` parameter. The XML docs on `DotNetCoreRestore` are OK, but the example shows a string with a globbing pattern. I also came across this problem and an issue wasn't created yyet.

@patriksvensson suggested that the `root` parameter on `DotNetCoreRestore` could become a `DirectoryPath`. However, on `dotnet restore` it actually can be a list of project.json, global.json or root directory.
- Is there any type on cake that can represent both files and directory paths?
- Maybe we could have a `params` argument?
- Or maybe we can assume the common use case, which I think will be a root directory for search, as Patrick suggested.

RLittlesII:
This resolves #1111 and #1204

@patriksvensson based on the comments you seemed to think that either a `DirectoryPath` or a list of `project.json` would work.  I have implemented the later, and was checking to make sure this is the appropriate path before I change the `DotNetCoreTest` alias as well.  I have accounted for a string pattern and a single file.  Would the `DirectoryPath` route be more efficient?
- [x] List of `DirectoryPath`
- [x]  List of `FilePath`
- [x]  Specific `FilePath`
- [x]  Specific `DirectoryPath`

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/DotNetCore/Restore/DotNetCoreRestorerFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/DotNetCore/Restore/DotNetCoreRestorerFixture.cs
@@ -2,7 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
+using System.Linq;
 using Cake.Common.Tools.DotNetCore.Restore;
+using Cake.Core.IO;
 using Cake.Testing;
 
 namespace Cake.Common.Tests.Fixtures.Tools.DotNetCore.Restore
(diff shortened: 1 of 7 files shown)


adamhathcock:
It looks like you're allowing a "list of paths" or a "list of directories" but not a mixture.  Would it better to just treat it as a single collection of strings then let the `dotnet restore` command deal with it?

devlead:
@adamhathcock  both inherit from path so that's an option too.

RLittlesII:
@adamhathcock I'd be fine with that.  So you're thinking a single method like:
 `DotNetCoreRestore(this ICakeContext, IEnumerable<string> paths)`.
I lose the ability to force the absolute path, but I can still deal with spaces.

patriksvensson:
It's probably better to have the four aliases you mention and simply accept `IEnumerable<Path>` in the tool itself.

RLittlesII:
Okay, I will get this cleaned up tonight.

adamhathcock:
Also, it appears that `dotnet build` does do globbing but it can accept more than one value.  Should that be rolled into this too?

I feel like all the dotnet commands and accept N paths.

patriksvensson:
Hmmm, maybe we should ignore globbing for now and wait how the CLI tools turn out when they've stabilized a bit.

adamhathcock:
Really all I need is to allow more than one path as input.  Whether it's a glob, file or directory I don't care.  The dotnet command will sort it.

RLittlesII:
@adamhathcock The original issue this PR was meant to resolve is to handle the change to `dotnet restore` not supporting globbing anymore.

@patriksvensson I believe there is value in this change.  We aren't actually handling globbing with the alias, we are just allowing the result of globbing to be accepted by the alias, which I don't believe it can handle in the current implementation.  This happens to cover @adamhathcock needs as well as the original issue #1111.  That said, if you would prefer to hold off, I am open to suggestion on handling these use cases or waiting until the dotnet cli tools are stable.  Just let me know how you would like me to proceed.

gep13:
@RLittlesII just wanted to touch base on this one.

Where are we with it?  Thanks!

RLittlesII:
@gep13 @patriksvensson wanted me to wait until things settled down with the dotnet CLI.  So I am waiting on some direction from the project team.

gep13:
@RLittlesII ok, makes sense.  I have updated the title to reflect this, just so we know where this PR is at a glance.  Feel free to update it again if you start working on it again.

RLittlesII:
@gep13 I am closing this PR because it looks like issue #1533 is updating the DotNetCore tooling.
- #1533 
- #1204 
- #1111

Issue #1138, opened 2016-08-08
Allow custom loggers in the VSTestSettings


Jericho:
VSTestSettings has a property called "Logger" with two possible values: None or Trx.

However, we need to be able to specify custom values. For example, when running on AppVeyor we can specify "/logger:AppVeyor".

As a workaround, I was able to specify the AppVeyor logger using ArgumentCustomization like so:

```
Task("Run-Code-Coverage")
	.IsDependentOn("Build")
	.Does(() =>
{
	var testAssemblyPath = string.Format("./MyLibraryUnitTests/bin/{0}/MyLibrary.UnitTests.dll", configuration);
	var vsTestSettings = new VSTestSettings();
	if (AppVeyor.IsRunningOnAppVeyor) vsTestSettings.ArgumentCustomization = args => args.Append("/logger:Appveyor");

	OpenCover(
		tool => { tool.VSTest(testAssemblyPath, vsTestSettings); },
		new FilePath("./CodeCoverageData/coverage.xml"),
		new OpenCoverSettings()
			.ExcludeByAttribute("*.ExcludeFromCodeCoverage*")
			.WithFilter("+[MyLibrary]*")
			.WithFilter("-[MyLibrary]MyLibrary.Properties.*")
	);
});

```

A few ideas were suggested:

- Change the Logger property to a string and create some nice extension methods to accomplish the same thing. Like .WithAppVeyorLogger() or similar (this would probably break existing scripts though).
- Perhaps a CustomLogger string?

Jericho:
I propose the following:

1. Mark the 'Logger' property as obsolete (I don't know what you policy is regarding obsolete properties but I imagine you want to keep it around for a release or two before removing it).
2. Also mark the VSTestLogger enum as obsolete. It wil no longer be necessary when the 'Logger' property is removed.
3. Add a new string property called 'LoggerName'
4. Add a few extension methods such as:
        .WithTrxLogger()
        .WithAppVeyorLogger()
        .WithCustomLogger(string loggerName)
        .WithoutAnyLogger()
5. As much as possible, keep the obsolete 'Logger' property and the new property in sync (for example: set LoggerName to "trx" if Logger is set to VSTestLogger.Trx; set LoggerName to a blank string if Logger is set to VSTestLogger.None; etc) and throw an exception if, for some reason, the two properties can't be kept in sync.
6. Add a new value called 'Custom' to the VSTestLogger enum. This will make it a little bit easier to keep the two properties in sync

[...]

PR #1143, opened 2016-08-09
GH1138: Custom vstest loggers

Jericho:
Allow custom loggers to be specified when running unit tests with VSTest. This was discussed in gitter on 8/2016 and is also discussed in issue #1138


patriksvensson:
@Jericho I have some thoughts about the new `LoggerName` property.

I don't think it needs to be kept in sync with `Logger` and vice versa, but simply make `LoggerName` take precedence over `Logger`, so if both are specified, `LoggerName` is used. Since `Logger` is obsoleted, both documentation and compiling the script will indicate that it should not be used so I don't think it's necessary with this fallback.

I'm thinking something like this:

```csharp
if(!string.IsNullOrEmpty(LoggerName))
{
  // Use LoggerName
}
else 
{
  if(Logger != null) 
  {
    // Convert Logger to a string representation
  }
}
```

Jericho:
I agree that `LoggerName` should take precedence, that's how I implemented it. As far as where the conversion Logger to/from a string representation should take place, I thought it made more sense to group this logic in the getter and setter of the ~Logger~ property because this logic is only needed as long as we keep the obsolete property around. When we decide to remove the property, the conversion logic needs to also be removed and I think it's make sense to have all the code located in one single file as opposed to be in two separate files.

patriksvensson:
@Jericho Hello again. Sorry for the late reply.

We've been discussing this change in the Cake maintainer team and decided that it's OK for the project to take a breaking change and simply change `Logger` to be a `string`. We don't think this will affect a lot of projects, and the change is trivial to make when updating to latest Cake. We would also keep the names of the properties in the settings class in sync with the actual tool which we want.

Would you want to make the change for this?

Jericho:
Sure no problem, I'll take care of it

Jericho:
yes no problem. Will take care of it late today.

Jericho:
Hopefully I squashed correctly, let me know...

gep13:
>@Jericho said...
Hopefully I squashed correctly, let me know...

So, looks like you still have 29 commits in this branch.  Would you like me to squash this down for you?

Jericho:
Sure

gep13:
@Jericho ah, looks like you got at least part of the way there.  Perhaps all that was missing was a `git push -f`.  Doing a normal push wouldn't work, and the changes would have been rejected.

Regardless though, I have just finished squashing and merging the changes that you made.  You can find them here: https://github.com/cake-build/cake/commit/ac8c58509db1c30388d752af8fe105523440949a

Due to the fact that I have done this, I will need to manually close this PR, however, correct attribution remains in place for the work that you carried out.

Thanks again for your help on this.

Jericho:
Thank you!

Issue #905, opened 2016-05-20
Add command line option to wait for user confirmation on failure (similar to batch: PAUSE)


Sam13:
Add a command line option which prevents Cake from exiting when a failure happened (e.g. in the script or loading a Cake assembly).
When Cake runs successful the process should be exited.

This enables creating Windows shortcuts to Cake scripts which the user can read the error message if a failure happens.

devlead:
Closed due to inactivity.

PR #1148, opened 2016-08-10
GH905: Add command line option to wait for user confirmation on failure.

Sam13:
Implemented my feature request #905.
Added `ConfirmError` property to `CakeOptions` and `WaitForInput()` to `IConsole`. In `Program.Main()` routine's exception handler wait for input if error confirmation is enabled.

This is useful when creating Windows shortcuts directly to Cake. Without that feature to console window closes and the user has no idea whether his script was sucessful or not.


Sam13:
Additionally it would be nice to be able to confirm errors which happen before command line parsing (e.g. AssemblyLoad failures). Any idea how this could be implemented?

Sam13:
I suggest to confirm errors when Cake is not running under Console (TBD how to determine this). And maybe remove this command line option again or combine with the running under console circumstance.

What do you think?

Sam13:
Anyone?

patriksvensson:
@Sam13 Sorry for the delay.

We prefer to keep direct user interaction out of Cake itself since it's all about automation, but I don't think that user confirmation on error needs to be added to Cake itself since it could be a simple method in the script's Teardown statement.

```csharp
Teardown((context, info) =>
{
  if(!info.Successful && context.Argument<bool>("prompt", false)) {
    System.Console.WriteLine("An error occured: {0}", info.Exception.Message);
    System.Console.WriteLine("Press any key to quit!");
    System.Console.ReadKey(true);
  }
});

Sam13:
@patriksvensson  Thanks for your answer. I use a similar construct to solve my requirement at the moment. But this is not enough: If an error happens before the script execution is started the error is still not confirmable.

Example: Run Cake from UNC without having .NET security configured correctly -> Console opens and exception loading the Cake.Core assembly, this means the script which the confirmation code resides cannot be executed. So no confirmation.

Sam13:
Will lookup a solution outside Cake if this behavior is not desirable within Cake.

devlead:
@Sam13 actually personally think the best option would be to create an "Cake Launcher" exe that you associate your .cake files with in windows.
Which basically just does a process start checks the exit code and pauses if non zero and exits with same code as Cake, that way you can handle all errors that might occur while launching Cake (even module exceptions etc.).

PR #1181, opened 2016-08-21
GH1180: Refactors container registration

### LINKED ISSUE #1180: Change API for registering dependencies with Cake
Currently there is some problems with `ICakeContainerRegistry` and `ICakeRegistrationBuilder`.

* Methods in both are generic, which makes it cumbersome to register types. This can be solved by changing methods to accept `Type` and add extension methods on top.
* `ICakeContainerRegistry` should be called `ICakeContainerRegistrar` since it's something that registers things with a container and not a registry over containers.
* The `ICakeContainerRegistry.RegisterModule` is out of place and should be removed. Could be added as an extension method.

patriksvensson:
Also adds a version checks for Cake.Core.dll when loading assemblies which would give consumers of incompatible modules a heads up what is going on.

Closes #1180


src/Cake.Core/Extensions/ContainerExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+// ReSharper disable once CheckNamespace
+namespace Cake.Core.Composition
+{
+    /// <summary>
+    /// Contains extension methods for <see cref="ICakeContainerRegistrar"/>.
+    /// </summary>
+    public static class ContainerRegistrarExtensions
+    {
+        /// <summary>
+        /// Registers a type with the container registry.
```
patriksvensson: Should be `registrar`.

Issue #856, opened 2016-04-30
Report test coverage for Cake to Coveralls.io


gep13:
It would be good to be able to report on/view the test coverage within Cake.  We should add this into the build.cake file, and report it to Coveralls.io.

Depends on https://github.com/cake-build/cake/issues/464

gep13:
This will be possible very soon, after this is published:

https://github.com/gep13/Cake.Coveralls

gep13:
Looks like there is an issue pushing the report.

https://ci.appveyor.com/project/cakebuild/cake/build/0.16.0-alpha.55.build.2654#L1382

Will investigate.

devlead:
Doesn't seem to be working well on our Jenkins
```
========================================
Run-Unit-Tests
========================================
Executing task: Run-Unit-Tests
At: Program.Main
An System.ComponentModel.Win32Exception occured: Access is denied
stack:    at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
   at OpenCover.Framework.ProfilerRegistration.ExecuteProcess(ProcessStartInfo psi) in c:\projects\opencover\main\OpenCover.Framework\ProfilerRegistration.cs:line 75
   at OpenCover.Framework.ProfilerRegistration.ExecuteRegsvr32(Boolean userRegistration, Boolean register, Boolean is64) in c:\projects\opencover\main\OpenCover.Framework\ProfilerRegistration.cs:line 97
   at OpenCover.Console.Program.RunWithContainer(CommandLineParser parser, Bootstrapper container, IPersistance persistance) in c:\projects\opencover\main\OpenCover.Console\Program.cs:line 178
   at OpenCover.Console.Program.Main(String[] args) in c:\projects\opencover\main\OpenCover.Console\Program.cs:line 70
A report has been sent to the OpenCover development team.
If you are unable to resolve the issue please contact the OpenCover development team
see https://www.github.com/opencover/opencover/issues
An error occured when executing task 'Run-Unit-Tests'.
Error: OpenCover: Process returned an error (exit code 1).
Build step 'Windows PowerShell' marked build as failure
Finished: FAILURE
```
http://cake-jenkins.azurewebsites.net/job/Cake/223/console

gep13:
Now that the Jenkins build is up and running again, closing to close this off as complete

PR #1182, opened 2016-08-23
GH856: Added pushing Code Coverage results

gep13:
Relates to #856


gep13:
@cake-build/team I am not convinced that this is doing what I think it is, but I thought I would put it out there for review.

Thoughts?

patriksvensson:
I think it looks good. Van't see any obvious problems with this. :+1:

patriksvensson:
Can't* 

Can't edit comments on the phone...

gep13:
@patriksvensson the main problem that I see just now is that the generated report only reports an 8% code coverage, which i really don't think is right:

image

I think there is an issue with running OpenCover with dotnet test, or I am simply doing something wrong.  I will need to dig into it some more.

gep13:
@cake-build/team ok, I have just updated this PR.  The end result of the build is a code coverage report that looks like this:

image

I am waiting on the report at coveralls.io to be processed, but I suspect that it will show a similar story.

Thoughts?

devlead:
Interesting, think Cake.Testing/Xunit should be excluded though.

gep13:
>@devlead said...
Interesting, think Cake.Testing/Xunit should be excluded though.

To be clear, are you saying exclude both `Cake.Testing` and `Cake.Testing.XUnit`?  @patriksvensson you agree?

patriksvensson:
@Ep13 Yes, I agree with @devlead

patriksvensson:
Sorry, that was supposed to be to @gep13 :smile:

devlead:
@gep13 @patriksvensson agree.

gep13:
Here is the updated report:

image

Rebasing just now...

Issue #702, opened 2016-02-22
Implement custom logger support for MSBuild


tlycken:
MSBuild has the option of registering custom loggers, via the `/logger:foo` switch. This can be utilized through something like this today (this is in use in our build script):

```
MSBuild(solutionPath, settings => 
{
    settings.Configuration = configuration;
    
    if (EnvironmentVariable("BUILD_DEFINITIONNAME") != null) // this only happens on TFS
    {
        settings.ArgumentCustomization = arguments =>
        {
            arguments.Append(string.Format(
                "/logger:{0}\\Agent\\Worker\\Microsoft.TeamFoundation.DistributedTask.MSBuild.Logger.dll",
                EnvironmentVariable("AGENT_HOMEDIRECTORY")));
            return arguments;
        };
    }
});
```

This is required to get "nice" output on TFS builds, so the web UI shows a list of errors rather than just "MSBuild failed" (see [this SO question](http://stackoverflow.com/questions/35548499) for an example).

It would be nice if this was available in a less roundabout way, e.g. through a property directly on `MSBuildToolSettings`, maybe something like this:

```
MSBuild(solutionPath, settings =>
{
    settings.Configuration = configuration;
    settings.Loggers.Add(/* long path to dll omitted */);
}
```

cpx:
I'm interested in fixing this if no one else is working on it yet.

The idea I have is basically the same as for adding/setting properties. MSBuildSettings would be extended with a Loggers property. Adding to this list would be done via an extension method WithLogger, taking the assembly name and optionally the class name or parameters for the logger. So the usage would be like this:

```csharp
MSBuild(solutionPath, settings =>
{
    settings
        .WithLogger("MyLoggerAssembly1.dll", "MyLoggerClass1", "SomeParams")
        .WithLogger("MyLoggerAssembly2.dll", "MyLoggerClass2", "SomeParams");
});
```

Issue #473, opened 2015-10-28
Support MSBuild logger switches


jrnail23:
See "Switches for Loggers" section of [MSBuild Command-Line Reference](https://msdn.microsoft.com/en-us/library/ms164311.aspx)

gep13:
Big thumbs up on this one! This is something that I have been thinking about, as I have a need to create a custom MSBuild Logger for something I am thinking about, so would be good to get OOTB support here! :+1:

patriksvensson:
Sounds like a good addition to me! :+1:

PR #1198, opened 2016-08-29
GH702 GH473: Added logger switch to MSBuild runner

cpx:
Relates to #702 and #473.

The gist of it is fairly simple. I've added an extension method for adding a logger switch to the MSBuildSettings class, similar to how property switches are added, like so:
```csharp
public static MSBuildSettings WithLogger(this MSBuildSettings settings, string loggerAssembly, string loggerClass = null, string loggerParameters = null)
```

Calling that method should append an argument to MSBuild, like so:
`/logger:loggerClass,loggerAssembly;loggerParameters`

In addition to this, I added a MSBuildLogger class to hold the above assembly, class and parameters values, and to the MSBuildSettings class I added a collection property to hold instances of those.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsExtensionsTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsExtensionsTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Linq;
 using Cake.Common.Tools.MSBuild;
 using Cake.Core.Diagnostics;
 using Xunit;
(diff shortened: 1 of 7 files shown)


patriksvensson:
@cpx Looks good to me! :+1: Great job! I left some (minor) comments for you.

cpx:
@patriksvensson Done!

src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsExtensionsTests.cs
```diff
@@ -261,5 +261,41 @@ public void Should_Return_The_Same_Configuration()
...
+                // Given
+                var settings = new MSBuildSettings();
+
+                // When
+                settings.WithLogger("LoggerAssembly1", "LoggerClass1", "LoggerParameters1");
+                settings.WithLogger("LoggerAssembly2", "LoggerClass2", "LoggerParameters2");
+
+                // Then
+                Assert.Equal(2, settings.Loggers.Count);
+                Assert.Equal("LoggerAssembly1", settings.Loggers[0].Assembly);
```
devlead: ICollection doesn't support indexing, LINQ could be one option istead
I.e.
```csharp
var first = settings.Loggers.First();
var second = settings.Loggers.Skip(1).First();
```
and use those variables for the asserts.

cpx:
Oops, was a bit quick to push there :) Now it should work.

devlead:
@cpx excellent, could you squash into 1 commit and rebase against latest develop, and ping it's when done.

cpx:
@devlead Done.

cpx:
Hmm, any idea why the OSX build failed? Is it due to changes in this PR or something else?

patriksvensson:
@cpx Merged :+1: Thanks for this contribution! :smile:

Issue #1184, opened 2016-08-24
Add include & exlude parameters to #tool directive


devlead:
Currently `#tool` directive exclusively looks for `*.exe`, this locks out a few scenarios including _CoreCLR_ as command line assemblies there have `dll` extension.

Proposed fix for this is to add 2 new optional parameters to the `#tool` directive that lets you specify your own globbing pattern to override the default behavior.

| parameter | description | example |
|-------------|------------- |----------|
| include      | pattern of which files to look for | `#tool #tool nuget:?package=Cake.Foo&include=./**/Foo.dll`|
| exclude     | pattern of which files not considered tools  | `#tool #tool nuget:?package=Cake.Foo&exclude=./**/*Serializer*.dll`|

You should also be able to specify multiple globbing patterns i.e.
`#tool #tool nuget:?package=Cake.Foo&include=./**/*.dll&include=./**/*.exe&exclude=./**/*Serializer*.dll&exclude=./**/Foo.dll`

patriksvensson:
@gep13 @devlead Anyone started with this yet or can I implement this?

devlead:
@patriksvensson no

PR #1202, opened 2016-08-30
GH1184: Added support for manually including/excluding files for tools

patriksvensson:
- [x] Query string should support multiple parameters with same name
- [x] Add include/exclude parameter
- [x] Better error messages
- [x] Tests

### DIFF
--- a/src/Cake.Core/Extensions/UriExtensions.cs
+++ b/src/Cake.Core/Extensions/UriExtensions.cs
@@ -4,15 +4,16 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 // ReSharper disable once CheckNamespace
 namespace Cake.Core
 {
     internal static class UriExtensions
     {
-        public static IDictionary<string, string> GetQueryString(this Uri uri)
+        public static IReadOnlyDictionary<string, IReadOnlyList<string>> GetQueryString(this Uri uri)
         {
-            var result = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            var result = new Dictionary<string, List<string>>(StringComparer.OrdinalIgnoreCase);
             if (uri != null)
             {
                 var query = uri.Query.TrimStart('?');
@@ -20,20 +21,29 @@ namespace Cake.Core
                 foreach (var part in parts)
                 {
                     var pair = part.Split(new[] { '=' }, StringSplitOptions.None);
-                    switch (pair.Length)
+
+                    if (pair.Length == 1)
                     {
-                        case 1:
-                            result[pair[0]] = string.Empty;
-                            break;
-                        case 2:
-                            result[pair[0]] = pair[1];
-                            break;
-                        default:
-                            throw new CakeException("Could not parse query string.");
+                        if (!result.ContainsKey(pair[0]))
+                        {
+                            result[pair[0]] = new List<string>();
+                        }
+                    }
+                    else if (pair.Length == 2)
+                    {
+                        if (!result.ContainsKey(pair[0]))
+                        {
+                            result[pair[0]] = new List<string>();
+                        }
+                        result[pair[0]].Add(pair[1]);
+                    }
+                    else
+                    {
+                        throw new CakeException("Could not parse query string.");
                     }
                 }
             }
-            return result;
+            return result.ToDictionary(kv => kv.Key, kv => kv.Value as IReadOnlyList<string>);
         }
     }
 }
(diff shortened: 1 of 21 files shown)


src/Cake.NuGet/NuGetContentResolver.cs
```diff
@@ -31,19 +40,56 @@ public IReadOnlyCollection<IFile> GetFiles(DirectoryPath path, PackageType type)
...
+            {
+                collection.Add(_globber.GetFiles(pattern));
+            }
+
+            // Include files.
+            if (package.Parameters.ContainsKey("include"))
+            {
+                foreach (var include in package.Parameters["include"])
+                {
+                    collection.Add(_globber.GetFiles(include));
```
patriksvensson: Add `path` to include pattern.

src/Cake.NuGet/NuGetContentResolver.cs
```diff
@@ -31,19 +40,56 @@ public IReadOnlyCollection<IFile> GetFiles(DirectoryPath path, PackageType type)
...
+                }
+            }
+
+            // Exclude files.
+            if (package.Parameters.ContainsKey("exclude"))
+            {
+                foreach (var exclude in package.Parameters["exclude"])
+                {
+                    collection.Remove(_globber.GetFiles(exclude));
```
patriksvensson: Add `path` to exclude pattern.

src/Cake.Core/Extensions/UriExtensions.cs
```diff
@@ -4,36 +4,46 @@
...
             if (uri != null)
             {
                 var query = uri.Query.TrimStart('?');
                 var parts = query.Split(new[] { '&' }, StringSplitOptions.RemoveEmptyEntries);
                 foreach (var part in parts)
                 {
                     var pair = part.Split(new[] { '=' }, StringSplitOptions.None);
-                    switch (pair.Length)
+
+                    if (pair.Length == 1)
```
patriksvensson: This is ugly... Fix this.

patriksvensson:
@cake-build/team Ok, I *think* this is ready for review.

gep13:
Relates to #1184

gep13:
@patriksvensson are you in a position to rebase this?  And I can try to take a look tonight?

patriksvensson:
@gep13 Sure, I will do this during the day.

patriksvensson:
@gep13 Rebased

src/Cake.NuGet.Tests/Unit/NuGetContentResolverTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                // Then
+                Assert.Equal(1, files.Count);
+                Assert.Equal("/Working/tools/Foo/Foo.dll", files.First().Path.FullPath);
+            }
+
+            [Theory]
+            [InlineData("nuget:?package=Foo&include=/**/*.XML")]
+            [InlineData("nuget:?package=Foo&include=**/*.XML")]
+            [InlineData("nuget:?package=Foo&include=./**/*.XML")]
```
gep13: @patriksvensson I am assuming that the include/exclude is not case sensitive, right?

patriksvensson: Good question. Will add some tests :smile:

patriksvensson: The include/exclude is case sensitive due to the globber. How should we proceed?

patriksvensson: We could change the interface for `IGlobber` to take a `PathComparer` and move the existing method to an extension method (that uses the default comparer). A breaking change, but something that would be useful in general. Or we could extend globbing to accept regex expressions, like here: https://en.wikipedia.org/wiki/Glob_(programming)

gep13: @patriksvensson hmm, I think Regex might be over complicating it.  I think I would prefer to add a breaking change, where the user can control the outcome, without having to dig into regex.  @devlead thoughts?

devlead: I'll say ship case sensitive. And if necessary break next release.

gep13: @devlead I would be ok with that as well.

patriksvensson: @devlead @gep13 Ok, sounds good to me :smile:

patriksvensson:
@gep13 Rebased against upstream/develop.

PR #1203, opened 2016-08-30
GH856: Correcting build error on Jenkins

gep13:
Relates to #856

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -125,6 +125,7 @@ Task("Run-Unit-Tests")
                 parameters.Paths.Files.TestCoverageOutputFilePath,
                 new OpenCoverSettings {
                     ReturnTargetCodeOffset = 0,
+                    Register = "user",
                     ArgumentCustomization = args => args.Append("-mergeoutput")
                 }
                 .WithFilter("+[*]* -[xunit.*]* -[*.Tests]* -[Cake.Testing]* -[Cake.Testing.Xunit]* ")
--- a/src/Cake.Common.Tests/project.json
+++ b/src/Cake.Common.Tests/project.json
@@ -1,5 +1,5 @@
 ﻿{
-  "version": "0.16.0-*",
+  "version": "0.15.3-*",
   "buildOptions": {
     "additionalArguments": [
       "/ruleset:../Test.ruleset",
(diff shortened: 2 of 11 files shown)


gep13:
@devlead ok, so that didn't work :-(

build.cake
```diff
@@ -125,6 +125,7 @@ Task("Run-Unit-Tests")
                 parameters.Paths.Files.TestCoverageOutputFilePath,
                 new OpenCoverSettings {
                     ReturnTargetCodeOffset = 0,
+                    Register = "user",
```
devlead: Isn't user default though?
https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/OpenCover/OpenCoverSettings.cs#L74

Think issue could actually be that server is 32 bit and need to specify Path32-

gep13: Ah, so it is :-)

gep13: @devlead what is the server 32 bit? Can't we change that?  Reluctant to change the build script specifically for when running on our Jenkins server.

devlead: @gep13 yes that's possible currently it's an Azure Web Site hosting jenkins, have to go up a pricing tier to allow 64bit.

One way around it would be an env var to just skip open cover where it's not supported.

gep13: @devlead ah, i see :-(

That would be one approach, yes.  However, now that OpenCover wraps the Test Runner, skipping OpenCover would also skip the Unit Testing, which is not ideal.  We would need to structure the task differently to allow running of the tests directly, when that environment variable is not set.

devlead: @gep13 well OpenCover  takes an `Action<ICakeContext> test` if we defined it before OpenCover is called it could be a simple `if` statement if it should be passed to open cover or directly evaluated against context as `test(Context);`.

just a thought :wink:  and 32-bit is just a hunch, could be that Azure Websites are more limited in allowing the kind of dll interop open cover is doing.

patriksvensson: @devlead @gep13 That would be good even if it doesn't solve the Jenkins issue. Right now the test runners aren't working in VS so I'm running all tests via Cake when developing. Running OpenCover adds about 45 seconds to the run for me so being able to disable that would be awesome :smile:

devlead: Agree 👍

Issue #1121, opened 2016-07-25
Added ability to call SetParameter method on TeamCity Provider


gep13:
As documented here:

https://confluence.jetbrains.com/display/TCD9/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-AddingorChangingaBuildParameter

gep13:
@kcamp thanks again!

PR #1114, opened 2016-07-25
Add support for 'setParameter' service message to TeamCityProvider + …

kcamp:
In working with a legacy build process (converting to Cake), I noticed the absence of a method exposing the setParameter service message in TeamCity.

see: https://confluence.jetbrains.com/display/TCD9/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-AddingorChangingaBuildParameter

Given the current implementation of TeamCityProvider (WriteServiceMessage) this was a simple addition that seems like it would be useful to more fully support TeamCity interop.

PR #1209, opened 2016-09-02
Expose TeamCityProvider.SetParameter method via the ITeamCityProvider interface

dbertram:
Support for TeamCity's "setParameter" service message was added in
https://github.com/cake-build/cake/pull/1114 but only the tests and implementor
were updated.

This change updates the interface to match, which is what is actually exposed to
Cake scripts.

Relates to https://github.com/cake-build/cake/issues/1121.


gep13:
@dbertram this is a great catch! Thank you very much for submitting this!

I will get this merged once the CI's have passed.

dbertram:
@gep13 Thanks for the quick response on this!

Issue #1207, opened 2016-09-01
Add back the script parameter to the bootstrap scripts


adamhathcock:
I was actually using this and it's not in the current master :)

Can I add it back?

adamhathcock:
Wrong script

PR #1211, opened 2016-09-05
Add script parameter back

adamhathcock:
(Not) discussed here :) https://github.com/cake-build/cake/issues/1207

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -4,6 +4,8 @@ This is a Powershell script to bootstrap a Cake build.
 .DESCRIPTION
 This Powershell script will download NuGet if missing, restore NuGet tools (including Cake)
 and execute your Cake build script with the parameters you provide.
+.PARAMETER Script
+The build script to execute.
 .PARAMETER Target
 The build script target to run.
 .PARAMETER Configuration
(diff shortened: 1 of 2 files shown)


kcamp:
I would also like to see this returned to the script as I am making use of the script parameter to separate concerns.  i.e., `build.cake` for build tasks, `integration.cake` to automate running integration suites.
`integration.cake` loads `build.cake` as a dependency, but otherwise keeps the set of tasks for running the integrations [and additional helpers/utilities in a `utils.cake`] out of the context of the main build script.

devlead:
@adamhathcock @kcamp what's the use case when building Cake?

The recommended bootstrapper still has this parameter:
[cake-build/resources/build.ps1#L43](https://github.com/cake-build/resources/blob/develop/build.ps1#L43)

kcamp:
@devlead -- yes.  i misread the comments/situation.  the redistributed version of the script is what i took this to refer to.  my mistake, disregard.

adamhathcock:
I didn't realize the "recommended" one is different from the official one.  Oops.

gep13:
The recommended one has always came from a known URL, as documented here:

http://cakebuild.net/docs/tutorials/setting-up-a-new-project

The one used in the Cake repo is the same as every other bootstrapper in that it is customized as per requirements for the project that it is being used on.

Issue #472, opened 2015-10-28
Support OctoPack


jrnail23:
I'm thinking this would be done via extensions to the existing MSBuild settings & fluent interface DSL, as it just involves adding OctoPack-specific properties to your MSBuild command.

http://docs.octopusdeploy.com/display/OD/Using+OctoPack

RichiCoder1:
Something like:
```C#
Task("Package")
	.Description("Package the website for Octopus")
	.Does(() =>
{
    MSBuild("./Dashboard.sln", settings =>
        settings
            .SetConfiguration(configuration)
            .WithTarget("Rebuild")
            .PublishPackageToHttp("http://my.server.com/api/v2")
            .PublishApiKey("903840932840spsda-sdasdasdas"));
});
```

gep13:
Sounds like a great addition to me :+1:

jrnail23:
@RichiCoder1, yeah, something like that, although there are several options and properties that can be set, so that might merit its own OctoPackSettings object and DSL builders.

patriksvensson:
@jrnail23 @RichiCoder1 Sounds good to me.

So something like this then?

```csharp
Task("Package")
    .Description("Package the website for Octopus")
    .Does(() =>
{
    MSBuild("./Dashboard.sln", settings =>
        settings
            .SetConfiguration(configuration)
            .WithTarget("Build")
            .WithOctoPack(new OctoPackSettings()
                 .SomeOctoPackThing()
                 .SomeOtherOctoPackThing());
});
```

jrnail23:
@patriksvensson, yeah, that's pretty much what I was thinking :smile:

arri-cc:
Is anyone working on this issue?  If not I would like to get it done

gep13:
@arri-cc I don't think anyone in the core team is looking at this yet, so unless @jrnail23 is currently looking at this, I would suggest that it is up for grabs.  Were you thinking about doing this as an Addin for Cake, or as a contribution into Cake directly?

arri-cc:
@gep13 I think targeting an enhancement in either `Cake.Common.Tools.MSBuild` or `Cake.Common.Tools.OctopusDeploy` so that would be direct right?  Are there any rules regarding dependencies between tools?

patriksvensson:
@arri-cc Would prefer if this was solved without involving MSBuild. Preferably this would be a wrapper around `Octo.exe`

RichiCoder1:
[...]

PR #1218, opened 2016-09-10
GH472: Added OctoPack tool

cpx:
See #472. Pretty much implemented in the exact same way as OctoPush.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPackTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPackTests.cs
@@ -1,0 +1,196 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Common.Tests.Fixtures.Tools;
+using Xunit;
+
+namespace Cake.Common.Tests.Unit.Tools.OctopusDeploy
+{
+    public sealed class OctoPackTests
+    {
+        public sealed class ThePackMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Id_Is_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                Assert.IsArgumentNullException(result, "id");
+            }
+
+            [Fact]
+            public void Should_Add_Id_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Version_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.Version = "1.2.3";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --version 1.2.3", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_OutFolder_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.OutFolder = "out";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --outFolder \"/Working/out\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_BasePath_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.BasePath = "base";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --basePath \"/Working/base\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Author_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.Author = "author";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --author \"author\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Title_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.Title = "title";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --title \"title\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Description_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.Description = "description";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --description \"description\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_ReleaseNotes_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.ReleaseNotes = "releasenotes";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --releaseNotes \"releasenotes\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_ReleaseNotesFile_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.ReleaseNotesFile = "releasenotes.md";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --releaseNotesFile \"/Working/releasenotes.md\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Include_To_Arguments_If_Not_Null()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.Include = new[]
+                {
+                    "bin/*.dll",
+                    "bin/*.pdb"
+                };
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --include \"bin/*.dll\" --include \"bin/*.pdb\"", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Overwrite_To_Arguments_If_True()
+            {
+                // Given
+                var fixture = new OctopusDeployPackerFixture();
+                fixture.Id = "MyPackage";
+                fixture.Settings.Overwrite = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("pack --id MyPackage --overwrite", result.Args);
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


trailmax:
I'm trying your pull request in a custom build of Cake and run into an issue. In `OctopusPackSettings` property `BasePath` should really be `DirectoryPath`, not a `FilePath`. Unless I'm missing something.

gep13:
>@trailmax said...
I'm trying your pull request in a custom build of Cake and run into an issue. In OctopusPackSettings property BasePath should really be DirectoryPath, not a FilePath. Unless I'm missing something.

@cpx is the same not also true for OutputFolder?

cpx:
That does sound correct, I'll update the PR :)

cpx:
@trailmax How did your cakefile look, and what version of Octo.exe do you have? For me it worked with Octo 3.4.2 and the following cake script:
```
Task("Pack")
        .Does(() => {
                OctoPack("MyPackage", new OctopusPackSettings {
                        Version = "1.0.0",
                        BasePath = "./bin",
                        OutFolder = "./out",
                        Overwrite = true
                });
        });
```

I'll change it anyway to DirectoryInfo since that is more correct - would just like to understand why there's a difference between our tests.

trailmax:
@cpx Octo.exe - the latest. Yes, what you have it there worked - when you specify the path as a `String` it is fine. But I had a `Parameters` class with all the directories listed and my directories were recorded as `DirectoryPath` objects. And conversion from `FilePath` to `DirectoryPath` failed on compile time.

cpx:
@trailmax: Got it, that explains it.

src/Cake.Common/Tools/OctopusDeploy/OctopusPackFormat.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Represents the format of an Octopus package
+    /// </summary>
+    public enum OctopusPackFormat
+    {
+        /// <summary>
+        /// NuGet package
+        /// </summary>
+        NuPkg,
```
patriksvensson: I'm thinking if we should rename this to `NuGet`. What do you think?

src/Cake.Common/Tools/OctopusDeploy/OctopusPackFormat.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    /// <summary>
+    /// Represents the format of an Octopus package
```
patriksvensson: Missing period after `package`.

src/Cake.Common/Tools/OctopusDeploy/OctopusPackSettings.cs
```diff
@@ -0,0 +1,71 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using Cake.Core.IO;
+using Cake.Core.Tooling;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    /// <summary>
+    /// Contains the settings used by OctoPack
```
patriksvensson: Missing period after `OctoPack`.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="id">The package ID.</param>
+        /// <param name="settings">The settings.</param>
+        public void Pack(string id, OctopusPackSettings settings)
+        {
+            if (id == null)
+            {
+                throw new ArgumentNullException(nameof(id));
+            }
+
+            var builder = new OctopusPackArgumentBuilder(id, _environment, settings);
```
patriksvensson: Not that familiar with the Octopus tool wrappers, but is there any specific reason that you put the argument building into it's own class? I'm not against it but it does break the convention we've used so far (to my knowledge at least) of putting the argument building together with the tool. Like I said, not against it since I like the separation, just a thought.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
```diff
@@ -0,0 +1,67 @@
...
+            _environment = environment;
+        }
+
+        /// <summary>
+        /// Gets the name of the tool.
+        /// </summary>
+        /// <returns>The name of the tool.</returns>
+        protected override string GetToolName()
+        {
+            return "Octo";
```
patriksvensson: Perhaps this should be `OctopusDeploy Pack` or something similar?

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployAliases.cs
```diff
@@ -1,125 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Packs the specified folder into an Octopus Deploy package
+        /// </summary>
+        /// <param name="context">The cake context</param>
+        /// <param name="id">The package ID.</param>
+        /// <param name="settings">The settings</param>
+        [CakeMethodAlias]
+        public static void OctoPack(this ICakeContext context, string id, OctopusPackSettings settings = null)
```
patriksvensson: Could we create an overload for this method (without the settings) instead of using default values. We've seen some funny things happening with older versions of Roslyn when doing this.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployAliases.cs
```diff
@@ -1,125 +1,149 @@
...
+            {
+                throw new ArgumentNullException(nameof(packagePaths));
+            }
+
+            var pusher = new OctopusDeployPusher(context.FileSystem, context.Environment, context.ProcessRunner, context.Tools);
+            pusher.PushPackage(server, apiKey, packagePaths.ToArray(), settings);
+        }
+
+        /// <summary>
+        /// Packs the specified folder into an Octopus Deploy package
```
patriksvensson: Missing period after `package`.

src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPackTests.cs
```diff
@@ -0,0 +1,160 @@
...
+namespace Cake.Common.Tests.Unit.Tools.OctopusDeploy
+{
+    public sealed class OctoPackTests
+    {
+        public sealed class ThePackMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Id_Is_Null()
+            {
+                var fixture = new OctopusDeployPackerFixture();
```
patriksvensson: We normally divides the Arrange,Act,Assert in `Given`/`When`/`Then` comments.
Would this be possible to change?

```csharp
public void Test()
{
  // Given
  var fixture = new MagicFixture();

  // When
  var result = fixture.DoMagic();

  // Then
  Assert.Equal(42, result);
}

File: src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs (hunk 4 above)
Comment (trailmax) [inline src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs]: I've done the same with my PR - I just followed the pattern in already existing functionality for Octopus. Also a lot of Octopus's parameters for different commands are the same - it helps to extract these out.

File: src/Cake.Common/Tools/OctopusDeploy/OctopusPackArgumentBuilder.cs [hunk 9]
```diff
@@ -0,0 +1,94 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Core;
+using Cake.Core.IO;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    internal class OctopusPackArgumentBuilder
```
patriksvensson: Class should be sealed.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
```diff
@@ -0,0 +1,67 @@
...
+using Cake.Core;
+using Cake.Core.IO;
+using Cake.Core.Tooling;
+
+namespace Cake.Common.Tools.OctopusDeploy
+{
+    /// <summary>
+    /// The Octopus deploy package packer
+    /// </summary>
+    public class OctopusDeployPacker : Tool<OctopusPackSettings>
```
patriksvensson: Class should be sealed.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
patriksvensson: Ok, wasn't aware of the existing functionality. I guess this can be left as it is then. 😄

src/Cake.Common/Tools/OctopusDeploy/OctopusPackFormat.cs
cpx: NuPkg is the constant used by Octo.exe. From the help:
```--format=VALUE         Package format. Options are: NuPkg, Zip.```

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
cpx: What's the convention? Should the tool name match the associated Cake type/member names, or should it match the underlying tool (which is octo.exe in this case)?

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
cpx: In this case though the argument builder isn't reusable for any other commands. Pack is a local command whereas all other commands work with a remote Octopus server. So it might make more sense to merge the OctopusPackArgumentBuilder into the tool.

src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPackTests.cs
cpx: Personally I prefer to skip the comments and just be strict in dividing the test using linebreaks :) But I'll change it so it matches the rest of the test suite

cpx:
I've pushed the requested changes now :) Naming of the format enum values and whether to have a separate argument builder is still open for discussion though.

src/Cake.Common/Tools/OctopusDeploy/OctopusPackFormat.cs
patriksvensson: Yes, sounds good to me.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
patriksvensson: This should be a human readable string just describing the tool in the log. For `nuget.exe pack` tool we have `NuGet Pack` if I remember correctly.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
patriksvensson: I leave this decision to you. If the other Octo tools does it like this, we can follow the convention and refactor this later if it makes sense, but if you want to change it, it's also fine.

src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctoPackTests.cs
patriksvensson: Perfect :)

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
cpx: Aha, in that case "Octopus Deploy Pack" seems better. Although in that case, the other Octopus tool names should also be changed since they are now named just "Octo"

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
cpx: I'd prefer to merge it into the tool then, since these arguments currently are not reusable for any other Octo command.

trailmax:
Regarding enum naming I prefer to keep it `NuPkg` because that's what the underlying tool is using. This way there is no divergence between Cake API and octo.exe API.

src/Cake.Common/Tools/OctopusDeploy/OctopusDeployPacker.cs
cpx: Actually, NuGet has the same tool name (NuGet) for all tools as well, coming from the NuGetTool base class.

cpx:
Merged the builder into the tool. I think the name should be kept as Octo since that seems to be the convention used by e.g. NuGet.

patriksvensson:
@cpx Sorry for abandoning this PR. Life simply got in the way. I think this looks good for merging.

Could you rebase this PR against develop?

Thanks!

cpx:
@patriksvensson NP, it happens :) will rebase as soon as I'm at my machine.

cpx:
@patriksvensson Done!

patriksvensson:
Merged! Thanks for this! 👍

PR #1219, opened 2016-09-12
InferRuntimes does not require arguments

RehanSaeed:
More details can be found [here](https://github.com/dotnet/cli/issues/2278) and [here](https://github.com/dotnet/cli/issues/2372). This is currently blocking us [here](https://github.com/RehanSaeed/Serilog.Exceptions/pull/7).

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Restore/DotNetCoreRestorerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Restore/DotNetCoreRestorerTests.cs
@@ -83,6 +83,22 @@ namespace Cake.Common.Tests.Unit.Tools.DotNetCore.Restore
                 // Then
                 Assert.Equal("restore \"./src/*\"", result.Args);
             }
+            
+            [Theory]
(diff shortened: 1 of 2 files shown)


gep13:
@RehanSaeed was there a reason that you closed this?

RehanSaeed:
I tried add using `--infer-runtimes` without Cake but it didn't help.

It turns out that the root cause is a known dotnet CLI issue (See the linked issue https://github.com/cake-build/cake/issues/1261 for details). There is a very nasty workaround but no fix on the horizon :(.

gep13:
@RehanSaeed understood. However, is the change you suggested still a correct change in terms of usage of the cli?

RehanSaeed:
Yes, it is correct but the docs for this option state:
> Temporary option to allow NuGet to infer RIDs for legacy repositories
I have spent some time googling this option but there isn't much info on what this means. To me it means that at some point fairly soon when the tooling matures, this option will go away. So is it still worth it?

gep13:
@RehanSaeed I see what you mean.  Yes, perhaps best to leave this until it settles down.

PR #1233, opened 2016-09-16
GH1318: Add the ability to use default credentials for HttpClient when downlo…

Rookian:
We should be able to authenticate against any web service using by UseDefaultCredentials.
DownloadFileSettings will need a new property UseDefaultCredentials that will be used for HttpClientHandler to set UseDefaultCredentials.
We need this, when we download documents from a Sharepoint website.

### DIFF
--- a/src/Cake.Common/Net/DownloadFileSettings.cs
+++ b/src/Cake.Common/Net/DownloadFileSettings.cs
@@ -18,5 +18,10 @@ namespace Cake.Common.Net
         /// Gets or sets the Password to use when downloading the file
         /// </summary>
         public string Password { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that controls whether default credentials are sent when downloading the file
+        /// </summary>
+        public bool UseDefaultCredentials { get; set; }
     }
 }
--- a/src/Cake.Common/Net/HttpAliases.cs
+++ b/src/Cake.Common/Net/HttpAliases.cs
@@ -197,7 +197,7 @@ namespace Cake.Common.Net
             // We track the last posted value since the event seems to fire many times for the same value.
             var percentComplete = 0;
 
-            using (var http = new HttpClient())
+            using (var http = new HttpClient(new HttpClientHandler { UseDefaultCredentials = settings.UseDefaultCredentials }))
             {
                 if (!string.IsNullOrWhiteSpace(settings.Username) && !string.IsNullOrWhiteSpace(settings.Password))
                 {


src/Cake.Common/Net/HttpAliases.cs
```diff
@@ -197,7 +197,7 @@ public static void DownloadFile(this ICakeContext context, Uri address, FilePath
             // We track the last posted value since the event seems to fire many times for the same value.
             var percentComplete = 0;
 
-            using (var http = new HttpClient())
+            using (var http = new HttpClient(new HttpClientHandler { UseDefaultCredentials = settings.UseDefaultCredentials }))
```
gep13: @Rookian How does this impact on a user who wants to use a username and password to perform the authentication, rather than the DefaultCredentials?  Does one override the other?

src/Cake.Common/Net/DownloadFileSettings.cs
```diff
@@ -18,5 +18,10 @@ public sealed class DownloadFileSettings
         /// Gets or sets the Password to use when downloading the file
         /// </summary>
         public string Password { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that controls whether default credentials are sent when downloading the file
```
gep13: Since this is a boolean property, the summary text should start with `Gets or sets a value indicating whether`

src/Cake.Common/Net/HttpAliases.cs
Rookian: Both are different kind of authentications afaik. You usually use DefaultCredentials, if you use NTLM.

gep13: @Rookian yes, I understand that.  However, the DownloadFile alias currently has the ability to specify a username and password for the download to happen.  The change that you are suggesting initializes the HttpClient with UseDefaultCredentials.  My question is, will the specified username and password still work, as a result of the initialization code that you have put in place?

Rookian: The existing behaviour did not change, because UseDefaultCredentials is a bool and therefore is false, which is the default for HttpClient / HttpClientHandler. So this is not a breaking change.

gep13: And what happens if both UseDefaultCredentials, and Username and Password are set?

Rookian: I would say that this makes no sense. The current implementation is using Basic Authentication. My pull request is about supporting NTLM. Maybe we should abstract both into a separate SettingsObject i.e. BasicAuthenticationSettings and NTLMAuthenticationSettings.

gep13: It might not `make any sense` but right now, based on your current implementation, it is possible, and without testing out the code, I am trying to get to the bottom of what `would` happen.  I would have thought that something like this makes sense:

```
using (var http = settings.UseDefaultCredentials ? new HttpClient(new HttpClientHandler { UseDefaultCredentials = true }): new HttpClient())
```

following by, within the settings:

```
if (!settings.UseDefaultCredentials && !string.IsNullOrWhiteSpace(settings.Username) && !string.IsNullOrWhiteSpace(settings.Password))
```

That way, you are forcing the user to either use one authentication mechanism, or the other, but not both at the same time.

Rookian: If you like I can think of an API that would address this issue. But this would be definitely a breaking change. Your implementation is still irritating, because the user can still mix the different configuration.

gep13:
@Rookian in line with our contributing process (https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md) can you please create an issue, and link to this PR from it?

Also, would it be possible for you to rebase your changes on top of the latest develop branch? Thanks!

gep13:
Relates to #1318

gep13:
@Rookian I have rebased and merged the changes in this PR, and you can find the end result here: https://github.com/cake-build/cake/commit/73a3c2a98e5b05f53153a72ade83945c4397de5a

As a result of rebasing and merging, I am going to manually close this PR, however, your changes have been correctly attributed.  Thanks!

Rookian:
Great!

PR #1234, opened 2016-09-16
[DOC] Adds default CPU count behavior to MSBuild settings documentation

enkafan:
Default behavior without setting the MaxCpuCount will be to compile
projects one at a time. The documentation indicated how to set this to use
all your processors, but didn't indicate the behavior if you didn't.
Because the value of 0 means use all available CPUs I suspect many people
like myself would think that 0 is the default so no value was needed to be
explicitly specified.

See https://msdn.microsoft.com/en-us/library/bb651793.aspx


gep13:
@enkafan thank you for your contribution!

Issue #1032, opened 2016-07-06
Support async callbacks


ejball:
It would be nice if callbacks (e.g. the action callback of `Does()`) supported async callbacks, so that async operations within the callback could be awaited rather than blocked.

``` csharp
Task("Whatever").Does(() =>
{
    ...
    var response = httpClient.GetAsync("http://example.com").GetAwaiter().GetResult();
    ...
}
```

vs.

``` csharp
Task("Whatever").Does(async () =>
{
    ...
    var response = await httpClient.GetAsync("http://example.com");
    ...
}
```

RichiCoder1:
Related: #156

ejball:
Definitely, but to be clear, I'd be just as happy if it blocked on the returned async task. I just want to be able to call `Async` methods with `await` syntax. :-)

devlead:
Relates to PR #1258

adamhathcock:
Updated my PR again https://github.com/cake-build/cake/pull/1258

adamhathcock:
For 0.22?  I'll rebase again in that case :)

devlead:
I've assigned this to 0.23.0

devlead:
Fixed by #1258

jnm2:
This update broke all my code that was doing this:

```cs
void ReusableMethod()
{
    // ...
}

Task("Foo").Does(ReusableMethod);
```

Now I get a series of these messages:

```
Error: Error occurred when compiling build script: error CS0121: The call is ambiguous between the following methods or properties: 'CakeTaskBuilderExtensions.Does(CakeTaskBuilder<ActionTask>, Action)' and 'CakeTaskBuilderExtensions.Does(CakeTaskBuilder<ActionTask>, Func<Task>)'
```

jnm2:
These are the workarounds:
 - `Task("Foo").Does(() => ReusableMethod());`
 - `Task("Foo").Does((Action)ReusableMethod());`

PR #1258, opened 2016-10-04
GH1032: Support Async tasks

adamhathcock:
Issue: https://github.com/cake-build/cake/issues/1032

I have some time inbetween jobs so trying to do this as I attempted to do an async Task.

I'm confused on how to test out an actual script.  I tried to do an integration test.

I went the route of "async by default" so the execute of the `ActionTask` transforms an Action into a Func<Task> then just handle it from there.

### DIFF
--- a/src/Cake.Core.Tests/Fixtures/ScriptHostFixture.cs
+++ b/src/Cake.Core.Tests/Fixtures/ScriptHostFixture.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Threading.Tasks;
 using Cake.Core.Diagnostics;
 using Cake.Core.IO;
 using Cake.Core.Scripting;
@@ -18,9 +19,9 @@ namespace Cake.Core.Tests.Fixtures
             {
             }
 
-            public override CakeReport RunTarget(string target)
+            public override Task<CakeReport> RunTargetAsync(string target)
             {
-                return new CakeReport();
+                return System.Threading.Tasks.Task.FromResult(new CakeReport());
             }
         }
 
@@ -48,7 +49,7 @@ namespace Cake.Core.Tests.Fixtures
             Context.Log.Returns(Log);
 
             Engine = Substitute.For<ICakeEngine>();
-            Engine.RunTarget(Context, Arg.Any<IExecutionStrategy>(), Arg.Any<string>())
+            Engine.RunTargetAsync(Context, Arg.Any<IExecutionStrategy>(), Arg.Any<string>())
                 .Returns(new CakeReport());
         }
 
--- a/src/Cake.Core.Tests/Unit/ActionTaskTests.cs
+++ b/src/Cake.Core.Tests/Unit/ActionTaskTests.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Threading.Tasks;
 using Cake.Core.Tests.Fixtures;
 using Xunit;
 
@@ -32,7 +33,7 @@ namespace Cake.Core.Tests.Unit
                 var task = new ActionTask("task");
 
                 // When
-                task.AddAction(c => { });
+                task.AddAction(c => Task.CompletedTask);
 
                 // Then
                 Assert.Single(task.Actions);
@@ -49,7 +50,7 @@ namespace Cake.Core.Tests.Unit
             }
 
             [Fact]
-            public void Should_Throw_On_First_Failed_Action()
+            public async Task Should_Throw_On_First_Failed_Action()
             {
                 // Given
                 var task = new ActionTask("task");
@@ -59,14 +60,14 @@ namespace Cake.Core.Tests.Unit
                 task.Actions.Add((c) => throw new NotImplementedException());
                 task.Actions.Add((c) => throw new NotSupportedException());
                 task.Actions.Add((c) => throw new OutOfMemoryException());
-                var result = Record.Exception(() => task.Execute(context));
+                var result = await Record.ExceptionAsync(() => task.Execute(context));
 
                 // Then
                 Assert.IsType<NotImplementedException>(result);
             }
 
             [Fact]
-            public void Should_Aggregate_Exceptions_From_Actions()
+            public async Task Should_Aggregate_Exceptions_From_Actions()
             {
                 // Given
                 var task = new ActionTask("task");
@@ -77,7 +78,7 @@ namespace Cake.Core.Tests.Unit
                 task.Actions.Add((c) => throw new NotSupportedException());
                 task.Actions.Add((c) => throw new OutOfMemoryException());
                 task.SetDeferExceptions(true);
-                var result = Record.Exception(() => task.Execute(context));
+                var result = await Record.ExceptionAsync(() => task.Execute(context));
 
                 // Then
                 Assert.IsType<AggregateException>(result);
@@ -88,7 +89,7 @@ namespace Cake.Core.Tests.Unit
             }
 
             [Fact]
-            public void Should_Only_Aggregate_Exceptions_When_There_Are_Many()
+            public async Task Should_Only_Aggregate_Exceptions_When_There_Are_Many()
             {
                 // Given
                 var task = new ActionTask("task");
@@ -97,7 +98,7 @@ namespace Cake.Core.Tests.Unit
                 // When
                 task.Actions.Add((c) => throw new NotImplementedException());
                 task.SetDeferExceptions(true);
-                var result = Record.Exception(() => task.Execute(context));
+                var result = await Record.ExceptionAsync(() => task.Execute(context));
 
                 // Then
                 Assert.IsType<NotImplementedException>(result);
--- a/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Cake.Core.Tests.Fixtures;
 using Xunit;
 
@@ -93,15 +94,15 @@ namespace Cake.Core.Tests.Unit
             public sealed class WithTarget
             {
                 [Fact]
-                public void Should_Throw_If_Target_Is_Null()
+                public async Task Should_Throw_If_Target_Is_Null()
                 {
                     // Given
                     var fixture = new CakeEngineFixture();
                     var engine = fixture.CreateEngine();
 
                     // When
-                    var result = Record.Exception(() =>
-                        engine.RunTarget(fixture.Context, fixture.ExecutionStrategy, null));
+                    var result = await Record.ExceptionAsync(() =>
+                        engine.RunTargetAsync(fixture.Context, fixture.ExecutionStrategy, null));
 
                     // Then
                     AssertEx.IsArgumentNullException(result, "target");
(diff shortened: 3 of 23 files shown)


patriksvensson:
@adamhathcock Sorry for not getting back to you sooner.

Is the purpose of this PR to add support for the `await` keyword in scripts?

Although I'm not opposed to the idea, I feel reluctant to make this many changes in the public APIs in `Cake.Core`. Is there any other way we could make this happen?

adamhathcock:
I could probably try to change things to be additive instead.

Would basically be putting 'Task.Wait' in the old methods be bad or would you want a completely separate path for async?

devlead:
@adamhathcock I think we could start simple with keeping the Tasks synchronous but allowing for running asynchronous code with in the Task.

And an example usage of this could be:
```cake
Task("Async-Test")
    .DoesAsync(async () => {
            using (var client = new HttpClient())
            {
                await client.GetAsync("https://raw.githubusercontent.com/cake-build/cake/develop/build.sh");
                var result = await client.GetAsync("https://raw.githubusercontent.com/cake-build/cake/develop/build.sh");
                Information("Build.sh {0}", result.StatusCode);
            }
    })
    .DoesAsync(async context => {
            using (var client = new HttpClient())
            {
                var result = await client.GetAsync("https://raw.githubusercontent.com/cake-build/cake/develop/build.ps1");
                Information("Build.ps1 {0}", result.StatusCode);
            }
});
```
Will run those 2 actions async when Cake task executes.  And output something like
```
Build.ps1 OK
Build.sh OK
```
build.sh last because it this time took longer for som reason to download.

A rough 5-min implementation could look something like
```csharp
public static class CakeTaskBuilderAsyncExtensions
{
    private static readonly ConcurrentDictionary<string, IList<Func<ICakeContext, Task>>> AsyncTasks = new ConcurrentDictionary<string, IList<Func<ICakeContext, Task>>>();
    public static CakeTaskBuilder<ActionTask> DoesAsync(this CakeTaskBuilder<ActionTask> builder, Func<Task> action)
    {
        return DoesAsync(builder, context => action());
    }

    public static CakeTaskBuilder<ActionTask> DoesAsync(this CakeTaskBuilder<ActionTask> builder, Func<ICakeContext, Task> action)
    {
        AsyncTasks.AddOrUpdate(
            builder.Task.Name,
            key => {
                var tasks = new List<Func<ICakeContext, Task>> {action};
                builder.Does(context => Task.WaitAll(tasks.Select(task => task(context)).ToArray()));
                return tasks;
            },
            (key, list) => {
                list.Add(action);
                return list;
            });
        return builder;
    }
}
```

RichiCoder1:
My two cents: I really, *really* believe that Async/Await should a first class citizen in Cake and I like the direction this PR is heading. I think just patching async support in and using blocking `Task.Wait*` APIs is a bad idea and setting yourself up for long term pain.

patriksvensson:
@RichiCoder1 Yes, but that is a way bigger task to take on. A lot of code will have to be rewritten (most of the core) and to be honest I'm not that the return on investment would be worth it. I think that the change Mattias propose is something that doesn't intrude on the internals and the day we know how to rewrite the engine, it would be trivial to make these changes fit with the new engine.

RichiCoder1:
@patriksvensson I'll continue to respectfully disagree :), as I believe it's worth the investment. That being said, if there is an engine rewrite being pondering, that would be a good time to introduce this kind of broad change.

gep13:
@RichiCoder1 out of interest, what are the use cases or benefits that you see for this type of support?

adamhathcock:
Looking at @devlead 's suggestion, it seems he's doing something to execute things in parallel which is not what I'm trying to accomplish here.

I'm attempting to just execute Tasks synchronously but they're using the `async` and `await` keywords to allow usage of Task-based APIs with the pretty `await` keyword.

I definitely agree running things in parallel is a big change but I'm not seeing what I did as a big change.  It's just pushing some of the idea that a `CakeTask` would use `Task` into the engine.  I recognize that could break things but it should be done sometime :)

Anyway, true parallel execution of tasks is out of scope for me.

devlead:
@adamhathcock well it's only async/parallel with in the same task, in general people only have one "does", but if the do have two async does with in same task wouldn't expectation be to do both async?

adamhathcock:
@devlead You can save `DoesAsync` as a method to mean actually run in parallel.

However, if it's just `Does(Func<Task>)` or `Does(Action)` it's run synchronously.

When people chain `await` in code, it's still run in sync.  There's no parallelization unless there's a `Task.Run` used or something.  I'm just covering the first case using `await`, not the second with parallelization.

adamhathcock:
@patriksvensson any comment? (Or anyone else)  I don't think @devlead 's suggestion is the use case I'm covering.

I just want to support async/await.  Not true parallel execution.  I think changing the internals to do Tasks is a first step for both even if it breaks some things.

devlead:
@cake-build/team what do you think?

adamhathcock:
@devlead @patriksvensson I've updated my branch (I'll rebase) but just wanted to make sure can this be put in?

async/await in Cake scripts are desired for me :)

Cheesebaron:
Tools such as Windows Azure Storage now only have Async methods. Do you really, want Cake AddIn devs to wrap everything in Task.Run or TaskCompletionSource?

patriksvensson:
@adamhathcock Sorry for the radio silence about this. I haven't been able to put in as much time as I would have wanted in the project.

Would adding async/await support for tasks break existing scripts?

adamhathcock:
@patriksvensson Nope!  It adds a new overload for `Task` to be used by async scripts.  Internally, the old method gets converted a completed task (this change is in CakeTaskBuilderExtensions.cs I think)

The big changes are internal to make the engine be a Task (async all the way) but the scripts themselves can be `Task` or not.

I think the tests say it all.  All of the old test scripts work and I added a new one that has an `async Task` task using `await` inside.

patriksvensson:
@adamhathcock Will take a closer look at the PR to understand better how it works, but this change sounds OK to me 👍

@cake-build/cake-team Any thoughts about this?

patriksvensson:
@adamhathcock Since this PR contains breaking changes (ABI compatibility), we've decided to merge this PR with 0.21 if it's OK with you (since that release will contain other breaking changes).

src/Cake.Core.Tests/Fixtures/ScriptHostFixture.cs
```diff
@@ -18,9 +19,9 @@ public TestingScriptHost(ICakeEngine engine, ICakeContext context)
             {
             }
 
-            public override CakeReport RunTarget(string target)
+            public override Task<CakeReport> RunTarget(string target)
             {
-                return new CakeReport();
+                return System.Threading.Tasks.Task.FromResult(new CakeReport());
```
RichiCoder1: Isn't `System.Threading.Tasks` redundant here?

RichiCoder1:
Oh, probably a dum question, but are the script engines we use smart enough to know not to end script exection after `RunTarget` fires since it's async now? (would need to make sure that this is testing with tasks that actually run ask (file i/o etc...) otherwise the tasks just run sync and it'll work like normal)

patriksvensson:
@RichiCoder1 Excellent point. We should absolutely make sure that errors are handled properly before merging this PR.

src/Cake.Core.Tests/Fixtures/ScriptHostFixture.cs
patriksvensson: Yes, I think so as well.

RichiCoder1:
@adamhathcock If you don't get there before more, I can throw together some simple `.cake`s that exercise these paths. Been forever since I ran cake from source, but I should be able to try this branch out tomorrow :)

devlead:
@RichiCoder1 @adamhathcock preferably as additions to the integration tests.

RichiCoder1:
Ooo, true. I don't think I've touched cake source code since they were introduced majorly. I'll look into it.

devlead:
@RichiCoder1 the integration tests are located at
https://github.com/cake-build/cake/tree/develop/tests/integration

adamhathcock:
@RichiCoder1 it's still synchronous in that code execution is paused then resumed when the `await` ed code is ready.  There doesn't need to be much special case.  This is just usage of the `async`/`await` keywords.  Nothing executes in parallel.

@patriksvensson fine by me.  It is a breaking change in that sense and makes sense to group with others.

adamhathcock:
Also, error handling with `async`/`await` doesn't require anything special unless you're rewriting the stack.  I don't think extra error handling is necessary.  If there's already tests, then issues would have shown up as the engine is Task based now.

src/Cake.Core.Tests/Fixtures/ScriptHostFixture.cs
adamhathcock: This is actually necessary because there's a build failure if you don't fully qualify.  `ScriptHost.Task` conflicts.

RichiCoder1:
@adamhathcock Yes, but suppose you have something like the following (pseudocode warning):
```
#addin MyAsync.Library

var target = Argument("target", "Default");

Task("Work-With-A-File")
   .Does(async () =>
{
    await DoFileThingAsync();
});

Task("Default").IsDependentOn("Work-With-A-File");

RunTarget(target); // <--- Without an await, will this wait until Work-With-A-File is done, or continue once the sync portion of the code finishes?
```

src/Cake.Core.Tests/Fixtures/ScriptHostFixture.cs
RichiCoder1: Ahh, that makes sense :). And it does look better than `using Task = System.Threading.Tasks.Task`; since it's only used one.

adamhathcock:
Ah crap.  I didn't realize those are the same.  People need to make a RunTargetAsync method and have the old RunTarget call that with a wait.

I'll fix this soon.  Late here.

adamhathcock:
Okay, updated.  `RunTarget` calls `RunTargetAsync` on the script host.  I did some renaming to add Async suffixes on things that looked public like `ICakeEngine` to make it clear.

I'll collapse the commits later.

RichiCoder1:
When this is merged, will probably want to document the difference somewhere after this is merged.

adamhathcock:
@patriksvensson and @RichiCoder1 still on the 0.21 milestone?

patriksvensson:
@adamhathcock yes, I think so. Am I right @cake-team?

devlead:
@adamhathcock @patriksvensson I would think v0.22.0 as we inserted the .NET tooling parity milestone as 0.21.0, it's nearly ready to ship though. And not much left in 0.22.0 so should be able to ship not to distant future.

If pr could be rebased into one commit against latest develop we could review it directly post 0.21.0 release.

adamhathcock:
@devlead rebased again

adamhathcock:
Disappointed this didn't make it into 0.22 . rebased again

adamhathcock:
ping @patriksvensson

devlead:
@adamhathcock I'll review this and if I find no issues it'll be shipped with 0.23.0

patriksvensson:
@adamhathcock I'm really sorry about this. It's been so much work with getting the 0.22 out and for some reason, we forgot to attach this issue to a milestone, which in turn made us forget about it.

This is something that we really want included in Cake, so we hope that you can forgive us for the oversight.

adamhathcock:
@patriksvensson No big deal.  I was just wanting it in.  Makes my life a lot easier :)

devlead:
Integration tests fail with 
```bash
Error: Error occurred when compiling build script: /pullrequest/tests/integration/Cake.Common/IO/FileAsync.cake(17,54):
error CS1503: Argument 1: cannot convert from 'Cake.Core.IO.FilePath' to 'string'
/pullrequest/tests/integration/Cake.Common/IO/FileAsync.cake(18,55):
error CS1503: Argument 1: cannot convert from 'Cake.Core.IO.FilePath' to 'string'
```

adamhathcock:
Needed a ToString() :)

adamhathcock:
Rebased

devlead:
@adamhathcock your changes have been merged, thanks for your contribution 👍

adamhathcock:
:)

PR #1263, opened 2016-10-08
Corrects the grammar "do/does" in exception messages and tests

stevensona:
Hey guys,
 I was turned onto the project by the Hanselminutes podcast. I dove right in and noticed that the exception messages had a grammar error. So I updated the messages and associated tests to reflect the change.


gep13:
@stevensona thank you very much for taking the time to make these corrections, it is much appreciated!

gep13:
>@stevensona said...
I was turned onto the project by the Hanselminutes podcast. I dove right in 

Glad to hear that you liked the show! We were very grateful to @shanselman for having @patriksvensson on.  Let us know if you have any questions about getting started with Cake.

PR #1268, opened 2016-10-11
Fix for unquoted VSTest settings file path

jnm2:
I wanted to add a different path VSTest argument and I looked at `SettingsFile` as a pattern to copy, but I believe I found some issues with its current implementation.

The `SettingsFile` argument is not quoted. Filenames that contain spaces will cause problems. I've confirmed the expected behavior at the command line. What I'm not sure about is what the resolution should be.

Is there written guidance on handling of whitespace and more specifically in conjunction with rooting relative paths?
Should it mimic the assembly path argument building, `assemblyPath.MakeAbsolute(_environment).FullPath.Quote()`? This means `Local.RunSettings` would become `/settings:"/Working/Local.RunSettings"`. Is that outcome correct in the context of Cake? The PR currently reflects this, but as you see it impacted the test. Is `/settings:"Local.RunSettings"` more desirable?

Should a test be added specifically for whitespace in every parameter like this or is it not worth it?

One last question that affects all quoting in general. `\"` is the standard escape sequence to allow a single quote to exist inside a quoted argument, and if I remember correctly  `\\` is the escape sequence that would allow a literal `\"` to exist. `string.Quote()` naively surrounds the string with quotes, but what happens if the path _were_ ever to end with a `\`? You end up building this string:
`app /path "path\with\trailing\slash\" /notapath 3` which gets parsed as:
`app`, `/path`, and `path\with\trailing\slash" /notapath 3`.
The correct built string would double just the ending slash:
`app /path "path\with\trailing\slash\\" /notapath 3` which gets parsed as:
`app`, `/path`, `path\with\trailing\slash\`, `/notapath`, and `3`.

Literal (post-escape) `"` is incidentally not allowed in any path, but both `\` and `"` present opportunities for injection attacks depending where the path is coming from.

I apologize for the complexity, but it would be good to fix this bug and also set a precedent which people like me can refer to when we add new parameters and paths.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
@@ -204,7 +204,21 @@ namespace Cake.Common.Tests.Unit.Tools.VSTest
             var result = fixture.Run();
 
             // Then
-            Assert.Equal("\"/Working/Test1.dll\" /Settings:Local.RunSettings", result.Args);
+            Assert.Equal("\"/Working/Test1.dll\" /Settings:\"/Working/Local.RunSettings\"", result.Args);
+        }
+
+        [Fact]
+        public void Should_Quote_Absolute_SettingsFile_Path()
+        {
+            // Given
+            var fixture = new VSTestRunnerFixture();
+            fixture.Settings.SettingsFile = new FilePath("Filename with spaces.RunSettings");
+
+            // When
+            var result = fixture.Run();
+
+            // Then
+            Assert.Equal("\"/Working/Test1.dll\" /Settings:\"/Working/Filename with spaces.RunSettings\"", result.Args);
         }
 
         [Fact]
--- a/src/Cake.Common/Tools/VSTest/VSTestRunner.cs
+++ b/src/Cake.Common/Tools/VSTest/VSTestRunner.cs
@@ -64,12 +64,12 @@ namespace Cake.Common.Tools.VSTest
             // Add the assembly to build.
             foreach (var assemblyPath in assemblyPaths)
             {
-                builder.Append(assemblyPath.MakeAbsolute(_environment).FullPath.Quote());
+                builder.AppendQuoted(assemblyPath.MakeAbsolute(_environment).FullPath);
             }
 
             if (settings.SettingsFile != null)
             {
-                builder.Append(string.Format(CultureInfo.InvariantCulture, "/Settings:{0}", settings.SettingsFile));
+                builder.AppendSwitchQuoted("/Settings", ":", settings.SettingsFile.MakeAbsolute(_environment).FullPath);
             }
 
             if (settings.InIsolation)
@@ -79,12 +79,12 @@ namespace Cake.Common.Tools.VSTest
 
             if (settings.PlatformArchitecture != VSTestPlatform.Default)
             {
-                builder.Append(string.Format(CultureInfo.InvariantCulture, "/Platform:{0}", settings.PlatformArchitecture));
+                builder.AppendSwitch("/Platform", ":", settings.PlatformArchitecture.ToString());
             }
 
             if (settings.FrameworkVersion != VSTestFrameworkVersion.Default)
             {
-                builder.Append(string.Format(CultureInfo.InvariantCulture, "/Framework:{0}", settings.FrameworkVersion.ToString().Replace("NET", "Framework")));
+                builder.AppendSwitch("/Framework", ":", settings.FrameworkVersion.ToString().Replace("NET", "Framework"));
             }
 
             if (settings.Logger == VSTestLogger.Trx)


gep13:
@jnm2 have a look here:

https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/GitReleaseManager/Create/GitReleaseManagerCreator.cs#L109

This is probably the "standard" way of doing it,

jnm2:
@gep13 Oh yes, I wanted to do that but unfortunately `AppendQuoted` inserts an extra space before the quote (`/settings: "filepath"` vs `/settings:"filepath"`). That's not going to be able to be a go-to option for every scenario.

As they are today, [`QuotedArgument.Render`](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/Arguments/QuotedArgument.cs#L30) and `RenderSafe` will improperly quote anything ending with a backslash. Also, if the quoted argument is not a path, it may contain a quote and that is not properly escaped. I think I have fairly efficient `string.IndexOf`-based code to do this job properly. Tests should be added. Should I open a separate issue or PR for this?

gep13:
@jnm2 can you make AppendSwitchQuoted work for you?  Where the separator is an empty string?

jnm2:
@gep13 That put me on the right track. In fact, all the `string.Format` calls really should have been `AppendSwitch` or `AppendSwitchQuoted` calls. Does this look good to you?

jnm2:
Here's a draft of proper process argument escaping with tests, following the [MSDN spec](https://msdn.microsoft.com/en-us/library/bb776391.aspx): https://gist.github.com/jnm2/c5c840bf317605a40f5f56f944db4892

Should I go ahead with a separate PR for that?

gep13:
That's one for @patriksvensson to hopefully take a look at.

jnm2:
Okay. That part doesn't necessarily have to be done before this PR is merged. It would touch separate code with wider impact than this PR.
After this PR is through I'll submit another that exposes `TestCaseFilter` and `TestAdapterPath`.

src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
```diff
@@ -204,7 +204,7 @@ public void Should_Use_SettingsFile_If_Provided()
             var result = fixture.Run();
 
             // Then
-            Assert.Equal("\"/Working/Test1.dll\" /Settings:Local.RunSettings", result.Args);
+            Assert.Equal("\"/Working/Test1.dll\" /Settings:\"/Working/Local.RunSettings\"", result.Args);
```
gep13: @jnm2 since you noticed this issue when there was a path with a space in it, can you add another test that has a path with a space in it?

gep13:
@jnm2 this looks good to me, just a quick comment.

If you can fix that up, and rebase on latest develop, than I can get this merged in.  Also, as per our contributing document, ideally, we would have an issue that goes along with this pull request.  Can you please create one, and then link through to it?

Thanks!

src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
jnm2: Absolutely. Got two questions:
 1. Does this mean I have to add a `Name with spaces.RunSettings` file to the project? Where is the `Local.RunSettings` file? Is VSTest actually running against a real file in this test or are you just checking the arguments string?
 2. Should a test be added specifically for whitespace in *every* parameter or is it not worth it? I'm asking because I'm about to add some more after this PR.

jnm2:
@gep13 Okay. Are you asking because I need to follow the whole issue template thingy?

src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
gep13: @jnm2 these are not integration tests, but rather unit tests, so there are no files on the file system.  If this is required, then we would mock those files using the IFileSystem interface.  However, I don't think that is required here, or did I miss something?  I would have thought we could just add another test similar to this one:

https://github.com/jnm2/cake/blob/d159d45b25120fdfa9a47112d05ddf44e258bcb8/src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs#L201

Where the created FilePath contains a space, and make sure that this is reflected correctly in the arguments that are created.  No?

No, I don't think that is required, but perhaps @devlead and @patriksvensson can chime in here as well.

gep13:
@jnm2 sorry, you lost me.

jnm2:
@gep13 I went to create an issue and was presented with a [template](https://github.com/cake-build/cake/issues/new) for repro steps, output log, etc. Can I just copy out of the original PR post?

gep13:
Gotcha.  The issue template is really intended for when reporting bugs, problems.  For an issue with the code base, it is not really required.  Perhaps a slimmed down version of the text you have in the PR will suffice.

jnm2:
@gep13 Are separate commits okay when adding new tests?

Created issue #1271.

gep13:
@jnm2 we tend to stick with 1 commit per PR, and just don't let the scope of the PR creep to far.  In this case, a single squashed commit makes sense.

jnm2:
Everything done.

PR #1273, opened 2016-10-12
GH1279: Adding all current parameters for VSTest

jnm2:
This adds the following parameters:

 * Parallel
 * EnableCodeCoverage
 * UseVsixExtensions
 * TestAdapterPath
 * TestCaseFilter
 * Diag

This is all the usable parameters as per the latest VSTest in VS2015.3, except the `/List*` parameters which didn't quite fit the flow because they return information instead of running tests.
The summaries are straight from the console documentation, except...

Major complaint: I think there is great value in copying the command line help description of each parameter. The build-enforced `Gets or sets` and `Gets or sets a value indicating whether` is frustrating to work into the summary and brings down the documentation quality here and on the website. I especially like the `If set,` because it clarifies instinctively what the default is. Also, [this made me sad](https://github.com/jnm2/cake/commit/cef2a28a7c846462625ca146c2b3e0622527fd09#diff-c12c4a8cdb9f063b6dcb007fa44e6c09R76). So did [this](https://github.com/jnm2/cake/commit/cef2a28a7c846462625ca146c2b3e0622527fd09#diff-c12c4a8cdb9f063b6dcb007fa44e6c09R61).

Question: Lots of tests on `Should_Not_Use_*_By_Default` that are entirely identical. Would it be better to have a single default settings test?

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/VSTest/VSTestRunnerTests.cs
@@ -153,7 +153,7 @@ namespace Cake.Common.Tests.Unit.Tools.VSTest
         }
 
         [Fact]
-        public void Should_Not_Use_Isolation_By_Default()
+        public void Should_Have_No_Args_By_Default()
         {
             // Given
             var fixture = new VSTestRunnerFixture();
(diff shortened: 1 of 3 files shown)


jnm2:
And this:
> /// Gets or sets a value indicating whether [suppress root directory].
>
> *[HeatSettings.cs](https://github.com/cake-build/cake/blob/88a61b352eb49e97390124d6144cdca38ef40645/src/Cake.Common/Tools/WiX/Heat/HeatSettings.cs#L161)*

gep13:
@jnm2 was there an issue raised for this?

As per our contributing guidelines, we like to have an issue that we can reference: https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md

jnm2:
@gep13 Will do. #1279

jnm2:
So I still have those two questions. Can I get rid of the inflexible "Gets or sets a value UK indicating whether", and can the defaults tests be merged into once since they're identical?

patriksvensson:
@jnm2 No, we want to test one thing in every test.

gep13:
>@jnm2 said...
Can I get rid of the inflexible "Gets or sets a value indicating whether"

This is a standard convention used across the whole of Cake, so no, we aren't going to change that in this instance.

patriksvensson:
@jnm2 Yes, @gep13 is correct about this.

However, I take back what I wrote to you before. I misunderstood the question. Those identical tests can in fact be replaced with a single test. I misunderstood, sorry.

jnm2:
@gep13 @patriksvensson :+1:

jnm2:
@gep13 @patriksvensson :+1: Done.

patriksvensson:
@jnm2 Merged! Thank you for your contribution! 😄 👍

Issue #1274, opened 2016-10-12
Add GitLab CI build system support


reicheltp:
I run my builds on GitLab CI and it would be awesome if there is 1st class support for this environment. 

I will do a PR soon if nobody have any objections.

gep13:
@reicheltp no objections from me 👍

antwilliams:
Well this is good/bad timing (depending on your point of view)!

I'm just finishing off the test cases for my implementation of this. https://github.com/WarwickControl/cake/tree/feature/gitlab-ci

gep13:
Ha ha :-D Great minds think a like!

Perhaps you and @reicheltp can work together to finalize the implementation, and then submit the PR.

@antwilliams this is part of the reason why in the contributing documentation we ask that you submit an issue before starting to work on a PR.  That way, it reduces the changes of duplication of effort, and also effort on something that might not get pulled into Cake.

antwilliams:
Oops, I missed that (first) step in the guide in my eagerness!

@reicheltp - do you already have an implementation for this functionality? I have just pushed a completed implementation and test fixture for GitLab CI support. 
I am just ensuring everything works correctly on a local GitLab installation and then I will create a PR. Let me know if you have anything to add to my code and I can add you as a collaborator.

reicheltp:
@antwilliams Good that I've created an issue first, before starting to implement this. 😄 I will take a look on the PR now.

antwilliams:
I've just rebased my PR based on the latest development branch, this includes the changes merged with PR #1222. 

I notice that this ticket has been tagged as being part of the 0.17 release, I don't know when that is scheduled for but I am going on holiday tomorrow for a week. I won't be able to respond to any review actions until I get back.

devlead:
Fixed by #1275

PR #1222, opened 2016-09-13
GH1280: Added Go.CD build provider

RehanSaeed:
Added support for Go.CD. See documentation here: https://docs.go.cd/current/faq/environment_variables.html

PR #1275, opened 2016-10-12
GH1274: Add GitLab CI Build System

antwilliams:
This PR adds support for the build system GitLab CI (#1274).

I have added support for reading the environment variables added by GitLab as they are documented in [the GitLab CI readme](https://docs.gitlab.com/ce/ci/variables/README.html).

There are test fixtures for the new classes and I have extended the Build System test cases to cover GitLab as well.

I have run the test suite locally and everything passes as well as testing the new code on a GitLab instance to verify that the properties are all parsed correctly.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Build/GitLabCIFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Build/GitLabCIFixture.cs
@@ -1,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Common.Build.GitLabCI;
+using Cake.Core;
+using NSubstitute;
+
+namespace Cake.Common.Tests.Fixtures.Build
+{
+    internal sealed class GitLabCIFixture
+    {
+        public ICakeEnvironment Environment { get; set; }
+
+        public GitLabCIFixture()
+        {
+            Environment = Substitute.For<ICakeEnvironment>();
+            Environment.GetEnvironmentVariable("CI_SERVER").Returns((string)null);
+        }
+
+        public void IsRunningOnGitLabCI()
+        {
+            Environment.GetEnvironmentVariable("CI_SERVER").Returns("yes");
+        }
+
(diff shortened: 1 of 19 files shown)


reicheltp:
On the whole it looks quite well but I would not recommend the way you handle with environment variables requiring a least version of GitLab CI / runner.

For example the variable `CI_PIPELINE_ID` is only available for GitLab CI 8.10 or higher. In your implementation `GitLabCI.Build.PipelineId` will also return 0 if not defined. I think that if running on GitLabCI < 8.10 it should ether return null or throw an exception that it is not available for this version of GitLab. 
Especially for `string` values which are not available. `string.Empty` is not the result someone would expect if a value is not set. 

May we should ether change the default values for `int` to `-1` and `string` to `null` or compare against `CI_SERVER_VERSION` and throw an exception.

antwilliams:
Thanks for the feedback @reicheltp, I agree on the different variables requiring different versions of GitLab.
I wasn't sure what the "Cake Philosophy" was when it came to handling situations like that. Should we throw or should we return some sort of default value?

The issue in my mind is that if we throw then we will potentially have different behaviour between variables that depend on the server version through `CI_SERVER_VERSION` and variables that depend on the runner. There isn't a variable defined (or at least a documented one) that lets you determine the version of the runner, this means we will have to handle that gracefully.

We could:
- validate version on server variables and ignore runner version
- continue as implemented (return `String.Empty` or `0` for missing values)
- ignore version but throw an exception if an environment variable is missing (probably due to incorrect version)

I copied the philosophy of ignoring missing environment variables from the implementation of the other build systems, Bamboo and BitBucket specifically (since I am familiar with those systems too). 

Do any of the Cake maintainers have any feedback?

antwilliams:
I've just rebased this after PR #1222 was merged (merging build system changes is a pain!). 

I notice that the ticket associated with this PR has been tagged as being part of the 0.17 release, I don't know when that is scheduled for but I am going on holiday tomorrow for a week. I won't be able to respond to any review actions until I get back.

gep13:
@antwilliams thanks for rebasing, and for letting us know.  It may not be released until you get back, as we have quite a backlog of PR's that we want to work through.

devlead:
@antwilliams your changes have now been merged, thanks for your contribution 👍

PR #1284, opened 2016-10-16
Adding extension method to add multiple targets to DotNetBuildSettings

wwwlicious:
allows adding multiple targets to a dotnetbuildsettings class

usage:
```csharp
DotNetBuild(target, settings => 
                    settings
                    .WithTarget(new[] { "targetOne" , "targetTwo" }));
```
rather than
```csharp
var targets = new[] { "targetOne" , "targetTwo" };
DotNetBuild(target, settings => { 
     foreach(var t in targets) {                    
        settings.WithTarget(t);
    }
    return settings;
});
```

### DIFF
--- a/src/Cake.Common/Tools/DotNetBuildSettingsExtensions.cs
+++ b/src/Cake.Common/Tools/DotNetBuildSettingsExtensions.cs
@@ -36,7 +36,7 @@ namespace Cake.Common.Tools
         /// <param name="settings">The settings.</param>
         /// <param name="targets">The .NET build targets.</param>
         /// <returns>The same <see cref="DotNetBuildSettings"/> instance so that multiple calls can be chained.</returns>
-        public static DotNetBuildSettings WithTarget(this DotNetBuildSettings settings, params string[] targets)
+        public static DotNetBuildSettings WithTargets(this DotNetBuildSettings settings, IEnumerable<string> targets)
         {
             if (settings == null)
             {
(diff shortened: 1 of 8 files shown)


src/Cake.Common/Tools/DotNetBuildSettingsExtensions.cs
```diff
@@ -31,6 +31,25 @@ public static DotNetBuildSettings WithTarget(this DotNetBuildSettings settings,
         }
 
         /// <summary>
+        /// Adds .NET build targets to the configuration.
+        /// </summary>
+        /// <param name="settings">The settings.</param>
+        /// <param name="targets">The .NET build targets.</param>
+        /// <returns>The same <see cref="DotNetBuildSettings"/> instance so that multiple calls can be chained.</returns>
+        public static DotNetBuildSettings WithTarget(this DotNetBuildSettings settings, params string[] targets)
```
patriksvensson: Would prefer if we took a `IEnumerable<string>` here instead of params. Also, the method should be called `WithTargets` (plural).

wwwlicious:
Just curious, what does `IEnumerable<string>` provide over params?

patriksvensson:
@wwwlicious Well, this way we don't restrict the input to the method to an array and it's also more implicit. This goes hand in hand with the conventions in the rest of the project of returning the most specific type and accepting the most generic type.

gep13:
@wwwlicious is there an issue associated with this PR?

wwwlicious:
adding to https://github.com/wwwlicious/Cake.Extensions instead

Issue #1287, opened 2016-10-17
Guard against invalid path environment variables


daveaglick:
### What You Are Seeing?

`ArgumentException` during tool resolution if user's `PATH` environment variable contains invalid paths according to `System.IO`.

### What is Expected?

Skip the invalid paths.

### What version of Cake are you using?

0.16.2

### Are you running on a 32 or 64 bit system?

64

### What environment are you running on?  Windows? Linux? Mac?

Windows

### Are you running on a CI Server?  If so, which one?

No

### How Did You Get This To Happen? (Steps to Reproduce)

Have an invalid path in my environment variable.

### Output Log

~~~sh
Performing debug...
Attach debugger to process 6788 to continue
Debugger attached
Analyzing build script...
Analyzing e:/Code/CakeDocs/build.cake...
Processing build script...
Installing addins...
Installing NuGet package Octokit...
Error: System.ArgumentException: Illegal characters in path.
   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   at System.IO.Path.IsPathRooted(String path)
   at Cake.Core.IO.Path..ctor(String path)
   at Cake.Core.IO.DirectoryPath..ctor(String path)
   at Cake.Core.Tooling.ToolResolutionStrategy.<>c.<GetPathDirectories>b__12_0(String p)
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at Cake.Core.Tooling.ToolResolutionStrategy.GetPathDirectories()
   at Cake.Core.Tooling.ToolResolutionStrategy.LookInPath(String tool)
   at Cake.Core.Tooling.ToolResolutionStrategy.Resolve(IToolRepository repository, String tool)
   at Cake.Core.Tooling.ToolLocator.Resolve(String tool)
   at Cake.Core.IO.NuGet.NuGetToolResolver.ResolvePath()
   at Cake.NuGet.NuGetPackageInstaller.GetNuGetPath()
   at Cake.NuGet.NuGetPackageInstaller.Install(PackageReference package, PackageType type, DirectoryPath path)
   at Cake.Core.Scripting.ScriptProcessor.InstallAddins(ScriptAnalyzerResult analyzerResult, DirectoryPath installPath)
[...]

PR #1288, opened 2016-10-17
GH1287: Adds a try-catch around code that reads PATH environment variable

daveaglick:
#1287

### DIFF
--- a/src/Cake.Core/Tooling/ToolResolutionStrategy.cs
+++ b/src/Cake.Core/Tooling/ToolResolutionStrategy.cs
@@ -147,7 +147,16 @@ namespace Cake.Core.Tooling
             {
                 var separator = new[] { _environment.Platform.IsUnix() ? ':' : ';' };
                 var paths = path.Split(separator, StringSplitOptions.RemoveEmptyEntries);
-                result.AddRange(paths.Select(p => new DirectoryPath(p)));
+                foreach (var p in paths)
+                {
+                    try
+                    {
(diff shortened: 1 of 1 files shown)


daveaglick:
On second thought, also trims spaces and quotes from start and end of paths. Given that it's not unusual to find quotes in the PATH string (this is what tripped me up), trimming them seems like a reasonable way to accept paths that the user thinks are valid but that `System.IO` won't accept.

I also still think we should have the empty catch guard since we  probably never want to fail the build due to an external configuration like environment variables.

src/Cake.Core/Tooling/ToolResolutionStrategy.cs
```diff
@@ -147,7 +147,16 @@ private List<DirectoryPath> GetPathDirectories()
             {
                 var separator = new[] { _environment.Platform.IsUnix() ? ':' : ';' };
                 var paths = path.Split(separator, StringSplitOptions.RemoveEmptyEntries);
-                result.AddRange(paths.Select(p => new DirectoryPath(p)));
+                foreach (var p in paths)
+                {
+                    try
+                    {
+                        result.Add(new DirectoryPath(p.TrimStart(' ', '"', '\'').TrimEnd(' ', '"', '\'')));
```
devlead: If we Trim both Start & End can't we just [Trim(chars)](https://msdn.microsoft.com/en-us/library/d4tt83f9.aspx)?

devlead:
@daveaglick LGTM 👍 ,  your changes have merged and will be released in the next version of Cake. Thanks for contributing!

Issue #1289, opened 2016-10-17
Add user agent for DownloadFile


daveaglick:
### What You Are Seeing?

Exception from sites that require a user agent header when calling `DownloadFile()` (the GitHub API is one such server that doesn't like agentless requests).

### What is Expected?

A successful file download.

### What version of Cake are you using?

0.16.2

### Are you running on a 32 or 64 bit system?

64

### What environment are you running on?  Windows? Linux? Mac?

Windows

### Are you running on a CI Server?  If so, which one?

No

### How Did You Get This To Happen? (Steps to Reproduce)

Use the following code in a task (note that you must also have Octokit as an addin for this sample):
```
GitHubClient github = new GitHubClient(new ProductHeaderValue("Cake"));
Release release = github.Repository.Release.GetAll("cake-build", "cake").Result.First();
FilePath releaseZip = DownloadFile(release.ZipballUrl);
Unzip(releaseZip, releaseDir);
```

### Output Log

<details>
  <summary>Click to expand</summary>
~~~sh
Downloading file: https://api.github.com/repos/cake-build/cake/zipball/v0.16.2
An error occured when executing task 'GetSource'.
Error: System.AggregateException: One or more errors occurred. ---> System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.WebException: The server committed a protocol violation. Section=ResponseStatusLine
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Net.Http.HttpClientHandler.GetResponseCallback(IAsyncResult ar)
   --- End of inner exception stack trace ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Cake.Common.Net.HttpClientExtensions.<DownloadFileAsync>d__1.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
[...]

PR #1290, opened 2016-10-17
GH1289: Sets the user agent for HTTP aliases, #1289

daveaglick:
#1289 

Implements a method to get a new `HttpClient` with the `User-Agent` header set to "Cake" and the current Cake version for all HTTP aliases.


devlead:
@daveaglick your changes has now been merged, thanks for contributing 👍

Issue #260, opened 2015-06-02
Add support for moving directories


coxp:
I think the API should follow that for moving files.

* `MoveDirectoryToDirectory(directoryPath, targetDirectoryPath)` - Moves an existing directory to a new location (e.g. Move "build/project" to "package" results in "build/" and "package/project"). Target location must exist but full path to target location (once combined with directoryPath) must not.

* `MoveDirectories(pattern, targetDirectoryPath)`  - Moves existing directories matching the specified patern to a new location. Actual move implementation as MoveDirectoryToDirectory.

* `MoveDirectories(directoryPaths, targetDirectoryPath)` - Moves existing directories to a new location. Actual move implementation as MoveDirectoryToDirectory.

* `MoveDirectory(directoryPath, targetDirectoryPath)` - Moves an existing directory to a new location, providing the option to specify a new directory name. Target location must not exist. (e.g. Move "build/project" to "package" results in "build/" and "package/" if "package" did not already exist and an error if it did, whereas move "build/project" to "package/project" results in "build/" and "package/project"). 

I think the MoveDirectory API should only work for directories on the same volume. Alternatively, if we determine the targetDirectoryPath is on a different volume (not sure if there is an existing method for this?) then we could implement a method to create a new directory on the the target volume, move the files within that directory, recursively apply for any sub-directories, and then delete the source directory.

RichiCoder1:
I could be reading this wrong, but I have some objections to the `MoveDirectoryToDirectory` and the suggested `MoveDirectory` semantics. Move directory should have a behavior as close to possible to existing "move" commands on systems as possible. There should be a single `MoveDirectory`, relative directories should be based of the CWD, there should be no cross volume restriction, and the implemenation should defer to .NET or native APIs for actually moving the directory. Overwrite should be an explict option you have to set, and Move-ing a directory in such a way that it would conflict should throw an exception unless that option is specifically set to true. I don't know how I feel about MoveDirectory creating any missing parent directories, but would not be a bad idea.

coxp:
The cross volume restriction for moving a directory is from the underlying .NET API for [Directory.Move] (https://msdn.microsoft.com/en-us/library/system.io.directory.move%28v=vs.110%29.aspx). In addition, you can't overwrite an existing directory.

[...]

PR #1302, opened 2016-10-19
GH260: MoveDirectory alias

daveaglick:
Adds new MoveDirectory alias and associated support methods and tests, #260


devlead:
@daveaglick merged, thanks for your contribution 👍

Issue #1166, opened 2016-08-17
Feature request: Support optional parameters on alias methods


IanMercer:
Currently an alias with an optional parameter is only registered as a method requiring that parameter. Attempting to call it without the parameter results in an error `No overload for method '<methodname>' takes 0 arguments`.

My suggestion is that `ScriptAliasFinder` be extended to handle optional parameters.

gep13:
@IanMercer can you point at what alias you are referring to? By convention, we suggest that an alias should only have mandatory parameters, in order to help with discoverability.

IanMercer:
I created an alias that finds all nuspec files and returns their matching csproj files. I evolved it to take an optional path to search under. I was hoping the assembly containing this alias method could be used with our existing cake scripts but when it was dropped into an existing repository it cause the build to fail.

If the guidance is not to do this, that's fine: people can create both methods instead.

devlead:
Currently we don't codegen default values for optional parameters. Could be something to investigate for v0.17.0.

kcamp:
I have found similar - If you create an alias in your own local add-in with an optional parameter, it doesn't get discovered or emitted properly via the MethodAliasGenerator.  
Since defaults have to be compile-time constants, I think support could _potentially_ be added in MethodAliasGenerator / GetProxyParameters method, but could yield a brittle implementation.
Creating the actual method overloads in your add-in aliases seems to be a more preferable solution, in my mind.

IanMercer:
Maybe `GetAliasMethods` could emit a warning if there are any `IsOptional` parameters on the method. That would at least help anyone else who tries this.

devlead:
It should be supported for `Context.YourMethod` as you're not going thru the proxy methods.

Supporting optional parameters could affect the website documentation generation as we haven't had that on aliases before.

kcamp:
if there is interest in supporting this, I have started a POC branch https://github.com/kcamp/cake/commit/8d142deef6c61c1d86543371ea361d26fc708ced

It would need to be refined further to deal correctly with numerics and testing support added in to cover all the cases in both generic and non-generic methods.   i can pursue it further if desired.

devlead:
@kcamp have you done any more work around this? Willing to submit PR for review and further discussion?

kcamp:
[...]

PR #1306, opened 2016-10-21
GH1166: refactor MethodAliasGenerator to support optional parameters

kcamp:
@devlead 

Here is a PR for review to address GH-1166, refactoring the parameter concerns in the method alias tokenizer to support optional params.

I have added a reasonable cover of tests to prove this out, but it is not exhaustive - it does not cover all numerics, etc., but it does cover what I would consider to be the trickier of cases concerning Nullable<T> and so forth.

There are probably a couple of places where the generated code could be cleaned up for brevity, but as a first pass, I opted to be explicit, particularly in light of the fact that this isn't code that needs to be readable/maintainable, and the emphasis should be on being explicit for the sake of operability.

Happy to re-work or expand test-coverage as it might be deemed necessary.. I'm not saying "don't merge", but I think this addition warrants any extra attention we might want to give it.

### DIFF
--- a/src/Cake.Core/Scripting/CodeGen/ParameterEmitter.cs
+++ b/src/Cake.Core/Scripting/CodeGen/ParameterEmitter.cs
@@ -1,0 +1,134 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Cake.Core.Scripting.CodeGen
+{
+    /// <summary>
+    /// Responsible for generating parameter tokens in method alias generation
+    /// </summary>
+    internal sealed class ParameterEmitter
+    {
+        internal static string Emit(ParameterInfo parameter, bool includeType)
+        {
+            return string.Join(string.Empty, BuildParameterTokens(parameter, includeType));
+        }
+
+        private static IEnumerable<string> BuildParameterTokens(ParameterInfo parameter, bool includeType)
+        {
+            if (parameter.IsOut)
+            {
+                yield return "out ";
+            }
+            else if (parameter.ParameterType.IsByRef)
+            {
+                yield return "ref ";
+            }
+            if (includeType)
+            {
+                if (parameter.IsDefined(typeof(ParamArrayAttribute)))
+                {
+                    yield return "params ";
+                }
+                yield return parameter.ParameterType.GetFullName();
+                yield return " ";
+            }
+            yield return parameter.Name;
+
+            // GH-1166; add support for specifying default parameter values
+            if (includeType && parameter.IsOptional)
+            {
+                yield return " = ";
+                yield return BuildDefaultParameterValueToken(parameter);
+            }
+        }
+
+        private static string BuildDefaultParameterValueToken(ParameterInfo parameter)
+        {
+            var type = parameter.ParameterType;
+            var value = parameter.RawDefaultValue;
+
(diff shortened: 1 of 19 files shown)


kcamp:
Will have a look at this in the next couple of days and rework.  Thanks.

gep13:
@kcamp  @patriksvensson just wanted to touch base with this one...

Where are we with this?  This is one of the two remaining items that I would like to get into the 0.17.0 release.  Should we bump this to the next release?

kcamp:
I have been re-working the implementation and have updated the PR.  It will still not be passing CI on Mono/Ubuntu, so the failures are expected.

I have refactored the `MethodAliasGenerator` class to separate out the generation/emission of the parameter tokens, in part to ease the creation of test cases for the parameters.

The original two failures centered around `System.Decimal` and `System.Enum` derivatives.  I have successfully patched the enum related-failures.  The failures relating to `System.Decimal` persist, and I am hoping that someone has an idea on how we can get that resolved.

The failure I am seeing is related _only_ to `Nullable<decimal>` with a non-null default values specified in the signature.  `decimal` parameters in other capacities (both required and optional, but absent the `Nullable<>`) work correctly.

There are two failing tests, both fail in a similar fashion:
```
Should_Return_Correct_Generated_Code_For_Method_Parameters ( "OptionalNullableDecimalWithNonNullDefault" )
Expected: ... (System.Decimal)1
Actual:     ... (System.Decimal)System.Reflection.Missing
``` 
For reference:
```
Mono JIT compiler version 4.6.1 (Stable 4.6.1.5, ef43c15, Weds. Oct 12
```
Any ideas on how I might work around it?  I can't find anything that speaks directly (or even indirectly) to what I am seeing in a search

Thanks.

kcamp:
@gep13 to answer your question more directly..  i think we're close to having this working.  see my last comment  for more details.  If we can figure out the last remaining test failure and I can get it fixed up, it should meet requirements and be complete.  I don't know the timeframe you are thinking for 0.17.0, but if I can get a hand on the last remaining issue in Linux/Mono and get things cleaned up, I could potentially have the PR updated with the final bits today/tomorrow.

kcamp:
@gep13 @devlead @patriksvensson 

I've figured out a way to patch the defect that was manifesting itself on Mono environments,  You can see that in commit 1009149

(I do not love having to add conditionals like that, but given the very repeatable failure on Ubuntu/OSX and the fact that this appropriately addresses them and passes checks, it seems like a reasonable solution.)

If you're happy with the way things look for this implementation, I will rebase + squash and update the PR.  Let me know.

src/Cake.Core/Scripting/CodeGen/ParameterEmitter.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Cake.Core.Scripting.CodeGen
+{
+    /// <summary>
+    /// Responsible for generating parameter tokens in method alias generation
+    /// </summary>
+    internal class ParameterEmitter
```
patriksvensson: Should be sealed.

kcamp: @patriksvensson will by synchronizing my branch against develop this morning; will take care of this request and squash.  thanks!

kcamp:
@patriksvensson requested changes to access modifier are made, rebased on develop + squashed.

gep13:
@kcamp similar to the suggestion I made about the NamespaceImport, would you be in a position to document the new optional parameters for aliases?

kcamp:
@gep13 sure, I can put together some docs.

Created cake-build/website#227 to track work + PR against.

Issue #1344, opened 2016-11-04
Cake's default tools / addins / modules paths are not so default as they seem.


nalla:
### What You Are Seeing?
Cake ist installing addins / tools and modules *not* relative to the build script.

### What is Expected?
Cake should install addins / tools and modules relative to the build script - per default configuration.

### What version of Cake are you using?
16.0.2

### Are you running on a 32 or 64 bit system?
64 bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)

```PowerShell
$> cd C:\
$> New-Item -Type Directory repo
$> cd repo
$> nuget install Cake -ExcludeVersion -OutputDirectory .\tools\foo
$> "#addin Cake.FileHelpers" > cake.build
$> .\tools\foo\Cake\Cake.exe .\cake.build
$> ls .\tools

    Directory: C:\repo\tools

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       04.11.2016     17:06                foo

$> ls .\tools\foo

    Directory: C:\repo\tools\foo

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       04.11.2016     17:07                Addins
d-----       04.11.2016     17:07                Cake
```

### What is Expected?

```
$> ls .\tools

    Directory: C:\repo\tools\foo

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       04.11.2016     17:07                Addins
d-----       04.11.2016     17:07                foo
```

nalla:
As already noted in the PR, only the addins folder was affected.

devlead:
fixed by #1348

RehanSaeed:
I'm trying to change the cake tools folder. I've changed this line in my `build.ps1` to point to a `cake-tools` folder:

```
$TOOLS_DIR = Join-Path $PSScriptRoot "cake-tools"
```

However, NuGet addins are still downloaded to a `tools` folder. Will this PR fix that?

devlead:
Cake's default folders can be changed thru configuration, read more about it at
http://cakebuild.net/docs/fundamentals/configuration

You can read more about the default values and how to override then at
http://cakebuild.net/docs/fundamentals/default-configuration-values

PR #1348, opened 2016-11-08
GH1344: addin-path: Use relative default path

nalla:
The default path that is used for addin installation is being constructed
by using the application root. This is not the propagated default behavior
on the website. Let's use the relative path `./tools/addins` as default.

This fixes (GH-1344)

---

It turned out that only the default of the addins installation folder was affected.

### DIFF
--- a/src/Cake.Core/Scripting/ScriptRunner.cs
+++ b/src/Cake.Core/Scripting/ScriptRunner.cs
@@ -138,8 +138,7 @@ namespace Cake.Core.Scripting
             _processor.InstallTools(result, toolsPath);
 
             // Install addins.
-            var applicationRoot = _environment.ApplicationRoot;
-            var addinRoot = GetAddinPath(applicationRoot);
+            var addinRoot = GetAddinPath(scriptPath.GetDirectory());
             var addinReferences = _processor.InstallAddins(result, addinRoot);
             foreach (var addinReference in addinReferences)
             {
@@ -150,6 +149,7 @@ namespace Cake.Core.Scripting
             var session = _engine.CreateSession(host, arguments);
 
             // Load all references.
+            var applicationRoot = _environment.ApplicationRoot;
             var assemblies = new HashSet<Assembly>();
             assemblies.AddRange(_conventions.GetDefaultAssemblies(applicationRoot));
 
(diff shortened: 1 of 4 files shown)


src/Cake.Core/Scripting/ScriptRunner.cs
```diff
@@ -212,15 +212,15 @@ private DirectoryPath GetToolPath(DirectoryPath root)
...
+        private DirectoryPath GetAddinPath(DirectoryPath root)
         {
             var addinPath = _configuration.GetValue(Constants.Paths.Addins);
             if (!string.IsNullOrWhiteSpace(addinPath))
             {
                 return new DirectoryPath(addinPath).MakeAbsolute(_environment);
             }
 
-            return applicationRoot.Combine("../Addins").Collapse();
+            return root.Combine("tools/Addins").Collapse();
```
patriksvensson: Tool path should not be hard coded, but should be fetched from the configuration file.

nalla: @patriksvensson Well, that depends. http://cakebuild.net/docs/fundamentals/default-configuration-values clearly states that the _Addins Path_'s default is `./tools/Addins`. No relationship to the _Tools Path_ is mentioned.

nalla: Furthermore, see https://github.com/cake-build/cake/blob/develop/src/Cake/Composition/ModuleLoader.cs#L55. The default _Modules Path_ is also not related to the set tools path.

nalla:
@patriksvensson thanks for your feedback. I added some justifications to the review.

src/Cake.Core/Scripting/ScriptRunner.cs
patriksvensson: @nalla That is an oversight. We should not hard code these paths anywhere. We should get the tools directory. The documentation should state something like `The default path is [tool path/Addins`.

nalla: @patriksvensson Then it would be wize to update https://github.com/cake-build/cake/blob/develop/src/Cake/Composition/ModuleLoader.cs#L55 as well - within this PR. To be some what consistent?

devlead:
@patriksvensson well the apparent scenario it'll affect is when Cake is in a central location like with the [Cake.Portable](https://chocolatey.org/packages/cake.portable) Chocolatey package. Don't think it'll break anything, more like create redundant packages, but that's likely what you want anyways so I don't have any issue with this.
@gep13 what's your thoughts?

gep13:
@devlead nope, I don't think I have an issue with this, unless I am missing something.

Issue #1352, opened 2016-11-11
Add support for importing namespaces at the assembly level


kcamp:
It would be useful to provide the ability to import namespaces in a global way at the assembly level. 

When creating an add-in library, there are namespaces that are desirable to be imported at a global level as opposed to decorating classes/methods in a redundant way, particularly when the classes within those namespaces are used from integration points that _aren't_ an alias method.   i.e., I have a base `BuildParameters` style class that exposes a certain amount of logic to the build script, but doesn't actually involve any method aliases.

Will submit PR for the same shortly.

devlead:
Fixed by #1353

gep13:
@kcamp I think we could benefit with some documentation to go along with this change.  Would you be in a position to help with this?

kcamp:
@gep13 i can absolutely lend a hand with this.  this would be under the cake-build/website project?  existing page/content or new?

gep13:
Yew, this would be on the website repo. Off the top of my head, not sure if there is an existing place for this or not. If not, then yes, a new page, and perhaps new group.

kcamp:
@gep13 sounds good.  i will make a pass at it and see what we can come up with.

devlead:
@kcamp @gep13 this would be under
http://cakebuild.net/docs/contributing/documentation

PR #1353, opened 2016-11-11
GH1352: add support for global namespace imports at the assembly level

kcamp:
This addresses GH-1352.  Changes opens the `CakeNamespaceImportAttribute` up to be applied at the assembly level and extends the logic in `ScriptAliasFinder` to import namespaces declared for import at the assembly level.


devlead:
@kcamp your changes have now been merged, thanks for cour contribution 👍

PR #1359, opened 2016-11-14
GH1053: Added support for custom #load schemes

### LINKED ISSUE #1053: Feature request downloadable .cake script directive
I need to use the #load but its not working with esey when i host it with nuget as it needs a path so i created the #nuscript directive.

hope this can get resolved and even added if its all done ok.

patriksvensson:
Also adds support for loading scripts from NuGet packages.

Closes #1053

### DIFF
--- a/src/stylecop.json
+++ b/src/stylecop.json
@@ -19,7 +19,6 @@
           "newlineAtEndOfFile": "allow"
       },
       "maintainabilityRules": {
-          
       },
       "orderingRules": {
           "usingDirectivesPlacement": "outsideNamespace",
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ dotnet-install.sh*
 launchSettings.json
 project.lock.json
 
+# Rider
+.idea/
+
 # Build related
 build-results/
 tools/Cake/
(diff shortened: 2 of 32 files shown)


build.cake
```diff
@@ -46,6 +46,7 @@ Setup(context =>
 //////////////////////////////////////////////////////////////////////
 
 Task("Clean")
+    .WithCriteria(() => false)
```
gep13: @patriksvensson don't think this is needed anymore 😄

src/stylecop.json
```diff
@@ -11,13 +11,13 @@
           "documentPrivateElements": false,
           "documentInterfaces": true,
           "documentPrivateFields": false,
+          "xmlHeader": false,
```
gep13: @patriksvensson should this be part of a PR for the other issue?

tlycken:
Nice - I've been looking forward to this!

What's the syntax to make use of it? Is it documented somewhere? (Couldn't find any docs chagnes in the commit history here...)

gep13:
>@tlycken said...
What's the syntax to make use of it? Is it documented somewhere? 

That is something that will make it's way onto the website.  Follow this issue [here](https://github.com/cake-build/website/issues/232).  Although this PR is closed, the milestone that this is included in has not been released yet, so you won't be able to make use of it, unless you are pulling Cake from our MyGet feed.

Having said that, it follows the same conventions as you can see here:

http://cakebuild.net/docs/fundamentals/preprocessor-directives

i.e. you should be able to do:

`#load nuget:?package=Cake.Recipe`

where Cake.Recipe is the name of the package that you want to load.

Issue #600, opened 2015-12-21
ParseAssemblyInfo does not work (on .dll and .vb)?


AnderssonPeter:
I have the following script:

```
var webAssemblyFile = File(@"Test.Web\bin\Test.Web.dll");
var webAssembly = ParseAssemblyInfo(webAssemblyFile);
Information("File: " + webAssemblyFile);
Information("AssemblyFileVersion: " + webAssembly.AssemblyFileVersion);
Information("AssemblyInformationalVersion: " + webAssembly.AssemblyInformationalVersion);
Information("AssemblyVersion: " + webAssembly.AssemblyVersion);
Information("Title: " + webAssembly.Title);
Information("File Exists: " + FileExists(webAssemblyFile).ToString());
```

And it generates the following output:

```
File: Test.Web/bin/Test.Web.dll
AssemblyFileVersion: 1.0.0.0
AssemblyInformationalVersion: 1.0.0.0
AssemblyVersion: 1.0.0.0
Title:
File Exists: True
```

Buf if i right click the dll file and use properties i see 2.5.3.0 as version..

i have also tried to point it to my AssemblyInfo.vb and it returns the same result..

```
File: Test.Web/My Project/AssemblyInfo.vb
AssemblyFileVersion: 1.0.0.0
AssemblyInformationalVersion: 1.0.0.0
AssemblyVersion: 1.0.0.0
Title:
File Exists: True
```

So what am i doing wrong?

gep13:
@Petoj87 can you provide a sample of your AssemblyInfo.vb file? Thanks!

Petoj87:
AssemblyInfo.vb:
```
Imports System
Imports System.Reflection
Imports System.Runtime.InteropServices

' General Information about an assembly is controlled through the following 
' set of attributes. Change these attribute values to modify the information
' associated with an assembly.

' Review the values of the assembly attributes
<Assembly: AssemblyTitle("Test.Web")> 
<Assembly: AssemblyDescription("")> 
<Assembly: AssemblyCompany("Test")> 
<Assembly: AssemblyProduct("Test")> 
<Assembly: AssemblyCopyright("Copyright © Test 2012")> 
<Assembly: AssemblyTrademark("")> 

<Assembly: ComVisible(False)> 

'The following GUID is for the ID of the typelib if this project is exposed to COM
<Assembly: Guid("334e9cee-0d47-4d70-924b-b5098a3432cb")>

' Version information for an assembly consists of the following four values:
'
[...]

PR #1380, opened 2016-11-25
GH600: Added ability to parse/create VB AssemblyInfo files

gep13:
Relates to #600


gep13:
@devlead what were the codefactor issues here? Something that needed to be taken care of?

devlead:
Complained about duplicated code in test fixture.

gep13:
Ah, is that all it was.  Coolio.

Issue #1386, opened 2016-11-30
Add CakeNamespaceImport for BuildSystem Aliases


phillipsj:
The following build system aliases do not have the CakeNamespaceImport attribute:

- TeamCity
- MyGet
- Bamboo

phillipsj:
Pull request will be coming shortly.

gep13:
@phillipsj thanks for adding this! :+1:

devlead:
Fixed by #1387

PR #1387, opened 2016-11-30
GH1386: Adding missing NamespaceImports for Build Aliases

phillipsj:
Adding the missing imports for TeamCity, MyGet, and Bamboo.


gep13:
Relates to #1386

PR #1388, opened 2016-12-01
Fix doc comments in InnoSetupAliases

thomaslevesque:
The comments were referring to the ".nsi" script file, which is the extension for NSIS, instead of ".iss", which is the extension for InnoSetup.


gep13:
@thomaslevesque this isn't something that I use, but I am willing to accept your authority on the correct file type here.

Thanks for submitting this PR!

Issue #1377, opened 2016-11-24
Support for uninstall packages using Chocolatey


mvput:
Add support for uninstalling packages using chocolatey. Currently there is no support for this. The case is to install tools using chocolatey during the build and uninstall when the build finishes to keep the build environment as clean as possible.

phillipsj:
@gep13 I can take this one if that works for you.

gep13:
@phillipsj sure thing, happy for you to take this.

devlead:
Fixed by #1390

PR #1390, opened 2016-12-02
GH1377: Support for uninstall packages using Chocolatey 

phillipsj:


phillipsj:
@gep13 Yep I will, it may be tomorrow.

gep13:
Perfect thanks!

phillipsj:
@gep13 It has been rebased and squashed.

gep13:
@phillipsj would you be in a position to correct the duplication error that CodeFactor is throwing?  I think correcting this is justified, as quite a few of the runners have the same properties.  You should be able to find examples in the code base where a common base class is used between runners.

phillipsj:
@gep13 Yep I will fix it this week, I have seen the other locations.

phillipsj:
@gep13 According to CodeFactor I need to implement more, but wanted to push so you could review.  Am I going in a good direction or would you prefer a different approach.

gep13:
@phillipsj the approach you are following here looks good to me!

gep13:
@phillipsj at some point can you rebase, and I will take a final look over it. Thanks!

devlead:
Relates to #1377

Issue #1394, opened 2016-12-06
Extend the ICakeEnvironment to contain the PathComparer


nalla:
As [discussed on Gitter](https://gitter.im/cake-build/cake?at=582972d188fa21d53c1b8ca7) the `ICakeEnvironment` could provide the environment specific `PathComparer` as an easy to use property. The comparer can then be accessed in a cake script just by calling

```c#
var filtered = GetFiles(./**/*.*).Distinct( Conext.Environment.PathComparer );
```

gep13:
@nalla thank you for raising this, but at this time, we don't think that this is required.

Comment from PR:

>@patriksvensson said...
Consumers could always use the PathComparer.Default to get the default path comparer for the environment they're on.

As a result, we have decided to close this issue.

Please feel free to re-open this issue if further discussion is required.

PR #1395, opened 2016-12-06
GH1394: Add path comparer property to CakeEnvironment

nalla:
The cake environment now contains an extra property that can be used to
directly access the environment specific path comparer. In a cake script
somebody would simply access the property via the context.

    var files = GetFiles(./**/*.*)
    var filtered = files.Distinct( Conext.Environment.PathComparer);

This fixes (GH-1394)


patriksvensson:
@devlead I have nothing against the property per se, but like you say, it's a breaking change so perhaps we shouldn't change this. Consumers could always use the `PathComparer.Default` to get the default path comparer for the environment they're on.

gep13:
@nalla first of all, I wanted to thank you for your contribution to the Cake project.  As always, this is much appreciated.  Given the fact that this would be a breaking change, and the fact that there is a workaround, I am of the opinion that we should close this issue, and the associated Pull Request.

Please let us know if you want to discuss this further.  Thanks.

PR #1402, opened 2016-12-08
GH1403: Add optional Go.CD Server URL Parameter to GetHistory

RehanSaeed:
A small tweak to the Go.CD GetHistory function to allow you to optionally set a GO.CD server URL.

### DIFF
--- a/src/Cake.Common/Build/GoCD/IGoCDProvider.cs
+++ b/src/Cake.Common/Build/GoCD/IGoCDProvider.cs
@@ -34,5 +34,14 @@ namespace Cake.Common.Build.GoCD
         /// <param name="password">The Go.CD password.</param>
         /// <returns>The Go.CD build history.</returns>
         GoCDHistoryInfo GetHistory(string username, string password);
+
+        /// <summary>
+        /// Gets the Go.CD build history, including the repository modifications that caused the pipeline to start.
+        /// </summary>
+        /// <param name="username">The Go.CD username.</param>
+        /// <param name="password">The Go.CD password.</param>
+        /// <param name="serverUrl">The Go.CD server URL.</param>
+        /// <returns>The Go.CD build history.</returns>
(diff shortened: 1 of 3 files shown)


gep13:
@RehanSaeed thank you for submitting this change.  Before going any further, can I please ask you to read our [contributing guide](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md) and please take the necessary steps.

RehanSaeed:
@gep13 Was in the middle of writing the issue https://github.com/cake-build/cake/issues/1403

gep13:
>@RehanSaeed said...
@gep13 Was in the middle of writing the issue #1403 which I assume is what you are referring to.

That's the one, yes.  That guideline is put in place to protect contributors.  In this case, it isn't/wasn't really needed, but we prefer to get sign off about any chances, prior to any work being done.  That way, there is less chance of a PR not being accepted.

src/Cake.Common/Build/GoCD/IGoCDProvider.cs
```diff
@@ -32,7 +32,8 @@ public interface IGoCDProvider
         /// </summary>
         /// <param name="username">The Go.CD username.</param>
         /// <param name="password">The Go.CD password.</param>
+        /// <param name="serverUrl">The Go.CD server url.</param>
         /// <returns>The Go.CD build history.</returns>
-        GoCDHistoryInfo GetHistory(string username, string password);
+        GoCDHistoryInfo GetHistory(string username, string password, string serverUrl = null);
```
patriksvensson: Would prefer to have two overloads here instead of an optional argument.

devlead: Yes, optional parameters breaks ABI, so overload is needed.

RehanSaeed:
Switched to two overloaded methods.

devlead:
@RehanSaeed your changes have been merged, thanks for contributing 👍

PR #1434, opened 2017-01-07
Fix commented example for DotNetCoreTest

mrahhal:


gep13:
@mrahhal thank you for taking the time to submit this change, really appreciated!

Issue #1438, opened 2017-01-11
Teach XmlPeek to silence warnings, if needed


nalla:
The current implementation of the `XmlPeek` alias throws a warning if it cannot find the requested `XPath`. While this is a good thing, there are those situations where could guess multiple values inside the given xml file. In those situations it would be pleasant if one could silence those warnings.

E.g.

```c#
Task( "ConsistencyCheck" )
	.Does( () =>
{
	bool consistent = true;

	Parallel.ForEach( GetFiles( "./src/**/packages.config" ), file =>
	{
		foreach( var dependency in GetDependencies() )
		{
			string version = XmlPeek( file, "/packages/package[@id='" + dependency.Id + "']/@version" );

			if( string.IsNullOrWhiteSpace( version ) )
				continue;

			if( version != dependency.Version )
			{
				Error( "You are using v" + version + " in " + file + " instead of v" + dependency.Version );
				consistent = false;
			}
		}
	} );

	if( !consistent )
		throw new Exception("Found at least one invalid consistency." );
} );
```

would be such a situation.

See https://gitter.im/cake-build/cake?at=58761aebcbcb28177055fb83 for a gitter reference.

devlead:
Fixed by #1440

PR #1440, opened 2017-01-11
xmlpeek: Add switch to suppress warnings

nalla:
The `XmlPeekSettings` learned to remember a setting if warnings should be suppressed or not. That setting will be evaluated before calling `context.Log.Warning()` if the `XmlPeek` alias can not find the provided xpath.

With this new setting in place, users can now _opt out_ of displaying the warning if the xpath was not found.

This fixes (GH-1438)


devlead:
@nalla your changes have been merged! Thanks for your contribution 👍

Issue #1294, opened 2016-10-18
Add signing of Cake binaries


patriksvensson:
We should sign all binaries with our code certificate.

devlead:
Fixed by #1444

PR #1444, opened 2017-01-11
GH1294: Added signing of artifacts.

patriksvensson:
Signs the following:

* Zip files (for GitHub and Homebrew)
* NuGet packages
* Chocolatey packages

Signing will only occur at a real release.
You can suppress signing by passing the `--skipsigning=true` parameter to the bootstrapper.

Related to #1294


devlead:
@patriksvensson Merged! Thanks for contributing 👍

gep13:
fyi... @onovotny @martinwoodward

onovotny:
Only thought -- you don't really need the `signclient.filter` file as by default it'll try to sign all known/signable file types in the package. That's more for when you want to specifically not sign something, like a 3rd party library included in the package.

devlead:
Got some third parties in the Cake/Cake.CoreCLR packages though.

patriksvensson:
Yes exactly. We don't want to sign third party assemblies in the Cake and Cake.CoreClr packages.

Issue #1424, opened 2016-12-29
Http call in unit test


aateeque:
### What You Are Seeing?
Failing unit test which attempt to make a web call as I am behind corp proxy

### What is Expected?
Unit tests should not make web calls

### What version of Cake are you using?
On the main branch

### Are you running on a 32 or 64 bit system?
64-bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### How Did You Get This To Happen? (Steps to Reproduce)
- Run tests using `dotnet test`
- Failing test: 
` Cake.Common.Tests.Unit.XML.XmlPeekAliasesTests+TheXmlPeekMethod.Should_Get_Node_Value_From_File_With_Dtd [FAIL]` 
- 

### Stack trace
```c#
  Starting:    Cake.Common.Tests
    Cake.Common.Tests.Unit.XML.XmlPokeTests+Transform.Should_Throw_On_Net_Core_Change_Attribute_From_Xml_File_With_Dtd [SKIP]
      .NET Core test.
    Cake.Common.Tests.Unit.XML.XmlPeekAliasesTests+TheXmlPeekMethod.Should_Get_Node_Value_From_File_With_Dtd [FAIL]
      System.Xml.XmlException : An error has occurred while opening external DTD 'http://www.apple.com/DTDs/PropertyList-1.0.dtd': The remote server returned an error: (407) Proxyentication Required.
      ---- System.Net.WebException : The remote server returned an error: (407) Proxy Authentication Required.
      Stack Trace:
           at System.Xml.XmlTextReaderImpl.Throw(Exception e)
           at System.Xml.XmlTextReaderImpl.PushExternalEntityOrSubset(String publicId, String systemId, Uri baseUri, String entityName)
           at System.Xml.XmlTextReaderImpl.DtdParserProxy_PushExternalSubset(String systemId, String publicId)
           at System.Xml.DtdParser.ParseExternalSubset()
           at System.Xml.DtdParser.ParseInDocumentDtd(Boolean saveInternalSubset)
           at System.Xml.DtdParser.Parse(Boolean saveInternalSubset)
           at System.Xml.DtdParser.System.Xml.IDtdParser.ParseInternalDtd(IDtdParserAdapter adapter, Boolean saveInternalSubset)
           at System.Xml.XmlTextReaderImpl.ParseDtd()
           at System.Xml.XmlTextReaderImpl.ParseDoctypeDecl()
[...]

PR #1452, opened 2017-01-12
GH1424: Store & Load DTD locally

aateeque:
This prevents network call in `XmlDocucument.Load` to load the DTD


devlead:
relates to #1424

devlead:
@aateeque merged! Thanks for contributing 👍

PR #1456, opened 2017-01-14
GH1455: Add log string & object methods

### LINKED ISSUE #1455: Logging throws exception when there are curly braces in the string
### What You Are Seeing?
The following line causes an exception saying "Input string was not in a correct format."
`Information("{Bon}jour");`

Output
```
Dummy-Task

An error occurred when executing task 'Dummy-Task'.
Error: Input string was not in a correct format.
```

### What is Expected?
Intuitively, I would expect the logging functions to not call into String.Format if the second argument is not passed in. Unlike the line of code above, the following doesn't cause any issues.
`Console.WriteLine("{Bon}jour");`

Version 0.17.0-beta.1+0.Branch.release/0.17.0.Sha.edc3cfda2d70f30acd59ba2092162ebe83482588

64 bit

Windows

No

### Output Log

GIST LINK - Please create a gist and link to that gist here

OR

~~~sh
PLACE LOG CONTENT HERE
~~~

devlead:
Enables to log just an object and if string bypass string format allowing any string to be logged. 

* fixes #1455 `Information("{Bon}jour");`


gep13:
@devlead your changes have now been merged, thank you!

PR #1464, opened 2017-01-19
GH1463: Add possible nuget path on Mac to tool resolver

### LINKED ISSUE #1463: NuGet Tool Locator system paths on mac need updating
### What You Are Seeing?
It looks like mono stopped creating `/usr/local/bin/nuget` at some point in new installs.  If the symlink is broken, it may cause errors in cake, although removing the symlink altogether seems to be ok as long as the nuget.exe is locatable in ./tools/ or some other known location.

We should probably add the 'new' known location for `nuget` to the list of possible system paths on mac:
https://github.com/cake-build/cake/blob/08907d1a5d97b66f58c01ae82506280882dcfacc/src/Cake.Core/IO/NuGet/NuGetToolResolver.cs#L24-L28

I'd suggest adding: `/Library/Frameworks/Mono.framework/Versions/Current/Commands/nuget` to the beginning of the list so it's searched for first.

0.17.0

64

Mac

Redth:
On Mac, the nuget command alias should exist in `/Library/Frameworks/Mono.framework/Versions/Current/Commands/nuget` and so we should check there first if we are going to look through the possible system paths nuget might be found in.

This fixes #1463

### DIFF
--- a/src/Cake.Core.Tests/Unit/IO/NuGet/NuGetToolResolverTests.cs
+++ b/src/Cake.Core.Tests/Unit/IO/NuGet/NuGetToolResolverTests.cs
@@ -117,6 +117,7 @@ namespace Cake.Core.Tests.Unit.IO.NuGet
             }
 
             [Theory]
+            [InlineData("/Library/Frameworks/Mono.framework/Versions/Current/Commands/nuget")]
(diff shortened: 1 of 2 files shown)


devlead:
@Redth merged,  thanks for contributing 👍

PR #1465, opened 2017-01-19
GH1469: Added a more reliable method to detect Mac OSX

Redth:
Previously on mono, the PlatformID enum was being used to detect OSX.  As of mono 2.2+ the value on Mac is always returned as 4, which makes it impossible to use to distinguish from any other unix based system.

The fix uses a method that’s tried and tested in MonoDevelop (Xamarin Studio & Visual Studio for Mac) calling out to `uname` and checking if the value is Darwin.

### DIFF
--- a/src/Cake.Core/Polyfill/EnvironmentHelper.cs
+++ b/src/Cake.Core/Polyfill/EnvironmentHelper.cs
@@ -14,6 +14,10 @@ namespace Cake.Core.Polyfill
 {
     internal class EnvironmentHelper
     {
+#if !NETCORE
+        private static bool? _isRunningOnMac;
+#endif
+
         public static bool Is64BitOperativeSystem()
         {
 #if NETCORE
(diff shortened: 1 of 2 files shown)


gep13:
@Redth Is there an open issue associated with this PR?  Thanks!

Redth:
@gep13 added issue #1465 to track this.

gep13:
Relates to https://github.com/cake-build/cake/issues/1469

src/Cake.Core/Polyfill/EnvironmentHelper.cs
```diff
@@ -40,8 +40,9 @@ public static PlatformFamily GetPlatformFamily()
                 return PlatformFamily.Windows;
             }
 #else
+            var isRunningOnMac = Native.MacOSX.IsRunningOnMac();
             var platform = (int)Environment.OSVersion.Platform;
-            if (platform == (int)PlatformID.MacOSX)
+            if (isRunningOnMac || platform == (int)PlatformID.MacOSX)
```
patriksvensson: Would it be possible to add this line last in the checks to avoid the call on Windows and Linux? Otherwise, would it be possible to cache the result.

Also, are we sure this won't throw on Windows and Linux?

src/Cake.Core/Polyfill/MacOSXNative.cs
```diff
@@ -0,0 +1,55 @@
...
+                    if (uname(buf) == 0)
+                    {
+                        string os = Marshal.PtrToStringAnsi(buf);
+                        if (os == "Darwin")
+                        {
+                            return true;
+                        }
+                    }
+                }
+                finally
```
patriksvensson: Perhaps we should catch any thrown exceptions here and return false, just in case.

Redth:
@patriksvensson I've added a try/catch around the pinvoke code so if anything manages to throw (like the cleanup) it should be caught.  

I've also moved the windows detection first, so the mac code should never be hit on windows.  We have to still check it before the linux detection code, since mac will be detected as linux (it returns PlatformID of 4 on mac which will be seen as linux).  I have added a cache of the mac detection result though now too.

src/Cake.Core/Polyfill/EnvironmentHelper.cs
```diff
@@ -14,6 +14,8 @@ namespace Cake.Core.Polyfill
 {
     internal class EnvironmentHelper
     {
+        private static bool? _isRunningOnMac;
```
devlead: This should be
 ```diff
+#if !NETCORE
         private static bool? _isRunningOnMac;
+#endif
```

Redth: Fixed this.

devlead:
@Redth merged, thanks for contributing 👍

Issue #1466, opened 2017-01-23
NUnitSettings does not have X86 property


flcdrg:
Unlike NUnit3Settings, NUnitSettings does not have an `X86` property, so there is no way to specify that NUnit tests should run in x86 mode.

gep13:
@flcdrg are you in a position to submit a PR to correct this?

For now, you should be able to use ArgumentCustomization to add the required additional parameters to the tool.

patriksvensson:
@flcdrg There is one: http://cakebuild.net/api/Cake.Common.Tools.NUnit/NUnit3Settings/B4C75865

gep13:
@patriksvensson that is `NUnit3Settings`.  The issue here is for `NUnitSettings`

patriksvensson:
@gep13 Ah, I misread.

flcdrg:
I probably can.

Yes, I since noticed ArgumentCustomization as another option, and also
found some comments on Gitter about someone using the NUnit3 with the
Driver package.

flcdrg:
@gep13 Actually using ArgumentCustomization wouldn't work for this, as you need to run a different test runner, rather than just pass in different arguments.

PR submitted, comments welcome :-)

devlead:
Fixed by #1468

PR #1468, opened 2017-01-24
GH1466: Support X86 property for NUnit

flcdrg:
- Implements #1466


devlead:
@flcdrg merged! Thanks for contributing 👍

Issue #1471, opened 2017-01-26
NuGetToolResolver does not use the environmental override by default


Texernie:
### What You Are Seeing?
When the NUGET_EXE environment variable is set and cake runs an extension like NuGetPack, the NuGetToolResolver class may not return the expected executable. This could cause a buggy nuget or a nuget that doesn't support the required features to be used.

### What is Expected?
The path set in the environment variable to be returned by default.

### What version of Cake are you using?
0.17.0

### Are you running on a 32 or 64 bit system?
64 bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
Originally found on TeamCity, but can reproduce it on the desktop.

### How Did You Get This To Happen? (Steps to Reproduce)

[Cake.zip](https://github.com/cake-build/cake/files/733356/Cake.zip)

The standard build.ps1 has been modified a bit so that a nuget path can be passed in. 
Make sure that there are 2 different copies of nuget.exe somewhere in the system.

.\build.ps1 -Script .\test.cake -NugetPath "C:\DevSVN\ACR-Play\Production\.nuget\"

### Output Log
[Cake_Output.txt](https://github.com/cake-build/cake/files/733363/Cake_Output.txt)

Texernie:
Commit a94d9ae7 created in fork https://github.com/Texernie/cake for this.

PR #1474, opened 2017-01-26
GH1471: Modified the order of resolution of the NuGet tool in the NuGetToolResolver class

Texernie:
This will fix issue #1471.

I simply changed the NuGetToolResolver class to prefer the environment's override if available.

### DIFF
--- a/src/Cake.Core/IO/NuGet/NuGetToolResolver.cs
+++ b/src/Cake.Core/IO/NuGet/NuGetToolResolver.cs
@@ -66,18 +66,6 @@ namespace Cake.Core.IO.NuGet
                 return _cachedPath.Path;
             }
 
-            // Try to resolve it with the regular tool resolver.
-            var toolsExe = _tools.Resolve("nuget.exe");
-            if (toolsExe != null)
-            {
-                var toolsFile = _fileSystem.GetFile(toolsExe);
-                if (toolsFile.Exists)
-                {
-                    _cachedPath = toolsFile;
-                    return _cachedPath.Path;
-                }
-            }
-
             // Check if path set to environment variable
             var environmentExe = _environment.GetEnvironmentVariable("NUGET_EXE");
             if (!string.IsNullOrWhiteSpace(environmentExe))
@@ -86,6 +74,18 @@ namespace Cake.Core.IO.NuGet
                 if (envFile.Exists)
                 {
                     _cachedPath = envFile;
+                    return _cachedPath.Path;
+                }
+            }
+
+            // Try to resolve it with the regular tool resolver.
+            var toolsExe = _tools.Resolve("nuget.exe");
+            if (toolsExe != null)
+            {
+                var toolsFile = _fileSystem.GetFile(toolsExe);
+                if (toolsFile.Exists)
+                {
+                    _cachedPath = toolsFile;
                     return _cachedPath.Path;
                 }
             }


devlead:
Hi @Texernie,
Thanks for contributing to Cake 👍 
We've discussed this internally within @cake-build/team and think changing tool ordering precedence for NuGet in this case  brings more risk than benefit.
So unfortunately I'll not accept the proposed PR.

That said as a work around it's still possible to solve your use case within Cake by in your script setup task register the path from the environment variable, that will override the default tool resolution behavior for all aliases using the nuget.exe.
Example below:
```cake
Setup(
    context => {
    var nugetExe = EnvironmentVariable("NUGET_EXE");
    if (!string.IsNullOrEmpty(nugetExe))
    {
        context.Tools.RegisterFile(nugetExe);
    }
});


Task("My-NuGetTask")
    .Does(()=> {
    Information("{0}", Context.Tools.Resolve("nuget.exe"));
});


RunTarget("My-NuGetTask");
```

Yet again, thanks for taking the time to contribute to Cake, all contributions and feedback is appreciated.

Texernie:
If I may inquire then, what is the purpose of saving the nuget.exe path to environment to be available to child processes in the bootstrap https://raw.githubusercontent.com/cake-build/resources/master/build.ps1 if it will not be used by the tools inside of Cake? Why save it to the environment if the filesystem will override the setting?

devlead:
Unified tool resolution wasn't avail when script was initially written.

Texernie:
The environment pull was added in https://github.com/cake-build/cake/commit/b4f2fd3ac30bc8fae9a72aa6454a2b6bdd8746f8 when the whole class was rewritten for the new resolution scheme. So the feature was added after INuGetToolResolver was added?

Texernie:
I'm not necessarily trying to be difficult, I am just trying to understand why the file system takes priority at the whim of whichever file it happens to find over a setting described exactly for this purpose.

Texernie:
As an end user, when I see a setting in the bootstrap file with the comment of "Save nuget.exe path to environment to be available to child processed", I would expect that executable to be used as a priority.

devlead:
No worries, appreciate the feedback.

Yes but if registered using `context.Tools.RegisterFile(nugetExe);` or via tool override in your script you would expect that to be used.

NUGET_EXE is a convention used by some tools, Cake supports but it's unified way to look for tools is
1. Use explicitly set tool path on [ToolSettings](http://cakebuild.net/api/Cake.Core.Tooling/ToolSettings) [ToolPath](http://cakebuild.net/api/Cake.Core.Tooling/ToolSettings/F6956DA8) property.
2. Via [IToolLocator](http://cakebuild.net/api/Cake.Core.Tooling/IToolLocator/) registered path (see example above).
3. Try to find the tool by searching `./tools/**/mytool.exe`.
4. Look through all paths in the `PATH` environment variable and look for `mytool.exe`.
5. Ask the tool wrapper for additional, alternative paths.

Main reason is though that the risk if regression is too big, as many use custom or no bootstrappers with their CI.

Tool resolution is documented at http://cakebuild.net/docs/tools/tool-resolution , but I'm going to look att improving that using the feedback given here.

Texernie:
Mattias, I totally understand the regression test issue. You have given me a work around that will work at least. That work around should probably be included in the documentation somewhere. I also understand your comment that if context.Tools.RegisterFile was used that the specific file registered would be expected to be used instead of a registry setting. Thank you for taking the time to explain the other ramifications and providing a work around.

PR #1508, opened 2017-02-23
[HOLD] Add IsDependentOnIfExists extension

### LINKED ISSUE #1507: Add support for dependencies to "maybe-existing" targets
Currently the `IsDependentOn` extension requires that the depentend on task exists. While writing Cake scripts which are shared among different projects it can be usefule to make a task in the shared script dependent on some other task, which serves as extension point for the concrete Cake script using the shared script.

pascalberger:
Add `IsDependentOnIfExists` extension.

Fixes #1507

### DIFF
--- a/src/Cake.Core/ICakeTaskInfo.cs
+++ b/src/Cake.Core/ICakeTaskInfo.cs
@@ -27,6 +27,6 @@ namespace Cake.Core
         /// Gets the task's dependencies.
         /// </summary>
         /// <value>The task's dependencies.</value>
-        IReadOnlyList<string> Dependencies { get; }
+        IReadOnlyList<CakeTaskDependency> Dependencies { get; }
     }
 }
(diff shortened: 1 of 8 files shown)


src/Cake.Core/ICakeTaskInfo.cs
```diff
@@ -27,6 +27,6 @@ public interface ICakeTaskInfo
         /// Gets the task's dependencies.
         /// </summary>
         /// <value>The task's dependencies.</value>
-        IReadOnlyList<string> Dependencies { get; }
+        IReadOnlyList<CakeTaskDependency> Dependencies { get; }
```
devlead: This is a breaking change, for any script / module / addin that uses/implements this interface.

Benefit needs to be evaluated by @cake-build/team

patriksvensson: Yes, I agree. There are some things planned for 0.19 that addresses task dependencies, so we should wait with this until we've had the chance to have a proper discussion about that as well before introducing something new.

gep13:
@pascalberger have you seen the recent PR that @patriksvensson did regarding IsDependeeOf?  Does that cover the requirements that were included in this PR?

pascalberger:
@gep13 Yes, seems so

https://github.com/cake-build/cake/pull/1819 is a better approach to tackle this and therefore closing this PR

gep13:
👍

PR #1519, opened 2017-03-04
GH1518: Adding BuildTools as a version to the MSBuildResolver.

### LINKED ISSUE #1518: Add "build tools" path for MSBuild 2017 to MSBuildResolver
This was also discussed in #1369 , MSBuild added a stand alone product separate from Visual Studio that installs in a location other than Enterprise, Professional, or Community.  You can assign this to me and I will get it added.

phillipsj:
I have added it to the array as an edition. Fixes #1518

### DIFF
--- a/src/Cake.Common/Tools/MSBuild/MSBuildResolver.cs
+++ b/src/Cake.Common/Tools/MSBuild/MSBuildResolver.cs
@@ -95,7 +95,8 @@ namespace Cake.Common.Tools.MSBuild
             {
                 "Enterprise",
                 "Professional",
-                "Community"
+                "Community",
+                "BuildTools"
             };
 
             var visualStudio2017Path = environment.GetSpecialPath(SpecialPath.ProgramFilesX86);


src/Cake.Common/Tools/MSBuild/MSBuildResolver.cs
```diff
@@ -95,7 +95,8 @@ private static DirectoryPath GetVisualStudioPath(ICakeEnvironment environment, M
             {
                 "Enterprise",
                 "Professional",
-                "Community"
+                "Community",
+                "BuildTools"
             };
 
             var visualStudio2017Path = environment.GetSpecialPath(SpecialPath.ProgramFilesX86);
```
Trass3r: This does not work with custom VS installation paths.
We use this:
```
DirectoryPath vsLatest  = VSWhereLatest();
FilePath msBuildPath = (vsLatest == null) ? null : vsLatest.CombineWithFilePath("./MSBuild/15.0/Bin/amd64/MSBuild.exe");
```

devlead: Correct.

gep13: @Trass3r yes, out of the box, Cake handles the default installation paths, and for anything else, there is the VSWhere Tool, which can then be used to set ToolPath on tools that require it, as described here:

http://cakebuild.net/blog/2017/03/vswhere-and-visual-studio-2017-support

Issue #1299, opened 2016-10-19
ProjectParser fails with content including wildcards


kcamp:
### What You Are Seeing?

Cake.Common.Solution.Project.ProjectParser fails when the *.csproj contains wildcards in a <Content Include="" /> block
### What is Expected?

Cake should not throw an exception.  

Not sure what I might expect with regard to the results.  Attempting to glob/resolve all files included in the pattern might be useful, but not throwing an exception would be the minimum viable
### What version of Cake are you using?

0.16.1
### Are you running on a 32 or 64 bit system?

64
### What environment are you running on?  Windows? Linux? Mac?

Windows
### Are you running on a CI Server?  If so, which one?

No
### How Did You Get This To Happen? (Steps to Reproduce)

Attempted a `Parse(...)` call on any  *.csproj containing the following:

```
<Content Include="app\**" />
```

I have written a failing test here -
https://github.com/kcamp/cake/commit/140e21728a751a808017f9994cc5126ba461295a
### Output Log

``` sh
An error occured when executing task 'Report-Project-References'.
Error: System.ArgumentException: Illegal characters in directory path (*).
Parameter name: path
   at Cake.Core.IO.Path..ctor(String path)
   at Cake.Core.IO.FilePath..ctor(String path)
   at Cake.Core.IO.FilePath.FromString(String path)
   at Cake.Core.IO.FilePath.op_Implicit(String path)
   at Cake.Common.Solution.Project.ProjectParser.<>c__DisplayClass3_1.<Parse>b__27(<>f__AnonymousType8`2 <>h__TransparentIdentifier3)
   at System.Linq.Enumerable.<>c__DisplayClass7_0`3.<CombineSelectors>b__0(TSource x)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Cake.Common.Solution.Project.ProjectParser.Parse(FilePath projectPath)
   at Namespace.ProjectReportingParser.ParseProject(FilePath projectPath)
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Namespace.ProjectReportingParser.ParseAll(String pattern)
[...]

PR #1529, opened 2017-03-09
[DO NOT MERGE] GH1299: ProjectParser - Fix "Illegal characters in directory path (*)" exception 

kcamp:
DO NOT MERGE.  I think this is worth further review/discussion on the implementation, but this is a start.

This addresses the "Illegal characters in directory path (*)" exception when a *.csproj contains wildcard `Include` attributes, as detailed in GH-1299

The issue, as originally filed, was easy enough to side-step in my own project(s) that were impacted.

I have found another project where we're less able to work around it - the project is auto-generated via the Swagger CodeGen tool and would represent significant burden in attempting to maintain the file beyond what the CodeGen tool does automatically, particularly in light of the fact that it's perfectly valid MSBuild syntax/content.

The couple of concerns that I can think of off the top of my head are this:

1.  This is still a semi-naive solution in that it won't consider globbing patterns that may exist outside of a project's working directory.  i.e., `<Compile Include="..\..\**\*.cs" />` or similar.

2. The globbing implementation may be better suited with a different approach.  I started to consider how to create a separate class/implementation that instantiates a new `Globber` with a copy of `CakeEnvironment` with the `WorkingDirectory` being overridden, but that proved more difficult than was warranted for a proof-of-concept with regard to testing concerns.  However, in hindsight, I think this is probably the right thing to do.

Consider the two main cases in which we can encounter this:
`<Compile Include="**\*.cs" />`
`<Compile Include="..\..\fake\**\*.cs" />`

I have confirmed (for my own self)  that the `Globber` implementation is able to handle directory traversals, i.e., `GetFiles("./../**/*.csproj")`
So if I were to mutate the implementation I have to be more correct in positioning the globber's working directory to that of the current project, that should create the correct scenario where we're able to resolve _all_ of the files whether they ancestors or descendants of the working directory.

So, a couple of questions:
1. Does the second option (above) sound more correct?  I think it does.   Is there a better way to deal with overridding the working directory for the consideration of the Globber itself?  It seems like it's pretty tightly coupled to the `ICakeEnvironment.WorkingDirectory` at the moment.  If there are suggestions, I'd like to hear them.

2. Is this change something we wish to pursue?  I think it's pretty close and has at least identified the deficiencies, it's just a matter of refactoring to cover the edge cases.  If it's not, we should at least mute the exception -or- provide an `Options` object (or parameter) that would turn off the parsing of the files, which would achieve the same.


kcamp:
I've refactored the implementation to be more inline with my idea # 2 ("The globbing implementation may be better suited with a different approach.", above)

This handles processing `Include` attributes correctly with test coverage for the general cases:
`<Compile Include="**\*.cs" />`
`<Compile Include="..\..\fake\**\*.cs" />`

This seems like a more correct approach.  Let me know what you think about this entire line of dev.

gep13:
@kcamp after a review of this PR by the @cake-build/cake-team we feel that this PR is no longer required.  As such, we are going to close this.  Please feel free to open this again, if you want to discuss this further.

Issue #1533, opened 2017-03-16
Update DotNetCore Aliases to match tooling 1.0


JonCubed:
With the release of VS 2017,  .NET Core Tooling has been bumped to 1.0. This has introduced a new build system (move from project.json to csproj and msbuild), changes to existing commands and added additional commands. It would be nice for the DotNetCore Aliases to reflect the changes, making it easier for ppl to migrate over to the new tooling.

I asked in gitter and @gep13 said you would be willing to accept a PR on this and I am happy to put some time in to do it. However I have some questions before starting

1. Given that VS 2015 wont use v1.0 of the tooling should we continue to support Project.Json version of the tooling for the time being?
2. If we support both we will need a new set of alias to distinguish, any thoughts on how you prefer for this to be? Does the new tooling become `DotNetCore*` and old become `DotNetCoreJson*` creating a breaking change or keep current as is and new Aliases called something like `DotNetCoreMSBuild*`?
3. Would it be easier for review process if I split this into multiple PRs?

**Tasks:**
- [x] Common Changes - ~#1549~
- [x] Add DotNetCoreClean - ~#1555~
- [x] Add DotNetCoreNugetPush - ~#1581~
- [x] Add DotNetCoreNugetDelete - ~#1565~ 
- [X] Add DotNetCoreMSBuild - ~#1599~
- [x] Add DotNetCoreVSTest - ~#1675~
- [x] Review DotNetCoreRestore - ~#1534~
- [x] Review DotNetCoreBuild - ~#1550~
- [x] Review DotNetCorePublish - ~#1552~
- [x] Review DotNetCoreTest - ~#1554~
- [x] Review DotNetCorePack - ~#1553~
- [x] Review DotNetCoreRun - no change required
- [x] Review DotNetCoreExecute -~#1551~
- [x] Review if any changes required for 1.0.4 of tooling - no change required

gep13:
1. If the "new" way of doing it, is going to be "the" way of doing it, then I would be happy updating the existing aliases to only use that method.  Community Members who require the earlier aliases have the option to pin to an earlier version of Cake.
2. I think this would get messy pretty quickly, and would prefer to not go down this route.
3. Possibly yes, depending on how much changes

@cake-build/team thoughts?

patriksvensson:
@gep13 @JonCubed

1. Agree with @gep13 
2. Agree with @gep13 
3. I think it makes sense to break each tool down to a separate PR since it will be very clear in the commit log and also the release notes.

[...]

PR #1534, opened 2017-03-16
GH1533: Updated DotNetCoreRestore inline with Tooling v1.0

JonCubed:
- Added new arguments as settings and updated restorer
- Removed settings that are no longer valid, if you prefer for them to be marked as obsolete instead let me know.
- updated unit tests, removed unnecessary test

See #1533 for more info


devlead:
@JonCubed  your changes have been merged, thanks for your contribution 👍

PR #1550, opened 2017-03-29
GH1533: Updated DotNetCoreBuild inline with Tooling v1.0

JonCubed:
* Removed settings that are no longer valid, if you prefer for them to be marked as obsolete instead let me know.
* Updated unit tests

See #1533 for more info

### DIFF
--- a/src/Cake.Common/Tools/DotNetCore/Build/DotNetCoreBuildSettings.cs
+++ b/src/Cake.Common/Tools/DotNetCore/Build/DotNetCoreBuildSettings.cs
@@ -12,11 +12,6 @@ namespace Cake.Common.Tools.DotNetCore.Build
     public sealed class DotNetCoreBuildSettings : DotNetCoreSettings
     {
         /// <summary>
-        /// Gets or sets the directory in which to place temporary outputs.
-        /// </summary>
-        public DirectoryPath BuildBasePath { get; set; }
-
-        /// <summary>
         /// Gets or sets the output directory.
         /// </summary>
         public DirectoryPath OutputDirectory { get; set; }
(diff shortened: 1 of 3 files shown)


src/Cake.Common/Tools/DotNetCore/Build/DotNetCoreBuildSettings.cs
```diff
@@ -42,11 +37,6 @@ public sealed class DotNetCoreBuildSettings : DotNetCoreSettings
         public string VersionSuffix { get; set; }
```
devlead: We should add a string Version { get; set; } too, equivalent to
```csharp
 ArgumentCustomization = arg=>arg.Append(string.Concat("/p:Version=", semVersion)),
``` 

We should also add an 
```csharp
public IDictionary<string, IList<string>> Properties => _properties;
```
To support MSBuild properties without ArgumentCustomization, similar too [MSBuildSettings.Properties](https://github.com/cake-build/cake/blob/1eb28cf18637a1316c8f10327b9d70b9975507ea/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs#L32)

We should also add an 
```csharp
public ISet<string> Targets { get; }
```

Running `dotnet msbuild -h` looks like it support most MSBuild so future refactoring could be to see what we have common between the .NET Core CLI and MSBuild alias.

devlead: Maybe version could be something common, know `dotnet pack` accepts & should have it too (haven't all yet but probably publish too).

patriksvensson: @devlead Should we really add `Version` now? Isn't it better to wait until the API settled a little bit? I would rather see MSBuild properties first.

If we add version we should add support for `AssemblyVersion` and `FileVersion` as well.

JonCubed: I was planning to add `VersionPrefix` after all the current tasks are done, along with `Properties`. I don't think we should use `Version`, I think it gets overridden in some situations unexpectedly and doesn't work with `VersionSuffix`.  

@patriksvensson if my memory is correct, when I was testing using `VersionPrefix` with the build command acted as `AssemblyVersion`

devlead:
@JonCubed  your changes have been merged, thanks for your contribution 👍

PR #1560, opened 2017-04-02
GH1557: Add support for MSBuild console logger parameters

### LINKED ISSUE #1557: Add support for MSBuild /consoleloggerparameters
There is currently no way in `MSBuildSettings` to set the `/consoleloggerparameters` command line switch (see [MSBuild Command Line Reference](https://msdn.microsoft.com/en-us/library/ms164311.aspx)). This parameter is useful, for instance, to force color output on Jenkins (`/consoleloggerparameters:ForceConsoleColor` - this specific parameter is not mentioned in the reference page, but was [added about a year ago](https://github.com/Microsoft/msbuild/pull/505)).

thomaslevesque:
Fixes #1557

Note: I took a straightforward approach (just a `ISet<string>` on `MSBuildSettings`), but then I realized that it should probably be made more similar to file logger parameters (which has a `MSBuildFileLogger` class with bool properties for each parameter). Let me know if I should refactor (and how: for instance, should I extract a common base class between `MSBuildFileLogger` and `MSBuildConsoleLogger`?)

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
@@ -1040,6 +1040,26 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
                 // Then
                 Assert.Equal(expected, result.Args);
             }
+
+            [Theory]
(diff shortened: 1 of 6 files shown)


devlead:
@thomaslevesque sorry for the late reply, this PR has conflicts with the current development branch, could you please rebase and resolve against latest develop.

devlead:
@thomaslevesque  your changes have been merged, thanks for your contribution 👍

PR #1563, opened 2017-04-04
Update OSX Run method

HustLion:
The OSX instructions are not working. Added a working one. The `./build.sh` doesn't work with osx, need the osx specific one. And the `tools` folder in the repo causes `Error: An item with the same key has already been added.` even if I use osx specific script. I highly suggest add `tools/` to `.gitignore`

Since this is a trivial build guidance change, no issue has been opened, if needed please comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ RunTarget(target);
 ./build.ps1
 ```
 
-##### Linux / OS X
+##### Linux
 
 ```console
 # Adjust the permissions for the bootstrapper script.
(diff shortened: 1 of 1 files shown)


gep13:
>@HustLion said...
The ./build.sh doesn't work with osx, need the osx specific one.

I am slightly confused by this.  Although we have separate URL's for OS x and Linux, the URL's actually resolve the to exact same file, which is this one:

https://raw.githubusercontent.com/cake-build/resources/master/build.sh

The bootstrapper also is responsible for downloading Cake, so I don't see any need to globally install Cake using homebrew.

gep13:
@HustLion Did you have any comments about the review that I did above?  Were you in a position to update the PR?  Thanks for your help!

HustLion:
@gep13 Thanks for your review! Sorry that I didn't reply earlier. Since you pointed out _the URL's actually resolve the to exact same file_, I think the issue might be in my operations. Currently I'm not using Cake so I can't really confirm the source of the issue. Closing this now. If later I find out that the issue comes up again, I will provide more info.

Issue #1564, opened 2017-04-05
DeleteDirectory cannot delete read-only files


DixonDs:
I would like to have a possibility to remove read-only files either by default or as another option to `DeleteDirectory` alias

PR #1574, opened 2017-04-18
GH1564: Allow deletion of read-only files

DixonDs:
Fixes https://github.com/cake-build/cake/issues/1564

### DIFF
--- a/src/Cake.Common/IO/DeleteDirectorySettings.cs
+++ b/src/Cake.Common/IO/DeleteDirectorySettings.cs
@@ -1,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Cake.Common.IO
+{
+    /// <summary>
+    /// Contains settings used by <c>DeleteDirectory</c>.
+    /// </summary>
+    public class DeleteDirectorySettings
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether to perform a recursive delete if set to <c>true</c>.
+        /// <remarks>
+        /// It is set to <c>false</c> by default.
+        /// </remarks>
+        /// </summary>
+        public bool Recursive { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to delete read-only files if set to <c>true</c>.
+        /// <remarks>
+        /// It is set to <c>false</c> by default.
+        /// </remarks>
+        /// </summary>
+        public bool Force { get; set; }
+    }
+}
--- a/src/Cake.Common/IO/DirectoryAliases.cs
+++ b/src/Cake.Common/IO/DirectoryAliases.cs
@@ -65,6 +65,7 @@ namespace Cake.Common.IO
         /// <param name="context">The context.</param>
         /// <param name="directories">The directory paths.</param>
         /// <param name="recursive">Will perform a recursive delete if set to <c>true</c>.</param>
+        [Obsolete("Use the overload that accepts DeleteDirectorySettings instead.")]
         [CakeMethodAlias]
         [CakeAliasCategory("Delete")]
         public static void DeleteDirectories(this ICakeContext context, IEnumerable<DirectoryPath> directories, bool recursive = false)
@@ -74,9 +75,39 @@ namespace Cake.Common.IO
                 throw new ArgumentNullException(nameof(directories));
             }
 
+            DeleteDirectories(context, directories, new DeleteDirectorySettings { Recursive = recursive });
+        }
+
+        /// <summary>
+        /// Deletes the specified directories.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// var directoriesToDelete = new DirectoryPath[]{
+        ///     Directory("be"),
+        ///     Directory("gone")
+        /// };
+        /// DeleteDirectories(directoriesToDelete, new DeleteDirectorySettings {
+        ///     Recursive = true,
+        ///     Force = true
+        /// });
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="directories">The directory paths.</param>
+        /// <param name="settings">The delete settings.</param>
+        [CakeMethodAlias]
+        [CakeAliasCategory("Delete")]
+        public static void DeleteDirectories(this ICakeContext context, IEnumerable<DirectoryPath> directories, DeleteDirectorySettings settings)
+        {
+            if (directories == null)
+            {
+                throw new ArgumentNullException(nameof(directories));
+            }
+
             foreach (var directory in directories)
             {
-                DeleteDirectory(context, directory, recursive);
+                DeleteDirectory(context, directory, settings);
             }
         }
 
@@ -95,9 +126,35 @@ namespace Cake.Common.IO
         /// <param name="context">The context.</param>
         /// <param name="directories">The directory paths.</param>
         /// <param name="recursive">Will perform a recursive delete if set to <c>true</c>.</param>
+        [Obsolete("Use the overload that accepts DeleteDirectorySettings instead.")]
         [CakeMethodAlias]
         [CakeAliasCategory("Delete")]
         public static void DeleteDirectories(this ICakeContext context, IEnumerable<string> directories, bool recursive = false)
+        {
+            DeleteDirectories(context, directories, new DeleteDirectorySettings { Recursive = recursive });
+        }
+
+        /// <summary>
+        /// Deletes the specified directories.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// var directoriesToDelete = new []{
+        ///     "be",
+        ///     "gone"
+        /// };
+        /// DeleteDirectories(directoriesToDelete, new DeleteDirectorySettings {
+        ///     Recursive = true,
+        ///     Force = true
+        /// });
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="directories">The directory paths.</param>
+        /// <param name="settings">The delete settings.</param>
+        [CakeMethodAlias]
+        [CakeAliasCategory("Delete")]
+        public static void DeleteDirectories(this ICakeContext context, IEnumerable<string> directories, DeleteDirectorySettings settings)
         {
             if (directories == null)
             {
@@ -107,7 +164,7 @@ namespace Cake.Common.IO
             var paths = directories.Select(p => new DirectoryPath(p));
             foreach (var directory in paths)
             {
-                DeleteDirectory(context, directory, recursive);
+                DeleteDirectory(context, directory, settings);
             }
         }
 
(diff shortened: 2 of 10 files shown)


mholo65:
@DixonDs 
Thanks for your PR.

We have been discussing this internally and would rather see this behavior as opt-in. What do you think of creating overloads for all ```DeleteDirectory``` aliases?

Preferred solution could be:
```csharp
DeleteDirectories(ICakeContext, IEnumerable<DirectoryPath>, DeleteDirectorySettings)
DeleteDirectories(ICakeContext, IEnumerable<string>, DeleteDirectorySettings)
DeleteDirectory(ICakeContext, DirectoryPath, DeleteDirectorySettings)
```
Where ```DeleteDirectorySettings``` could look like this:
```csharp
public class DeleteDirectorySettings
{
   public bool Recursive { get; set; }
   public bool Force { get; set; }
}
```

toddstewart:
I like the idea of the Force property but you have to take into mind files that are in use.  I have run into where a file in the packages directory is in use and cannot be deleted.

toddstewart:
Yes I agree with opt-in but if the property is called Force it does not indicate it is only to delete read only files.  Just wanted to point that out.

mholo65:
@toddstewart sorry, didn't see your comment. Just submitted a review comment 😄 

Is there any better name to use than `force`?  Force is quite commonly used in same situtations.. Even though it's called force, it might fail miserably if a) user doesn't have rights to remove files/directories, b) files/directories are in use c) anything else..

devlead:
@mholo65 @toddstewart I think `Force` is an excellent name.

DixonDs:
@mholo65 I've added overloads with `DeleteDirectorySettings`, but removed the ones with `bool recursive` which is a breaking change. Let me know if I need to put back those overloads with the boolean `recursive` flag.

src/Cake.Core/IO/IDirectory.cs
```diff
@@ -32,7 +32,8 @@ public interface IDirectory : IFileSystemInfo
         /// Deletes the directory.
         /// </summary>
         /// <param name="recursive">Will perform a recursive delete if set to <c>true</c>.</param>
-        void Delete(bool recursive);
+        /// <param name="force">Will delete read-only files if set to <c>true</c></param>
+        void Delete(bool recursive, bool force);
```
mholo65: This is also a breaking change.

mholo65: @DixonDs yes, exactly.
Just wondering if the attribute modification should be done elsewhere? E.g. introduce new interface for that.

Thoughts on this @patriksvensson @devlead

devlead: @mholo65 don't want to litter code base with allot of interfaces either, so one option could be to get this in a milestone/release whilst we have other breaking changes that need a minimum cake version set, i.e. the bigger roslyn refactoring 0.21:isch time frame.

devlead: If doing so that would involve [IFile::Delete()](https://github.com/DixonDs/cake/blob/e515d2408311e5b8b5d0736e33f566f9e7212f11/src/Cake.Core/IO/IFile.cs#L42) too.

DixonDs: @devlead @mholo65 It could be done in internal `DirectoryDeleter` if `IFile` provided access to file attributes but it doesn't. But if exposing getting/setting attributes in `IFile` deserves a breaking change on its own, that could be an alternative way to go.

Please, advise

mholo65: Sounds good to me!

mholo65: @DixonDs `IFile`provides the `Path` so you can create a `System.IO.FileInfo` from that.

DixonDs: @mholo65 Doesn't that defeat the purpose of abstraction? Will`DirectoryAliasesTests` work where the fake file system is used?

DixonDs:
@mholo65 @devlead Could you please review the latest changes?

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -67,7 +67,39 @@ public static ConvertableDirectoryPath Directory(this ICakeContext context, stri
         /// <param name="recursive">Will perform a recursive delete if set to <c>true</c>.</param>
         [CakeMethodAlias]
         [CakeAliasCategory("Delete")]
-        public static void DeleteDirectories(this ICakeContext context, IEnumerable<DirectoryPath> directories, bool recursive = false)
+        public static void DeleteDirectories(this ICakeContext context, IEnumerable<DirectoryPath> directories,
+            bool recursive = false)
```
mholo65: Could you move this back up to the previous line.

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -67,7 +67,39 @@ public static ConvertableDirectoryPath Directory(this ICakeContext context, stri
...
+        {
+            if (directories == null)
```
mholo65: Remove redundant null check. This is also checked in overload.

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -97,7 +129,39 @@ public static void DeleteDirectories(this ICakeContext context, IEnumerable<Dire
         /// <param name="recursive">Will perform a recursive delete if set to <c>true</c>.</param>
         [CakeMethodAlias]
         [CakeAliasCategory("Delete")]
-        public static void DeleteDirectories(this ICakeContext context, IEnumerable<string> directories, bool recursive = false)
+        public static void DeleteDirectories(this ICakeContext context, IEnumerable<string> directories,
+            bool recursive = false)
```
mholo65: Could you move this back up to the previous line.

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -97,7 +129,39 @@ public static void DeleteDirectories(this ICakeContext context, IEnumerable<Dire
...
+        {
+            if (directories == null)
```
mholo65: Remove redundant null check. This is also checked in overload.

src/Cake.Testing/FakeFile.cs
```diff
@@ -158,9 +164,11 @@ private long GetPosition(FileMode fileMode, out bool fileWasCreated)
                     case FileMode.Truncate:
                         Length = 0;
                         return 0;
+
```
mholo65: Remove extra line

src/Cake.Testing/FakeFile.cs
```diff
@@ -158,9 +164,11 @@ private long GetPosition(FileMode fileMode, out bool fileWasCreated)
...
                     case FileMode.Open:
                     case FileMode.OpenOrCreate:
                         return 0;
+
```
mholo65: Remove extra line

src/Cake.Testing/FakeFile.cs
```diff
@@ -176,6 +184,7 @@ private long GetPosition(FileMode fileMode, out bool fileWasCreated)
                         fileWasCreated = true;
                         Exists = true;
                         return Length;
+
```
mholo65: Remove extra line

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -126,7 +190,29 @@ public static void DeleteDirectories(this ICakeContext context, IEnumerable<stri
...
+        /// });
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="path">The directory path.</param>
+        /// <param name="settings">The delete settings.</param>
+        [CakeMethodAlias]
+        [CakeAliasCategory("Delete")]
+        public static void DeleteDirectory(this ICakeContext context, DirectoryPath path,
+            DeleteDirectorySettings settings)
```
mholo65: Move this up one line

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -97,7 +129,39 @@ public static void DeleteDirectories(this ICakeContext context, IEnumerable<Dire
...
+        /// });
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="directories">The directory paths.</param>
+        /// <param name="settings">The delete settings.</param>
+        [CakeMethodAlias]
+        [CakeAliasCategory("Delete")]
+        public static void DeleteDirectories(this ICakeContext context, IEnumerable<string> directories,
+            DeleteDirectorySettings settings)
```
mholo65: Move this up one line

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -67,7 +67,39 @@ public static ConvertableDirectoryPath Directory(this ICakeContext context, stri
...
+        /// });
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="directories">The directory paths.</param>
+        /// <param name="settings">The delete settings.</param>
+        [CakeMethodAlias]
+        [CakeAliasCategory("Delete")]
+        public static void DeleteDirectories(this ICakeContext context, IEnumerable<DirectoryPath> directories,
+            DeleteDirectorySettings settings)
```
mholo65: Move this up one line

DixonDs:
@mholo65 @devlead Could you please review again?

mholo65:
@DixonDs 
> Could you please review again?

Looks good to me. I was thinking if we should obsolote the old aliases at the same time. Thoughts @cake-build/cake-team?

If @devlead or any other in team doesn't have any opinions on this, would you @DixonDs mind [squashing commits](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md#respond-to-feedback-on-pull-request) before we do the final review. Thanks!

DixonDs:
@mholo65 
>would you @DixonDs mind squashing commits before we do the final review

I will do later, but I am just wondering if it makes much sense if you can just complete PR with merge-squash.

gep13:
@DixonDs no, that is not the process that we follow on this repository, as the git timeline isn't as clear.

DixonDs:
@mholo65 Could you please do the final review?

src/Cake.Common/IO/DeleteDirectorySettings.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether to perform a recursive delete if set to <c>true</c>.
+        /// </summary>
+        public bool Recursive { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to delete read-only files if set to <c>true</c>.
+        /// </summary>
+        public bool Force { get; set; }
```
patriksvensson: Perhaps we should add `<remarks>` here that this is `false` by default.

src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -99,6 +129,31 @@ public static void DeleteDirectories(this ICakeContext context, IEnumerable<Dire
         [CakeAliasCategory("Delete")]
         public static void DeleteDirectories(this ICakeContext context, IEnumerable<string> directories, bool recursive = false)
```
mholo65: @Cake-build/team should we obsolete these?

mholo65: Ping @cake-build/cake-team

patriksvensson: @mholo65 Yes, I think we should. Obsolete with a warning message for now.

gep13: @mholo65 I am on the fence on this one.  They are still valid, as they can be easily mapped into the new way of doing things, so part of me says to leave them.  Having said that, would also be happy to see them removed.

mholo65: @gep13 yes, leave them for a release or two. But obsolete with warning :wink:

DixonDs:
@mholo65 done

devlead:
@DixonDs CS0618 needs to be disabled case by case in the test project

i.e.
```csharp
#pragma warning disable CS0618 // Type or member is obsolete
                var result = Record.Exception(() => DirectoryAliases.DeleteDirectory(null, "/Temp/DoNotExist"));
#pragma warning restore CS0618 // Type or member is obsolete
```

devlead:
@DixonDs  your changes have been merged, thanks for your contribution 👍

PR #1586, opened 2017-04-28
WIP: (GH-1561) Add methods for argument checking

### LINKED ISSUE #1561: Add helper for argument validation
Each Cake addin requires some kind of argument validation (or at least should :). Instead of having to implement this in every addin it would be nice if this could be provided for common cases (null check, empty string check, ...) by Cake.Core.

pascalberger:
Adds a class for argument validation as discussed in #1561.

Implemented it myself instead of using the [ASP.NET version](https://github.com/aspnet/EntityFramework/blob/dev/src/Shared/Check.cs) but should provide very similar capabilities.

Fixes #1561

### DIFF
--- a/src/Cake.Core.Tests/Unit/CakeArgumentChecksTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeArgumentChecksTests.cs
@@ -1,0 +1,654 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace Cake.Core.Tests.Unit
+{
+    public sealed class CakeArgumentChecksTests
+    {
+        public sealed class TheNotNullExtension
+        {
+            [Fact]
+            public void Should_Throw_If_Parameter_Is_Null()
+            {
+                // Given
+                const string value = "foo";
+
+                // When
+                var result = Record.Exception(() => value.NotNull(null));
+
+                // Then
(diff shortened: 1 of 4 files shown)


pascalberger:
@mholo65 While making the methods available to Cake scripts makes sense, my main goal was also to provide this functionality to addins which wouldn't work with your proposal

mholo65:
@pascalberger 
> my main goal was also to provide this functionality to addins which wouldn't work with your proposal

I understand your concerns, but at this point we would not like to have this as exposed as `public` functionality in `Cake.Core`. We however like the idea of changing all occurrences of `if (foo == null) { throw ... }` to the more convenient  `_foo = Check.NotNull(foo, nameof(foo))`. Additionally we would also like to expose it as _aliases_.

However, to have this _boilerplate_ code available also for `addin` developers, there are several possibilities:
* We provide a _source only_ NuGet package (`Cake.Shared.Source`) which would contain helpful boilerplate code (like this). Only problem is that as of VS2017, there seem to be some issue regarding tooling and _source only_ NuGet packages (both me and @patriksvensson have failed getting e.g. [XUnit.Assert.Source](https://www.nuget.org/packages/xunit.assert.source) to work in VS2017 with `*.csproj`).
* You include the source in you projects using a tool like [T4Include](https://github.com/mrange/T4Include)
* You download the source manually and include it in your projects.

pascalberger:
@mholo65 OK, but in this case its not my priority to implement this functionality and I'll leave it to someone else to do so.

mholo65:
@pascalberger ok. We are looking into different possibilities on how to expose this kind of functionality for addin-developers. A source only NuGet (once VS2017 tooling supports it) or a support library might be best solution.

PR #1600, opened 2017-05-07
Make runner internal

pascalberger:
Currently all tool runner are public. IMHO they should always be called through aliases and since they are sealed it doesn't bring any benefit to have them public.


patriksvensson:
@pascalberger I don't agree. Sometimes (like in a Cake.Frosting application) you want to instantiate the tool yourself with a different set of dependencies (I've done this myself in some projects).

phillipsj:
That was exactly the use cake I was thinking about too.

pascalberger:
@patriksvensson OK, makes sense

PR #1605, opened 2017-05-15
Fix the contribution-guidelines link again

tom-dudley:
Looks like the link I changed in https://github.com/cake-build/cake/pull/1604 wasn't quite right. This removes the trailing slash which broke the link. Actually tested it this time!


gep13:
Doh! I should have checked this as well!  Thanks for taking the time to fix again!

Issue #1607, opened 2017-05-16
ToolResolutionStrategy fails unexpectedly with Cake.LongPath.Module


kcamp:
### What You Are Seeing?
Cake build fails to install nuget packages via `#load` or `#addin` with a fatal error. `The directory name is invalid`

### What is Expected?
I expect any issues to be sidestepped and the build to complete.

### What version of Cake are you using?
0.19.4

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows8

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)
This is 100% environmental and occurs on a single developer machine.  I cannot repro this or even determine what value (in `%PATH%`) it is failing on in writing some ad hoc unit tests to try to diagnose.

### Output Log
~~~sh
An error occurred when executing task 'Restore-Nuget-Packages'.
Error: System.IO.IOException: The directory name is invalid.

   at Pri.LongPath.Common.ThrowIOError(Int32 errorCode, String maybeFullPath)
   at Pri.LongPath.FileSystemInfo.Refresh()
   at Pri.LongPath.FileInfo.get_Exists()
   at Cake.Core.IO.FileSystemExtensions.Exist(IFileSystem fileSystem, FilePath path)
   at Cake.Core.Tooling.ToolResolutionStrategy.LookInPath(String tool)
   at Cake.Core.Tooling.ToolResolutionStrategy.Resolve(IToolRepository repository, String tool)
   at Cake.Core.Tooling.ToolLocator.Resolve(String tool)
   at Cake.Core.Tooling.Tool`1.GetToolPathUsingToolService(TSettings settings)
   at Cake.Core.Tooling.Tool`1.GetToolPath(TSettings settings)
   at Cake.Core.Tooling.Tool`1.RunProcess(TSettings settings, ProcessArgumentBuilder arguments, ProcessSettings processS
ettings)
   at Cake.Core.Tooling.Tool`1.Run(TSettings settings, ProcessArgumentBuilder arguments, ProcessSettings processSettings
, Action`1 postAction)
   at Cake.Core.Tooling.Tool`1.Run(TSettings settings, ProcessArgumentBuilder arguments)
   at Cake.Common.Tools.NuGet.Restore.NuGetRestorer.Restore(FilePath targetFilePath, NuGetRestoreSettings settings)
[...]

PR #1608, opened 2017-05-17
GH1607: Make ToolResolutionStrategy more resilient to exceptions 

kcamp:
This addresses the failures described in GH-1607 by surrounding the attempt to access the file system (and consequently any `IFileSystem` implementation being registered via module) with a `try/catch` to suppress any errors that may be environmental so as to fully iterate the paths represented in the PATH environment variable.

### DIFF
--- a/src/Cake.Core.Tests/Unit/Tooling/ToolResolutionStrategyTests.cs
+++ b/src/Cake.Core.Tests/Unit/Tooling/ToolResolutionStrategyTests.cs
@@ -2,8 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using Cake.Core.IO;
 using Cake.Core.Tests.Fixtures;
+using Cake.Core.Tooling;
 using Cake.Testing;
+using NSubstitute;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


kcamp:
Any possibility of getting this merged for 0.23.0?  It seems like a safe improvement and the inability to gracefully side-step bad `PATH` values proves to be an occasional nuisance for devs in my org.  Thanks!

src/Cake.Core.Tests/Unit/Tooling/ToolResolutionStrategyTests.cs
```diff
@@ -177,6 +181,42 @@ public void Should_Return_Null_If_Tool_Could_Not_Be_Resolved()
                 // Then
                 Assert.Null(result);
             }
+
+            [Fact]
+            public void Should_Resolve_Tool_From_Environment_Variable_Gracefully_Proceed_If_FileSystem_Throw_Exception()
+            {
+                // Given
+                // The fixture is limiting  in how we can mock the dependencies, so start with the fixture
```
patriksvensson: No comments except `Given`, `When`, `Then` should be necessary here.

kcamp:
@patriksvensson No problem, got those cleaned up 👍

patriksvensson:
Merged! Thanks for your contribution! 👍

Issue #1614, opened 2017-05-24
Correct the class that TeamCityEnvironmentInfo inherits from


phillipsj:
Line 13 of the TeamCityEnvironmentInfo.cs inherits from AppVeyorInfo, not TeamCityInfo.

phillipsj:
@gep13 @devlead I can take this one, is it possible to get this turned around pretty fast.

devlead:
Fixed by #1615

Issue #1431, opened 2017-01-05
Add BuildStatisticValue method to TeamCityProvider 


marekgd:
or make WriteServiceMessage method public in TeamCityProvider 
so we can send any message to TeamCity

aateeque:
`ITeamCityProvider` already provides several constructs to speak to TeamCity. Can you share example & details of what `BuildStatisticValue` will do?

marekgd:
This is needed to integrate Code Coverage report - OpenCover
http://www.fabienruffin.com/integrating-opencover-test-coverage-statistics-with-teamcity/

kcamp:
@devlead Any thoughts on this?  I also have a need to extend the support of the ITeamCityProvider to support the service messages surrounding tests (testStarted, testFinished, testFailed, etc.)

I've started down the road of implementing extension methods, but the real sticking point is that the `WriteServiceMessage(string message, Dictionary<string, string> values)` method isn't exposed, nor is the `ICakeLog` on a given `ITeamCityProvider` instance.  Here is a very rough idea of the direction in which I am headed (and the limitations of it), for reference -- 707bac3af

As was mentioned, if we can expose the generic `WriteServiceMessage` method somehow, that would go a long way in letting any consumer write any messages, but it would be a breaking change for the `ITeamCityProvider` interface, so I don't know if that's an option at this point.

To avoid the breaking change, we could also create a new interface that represents the superset of all TC related functionality, have `TeamCityProvider` implement _that_ interface, and then add a `TeamCity2` property (or method?) to the `BuildSystem` class, which should prove workable, given that the `BuildSystem` class is sealed and implements/inherits no abstractions.  I put together a quick POC of this (minus tests, but it should be representative) that can be seen at b258a720a.  This solution would largely negate the refactoring done in the previously referenced commit, but I've left it intact for now.

Thoughts?  Not being able to utilize the full spread of TeamCity service messages is definitely limiting, it would be helpful to have this functionality.  I am happy to continue with this work & create a PR if there's a direction we can agree on.  Thanks!

PR #1615, opened 2017-05-24
GH1614:  Corrected inheritance issue.

phillipsj:
This just changes the inheritance to the correct abstract class.


kcamp:
Since this is technically a breaking change, could we potentially coordinate this change with another change to TeamCity objects in the same release/milestone?

Ideally, there are some changes outlined in GH-1431 that would go a long way in opening up the service message API on the `TeamCityProvider` in an effort to complete that API; more details are outlined in that issue.  Apologies for the random comment, but it seems opportune.  Thanks!

phillipsj:
@kcamp I agree if we have more teamcity updates that can go in.  I think the TeamCity provider needs some love compared to the others.

gep13:
Relates to #1614 1614

devlead:
@phillipsj  your changes have been merged, thanks for your contribution 👍

Issue #64, opened 2014-07-05
Documentation


patriksvensson:
- [x] Add XML documentation for public classes and members (Cake.Core and Cake.Common).
- [x] Generate API documentation (method and property aliases) for site.

PR #1649, opened 2017-06-16
GH1989: Removed unnecessary parameter

SteveGTR:
I removed the forward slash parameter because it's unnecessary because line #64 that replaces all forward slashes with backslashes. Hardly a critical bug, but as I was studying the code, saw it, and thought I'd comment. The code was introduced in commit ce0c04e3 on 6/9/2014.

### DIFF
--- a/src/Cake.Core/IO/Path.cs
+++ b/src/Cake.Core/IO/Path.cs
@@ -71,7 +71,7 @@ namespace Cake.Core.IO
             }
 
             // Remove trailing slashes.
-            FullPath = FullPath.TrimEnd('/', '\\');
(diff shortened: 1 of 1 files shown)


gep13:
@SteveGTR were you in a position to rebase this PR as requested by @devlead?  If not, we can look into doing that work for you.

Also, was an issue raised for this PR?  As per our contributing documentation, we prefer that this is the case.  Thanks for your help!

Roemer:
PR #1756 replaces this one as this one seems dead.

devlead:
Relates to #1989

PR #1664, opened 2017-06-27
(build) global.json fix

JonCubed:
bump dotnet sdk to 1.0.4

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "projects": [ "src" ],
     "sdk": {
-        "version": "1.0.0-preview2-003121"
+        "version": "1.0.4"
     }
 }


SIkebe:
@mholo65 (Probably) not to reference a wrong version of SDK accidentally. By default, .NET Core projects use the latest one installed on your machine. Same as pinning the Cake version in packages.config.

@JonCubed Is that right?

JonCubed:
@SIkebe correct

@mholo65 current [cake script](https://github.com/cake-build/cake/blob/develop/build.ps1#L36) pin the sdk to 1.0.4 so thought best to keep global.json the same

mholo65:
@JonCubed Thanks for your contribution!

PR #1666, opened 2017-06-28
GH1665: Implemented NUnit3 params and multiple results

### LINKED ISSUE #1665: NUnit3Settings: Params and multiple results
There are some important things missing from the ```NUnit3Settings``` class:

### Params
It is possible to pass parameters to the tests (which can be read with ```TestContext.Parameters[]```.
An example would be:
```--params=uia=2```
So it's one or more key-value pairs.
It would be great to have that in the ```NUnit3Settings``` as well.

### Multiple results
In some cases, I need to be able to have multiple result formats.
For example when using Jenkins: Jenkins only understands the v2 format but the v3 format has more information. So I need to have the v2 for Jenkins and v3 in the artefacts.
NUnit-Console allows to receive multiple ```--result``` so the current property should be extended to be a list of a new class like ```NUnit3Result``` which has a ```Format```, ```Transform``` and ```Name``` property which would replace the current result properties from ```NUnit3Settings```

### PR DESCRIPTION
**Pull request by Roemer**
This is the PR to fix #1665
This is a breaking change.
The ```Results```, ```ResultFormat``` and ```ResultTransform``` are now combined in the ```NUnit3Result``` class and creation looks like:
```csharp
settings.Results = new[]
{
    new NUnit3Result { FileName = "NewTestResult.xml", Format = "nunit2", Transform = "nunit2.xslt" },
    new NUnit3Result { FileName = "NewTestResult2.xml", Format = "nunit3" }
};
```
The parameters can be added like:
```csharp
settings.Params = new Dictionary<string, string>
{
    ["one"] = "1",
    ["two"] = "2",
    ["three"] = "3"
};
```

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using Cake.Common.Tests.Fixtures.Tools;
 using Cake.Common.Tools.NUnit;
 using Cake.Core;
@@ -175,7 +176,7 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
             {
                 // Given
                 var fixture = new NUnit3RunnerFixture();
-                fixture.Settings.Results = "NewResults.xml";
+                fixture.Settings.Results = new[] { new NUnit3Result { FileName = "NewResults.xml" } };
                 fixture.Settings.NoResults = true;
 
                 // When
@@ -191,7 +192,7 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
             {
                 // Given
                 var fixture = new NUnit3RunnerFixture();
-                fixture.Settings.Results = "NewTestResult.xml";
+                fixture.Settings.Results = new[] { new NUnit3Result { FileName = "NewTestResult.xml" } };
 
                 // When
                 var result = fixture.Run();
@@ -218,9 +219,11 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
                 fixture.Settings.OutputFile = "stdout.txt";
                 fixture.Settings.ErrorOutputFile = "stderr.txt";
                 fixture.Settings.Full = true;
-                fixture.Settings.Results = "NewTestResult.xml";
-                fixture.Settings.ResultFormat = "nunit2";
-                fixture.Settings.ResultTransform = "nunit2.xslt";
+                fixture.Settings.Results = new[]
+                {
+                    new NUnit3Result { FileName = "NewTestResult.xml", Format = "nunit2", Transform = "nunit2.xslt" },
+                    new NUnit3Result { FileName = "NewTestResult2.xml", Format = "nunit3" }
+                };
                 fixture.Settings.Labels = NUnit3Labels.All;
                 fixture.Settings.TeamCity = true;
                 fixture.Settings.NoHeader = true;
@@ -234,6 +237,12 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
                 fixture.Settings.DisposeRunners = true;
                 fixture.Settings.ShadowCopy = true;
                 fixture.Settings.Agents = 3;
+                fixture.Settings.Params = new Dictionary<string, string>
+                {
+                    ["one"] = "1",
+                    ["two"] = "2",
+                    ["three"] = "3"
+                };
 
                 // When
                 var result = fixture.Run();
@@ -244,10 +253,14 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
                         "--stoponerror \"--work=/Working/out\" \"--out=/Working/stdout.txt\" " +
                         "\"--err=/Working/stderr.txt\" --full " +
                         "\"--result=/Working/NewTestResult.xml;format=nunit2;transform=/Working/nunit2.xslt\" " +
+                        "\"--result=/Working/NewTestResult2.xml;format=nunit3\" " +
                         "--labels=All --teamcity --noheader --nocolor --verbose " +
                         "\"--config=Debug\" \"--framework=net3.5\" --x86 " +
                         "--dispose-runners --shadowcopy --agents=3 " +
-                        "--process=InProcess --domain=Single", result.Args);
+                        "--process=InProcess --domain=Single " +
+                        "\"--params=one=1\" " +
+                        "\"--params=two=2\" " +
+                        "\"--params=three=3\"", result.Args);
             }
 
             [Fact]
--- a/src/Cake.Common.Tests/Unit/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporterTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/SpecFlow/TestExecutionReport/SpecFlowTestExecutionReporterTests.cs
@@ -304,8 +304,7 @@ namespace Cake.Common.Tests.Unit.Tools.SpecFlow.TestExecutionReport
                 var nUnit3Settings = new NUnit3Settings
                 {
                     ShadowCopy = false,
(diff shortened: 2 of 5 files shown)


Roemer:
CodeFactor shows duplication error on a part I haven't modified (in ```NUnitRunner``` and ```NUnit3Runner```).

mholo65:
@Roemer thanks for your PR. I've assigned the related issue to the v0.22.0 milestone which will include other breaking changes.

@devlead do you know why CodeFactor complains on duplicates now? Code in question have not been touched.

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
```diff
@@ -234,6 +237,12 @@ public void Should_Set_Commandline_Switches()
                 fixture.Settings.DisposeRunners = true;
                 fixture.Settings.ShadowCopy = true;
                 fixture.Settings.Agents = 3;
+                fixture.Settings.Params = new Dictionary<string, string>
+                {
+                    ["one"] = "1",
+                    ["two"] = "2",
+                    ["three"] = "3"
```
mholo65: I've completely missed that one can initialize Dictionary like this in C# 6

Roemer: Kinda nice isn't it :)

mholo65: Yep, it's much cleaner!

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
```diff
@@ -218,9 +219,11 @@ public void Should_Set_Commandline_Switches()
                 fixture.Settings.OutputFile = "stdout.txt";
                 fixture.Settings.ErrorOutputFile = "stderr.txt";
                 fixture.Settings.Full = true;
-                fixture.Settings.Results = "NewTestResult.xml";
-                fixture.Settings.ResultFormat = "nunit2";
-                fixture.Settings.ResultTransform = "nunit2.xslt";
+                fixture.Settings.Results = new[]
+                {
+                    new NUnit3Result { FileName = "NewTestResult.xml", Format = "nunit2", Transform = "nunit2.xslt" },
```
ChrisMaddock: I don't believe combining format and transform like this is currently supported in NUnit - so this test currently tests invalid usage. As additional functionality, it would be cool if Cake would validate against this! 🙂

ChrisMaddock: Of course, should someone later make a PR to NUnit to support this, this behaviour may change - so forbidding it in Cake _may_ not be sensible. I'm not sure what Cake's general opinion on such things is.

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
```diff
@@ -244,10 +253,14 @@ public void Should_Set_Commandline_Switches()
                         "--stoponerror \"--work=/Working/out\" \"--out=/Working/stdout.txt\" " +
                         "\"--err=/Working/stderr.txt\" --full " +
                         "\"--result=/Working/NewTestResult.xml;format=nunit2;transform=/Working/nunit2.xslt\" " +
```
ChrisMaddock: This line is also invalid NUnit, although I see it isn't from this PR. Maybe there's more to be changed here!

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
```diff
@@ -244,10 +253,14 @@ public void Should_Set_Commandline_Switches()
...
+                        "\"--result=/Working/NewTestResult2.xml;format=nunit3\" " +
                         "--labels=All --teamcity --noheader --nocolor --verbose " +
                         "\"--config=Debug\" \"--framework=net3.5\" --x86 " +
                         "--dispose-runners --shadowcopy --agents=3 " +
-                        "--process=InProcess --domain=Single", result.Args);
+                        "--process=InProcess --domain=Single " +
+                        "\"--params=one=1\" " +
+                        "\"--params=two=2\" " +
+                        "\"--params=three=3\"", result.Args);
```
ChrisMaddock: This NUnit syntax currently currently falls over if the parameter/name includes a semi-colon. It's a flaw with NUnit. 😞 https://github.com/nunit/nunit-console/issues/23 Maybe Cake could guard against that - I imagine when we fix it, it may involve a new command line option.

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
mholo65: @ChrisMaddock ok, so it's either `Format` or `Transform`. Makes sense.

Anyhow, we have no sanity check around that currently either https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/NUnit/NUnit3Runner.cs#L128

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
mholo65: @ChrisMaddock What will happen when combining `format` and `transform`? Does it give an error or does one have precedence over the other?

mholo65:
Thanks @ChrisMaddock for reviewing this! 👍

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
Roemer: I will try this tomorrow and see how it behaves. `Transform` probably only works if the `Format` is not set or is set to `nunit3` (default).

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
ChrisMaddock: I haven't tried it in Cake - but NUnit itself currently crashes! https://github.com/nunit/nunit-console/issues/252 We'll be fixing that one - but it will be by returning an error message from the NUnit Console, rather than creating a feature!

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
ChrisMaddock: If `transform` is set - format should be unset. (Internally - NUnit will set format to 'user' - however, there's no need to set this on the command line.)

Roemer: So should we make this check or let the user set both (and nunit will correctly unset `format`)?

mholo65: @Roemer I'd say it's out of scope for this PR. We should raise another issue to track that (NUnit will crash if both are supplied https://github.com/nunit/nunit-console/issues/252).

Are you interested in raising issue and submitting a PR for that? Note, the PR should be submitted after this is merged to avoid merge conflicts.

Roemer: @mholo65 Sure, I can do that.

Roemer: See #1671

src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
Roemer: So should I take care of this in this PR?

devlead:

Roemer:
Awesome, thanks :+1:

PR #1673, opened 2017-07-09
docs: the xunit.runners package was deprecated, use this instead

ghuntley:


gep13:
@ghuntley thanks for making this change. Will merge once CI's finish.

Issue #1679, opened 2017-07-12
Cake on dotnet core doesn't load reference dll correctly if referenced from a subdirectory


phenixdotnet:
### What You Are Seeing?
If you're using Cake.CoreCLR, the referenced libs (using #reference) aren't loaded correctly if the cake script is in a subdirectory. 
```
Error: System.ArgumentException: Absolute path information is required.
Parameter name: assemblyPath
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
   at Cake.Core.Reflection.AssemblyLoader.Load(FilePath path, Boolean verify)
   at Cake.Core.Scripting.ScriptRunner.Run(IScriptHost host, FilePath scriptPath, IDictionary`2 arguments)
   at Cake.Commands.BuildCommand.Execute(CakeOptions options)
   at Cake.CakeApplication.Run(CakeOptions options)
   at Cake.Program.Main()
```

### What is Expected?
The lib should be loaded and available in the cake script

### What version of Cake are you using?
Cake.CoreCLR version 0.21.0-dotnetV10tooling0001 from myget
Cake.Docker version 0.7.7 from nuget

### Are you running on a 32 or 64 bit system?
x64

### What environment are you running on?  Windows? Linux? Mac?
- Windows x64 
- Mac OS 16.6.0 Darwin

### How Did You Get This To Happen? (Steps to Reproduce)
You can find a repro project here: https://github.com/phenixdotnet/cake-repro

The important thing is to use a cake script file from a subdirectory (in the repro example, the build/docker.cake file)

### Output Log

logs from .\build.ps1 -v Diagnostic and .\build.ps1 -v Diagnostic -Experimental

~~~sh
Preparing to run build script...
Feeds used:
  C:\Users\vlaine\.nuget\packages\
  https://api.nuget.org/v3/index.json
  https://www.myget.org/F/cake/api/v3/index.json
  C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\

Restoring NuGet package Cake.CoreClr.0.21.0-dotnetV10tooling0001.
Restoring NuGet package Cake.Docker.0.7.7.
Adding package 'Cake.CoreClr.0.21.0-dotnetV10tooling0001' to folder 'C:\dev\Test\ReproCake\tools'
Adding package 'Cake.Docker.0.7.7' to folder 'C:\dev\Test\ReproCake\tools'
[...]

PR #1680, opened 2017-07-12
Fix bug #1679 : Cake on dotnet core doesn't load reference dll correctly if referenced from a subdirectory

phenixdotnet:
This PR fix the bug #1679 by using ```path.MakeAbsolute(env)```


devlead:
@phenixdotnet  your changes have been merged, thanks for your contribution 👍

PR #1683, opened 2017-07-13
GH1682: Add version to restore, to fix package reference issue

### LINKED ISSUE #1682: Pre-Release PackageReferences get wrong dependency version number
This is because of an bug in MSBuild / .NET CLI ( NuGet/Home#4337 )
image

devlead:
* Fixes #1682


mholo65:
@gep13 this should be added to `Cake.Recipe` also

Issue #1521, opened 2017-03-06
Tools resolution doesn't respect added 'prerelease' filter


matkoch:
### What You Are Seeing?

Tool "GitVersion" was not updated to v4.0.0-betaxxxx.

### What is Expected?

An update.

### What version of Cake are you using?

Latest I guess on Win10/x64.

### How Did You Get This To Happen? (Steps to Reproduce)

- Had `#tool "nuget:?package=GitVersion.CommandLine"`
- Changed to `#tool "nuget:?package=GitVersion.CommandLine&prerelease"`
- Nupkg inside `tools` folder was still the stable release

This hurts probably the most with build servers.

patriksvensson:
@matkoch You will need to clear the folder manually before the new version will be downloaded.

matkoch:
@patriksvensson Shouldn't the `#tool` directive always pick the highest possible package, if no version was specified? I think that makes sense.

patriksvensson:
@matkoch Not strange. You might notice that the tools folders are not versioned. We simply assume that the versions are correct. There is an issue for fixing all this, but nothing that is being actively worked on.

boebeng:
@patriksvensson Is there any progress on this? I recently started with Cake build and it seems quite interesting and we would like to replace psake with Cake build. This issue is somehow blocking me.
Is there a workaround? For example an automatism to delete the tools folder upfront?

gep13:
Closing as a duplicate of https://github.com/cake-build/cake/issues/753

daveaglick:
For anyone interested in this issue, there's an easy workaround. I just stick this in my PowerShell bootstraper:

```
# Delete the tools and addins folders
Remove-Item $TOOLS_DIR\* -Force -Recurse -ErrorAction Ignore
```

You can make it more specific too if you know it's only one tool or addin you want to ensure is always fresh.

matkoch:
@daveaglick thanks, `-ErrorAction Stop` would be better I think. Otherwise you could have strange side effects again.

Issue #753, opened 2016-03-12
Tool Versioning


dbent:
To make sure builds are repeatable for all users and that upstream changes don't impact the build process I always peg tools to particular versions, e.g:

```
#tool nuget:?package=Foo.Bar.Baz&version=1.0.0
```

This works well except in one particular case. Assume upstream makes significant backwards-incompatible changes and I update the build to support those changes. In that case I'll bump the version number as necessary:

```
#tool nuget:?package=Foo.Bar.Baz&version=2.0.0
```

This will then work fine for fresh clones of the project, but for anyone updating an existing clone, the installed `1.0.0` version of the tool will persist. Thus their build will break until they delete the `tools/Foo.Bar.Baz` directory to force an update to `2.0.0`

An "easy" solution would be to add the version to the directory so that instead of just `tools/Foo.Bar.Baz` you would have `tools/Foo.Bar.Baz.1.0.0` and `tools/Foo.Bar.Baz.2.0.0`. However, that complicates resolving which tool to use if multiple copies may exist. #736 may be relevant here.

Thoughts?

patriksvensson:
@dbent Yes, this need to be done in #736 as you mention for things to work reliably. Just changing the name of the folders would randomize what version that will be used as it is now. 

I've started working on it [here](https://github.com/patriksvensson/cake/tree/f/toolresolution) but it's not ready yet, but folder version support will be added as well once ready.

robertcoltheart:
The .nupkg files are preserved when they are downloaded to `Tools` aren't they? Why not use the version of the nupkg to see whether they need to be updated or not as compared to the `packages.config`?

saasen:
When running cake on our build server (TeamCity), it's as @dbent says. Since the folders or tools are not versioned, it will not download the latest tool when changing the version of a tool directive. This could be solved by cleaning the directory completely, but it would slow down builds a lot.

robertcoltheart:
@saasen This is solved by putting tool dependencies in the `packages.config` and using the latest bootstrapper from [here](http://cakebuild.net/docs/tutorials/setting-up-a-new-project). Changes to the `packages.config` and now MD5 checked and tools and re-downloaded if the versions change.

saasen:
Thanks a lot @robertcoltheart. I'll try that.

saasen:
@robertcoltheart: I've played around with this, and it seems like I can't get it to work. Let me explain.

Here's the old way of setting up tools and addins in `build.cake`:

```
[...]

PR #1706, opened 2017-07-30
GH753: Version NuGet package references on disk.

### LINKED ISSUE #1572: Changing version of nuget package used in #load does not download latest nuget package
### What You Are Seeing?

I have a build.cake script that contains the following pre-processor directive:

    #load "nuget:?package=MyPackage&version=0.1.0-feat-mg-16017-co0020"

I can successfully build my project through `build.ps1`. But then, I want to use a new version of this nuget package. I therefore change my load directive to use version 21:

    #load "nuget:?package=MyPackage&version=0.1.0-feat-mg-16017-co0021"

When I start build.ps1, it continues to use version 20.

### What is Expected?

build.ps1 / cake should download version 21 of my nuget pacakge and use it.

0.19.3

64

Windows

No

See above

patriksvensson:
If a version is specified when using either `#load`, `#tool` or `#addin`,
the package folder on disk will be suffixed with the version number.

Not specifying a version will keep the current behavior.

Closes #1572

### DIFF
--- a/src/Cake.NuGet/NuGetPackageInstaller.cs
+++ b/src/Cake.NuGet/NuGetPackageInstaller.cs
@@ -12,6 +12,15 @@ using Cake.Core.IO;
 using Cake.Core.IO.NuGet;
 using Cake.Core.Packaging;
 
+#if NETCORE
+using NuGet.Versioning;
+#else
+using NuGet;
+#endif
+
+using IFileSystem = Cake.Core.IO.IFileSystem;
+using PackageReference = Cake.Core.Packaging.PackageReference;
+
 namespace Cake.NuGet
 {
     /// <summary>
@@ -117,44 +126,33 @@ namespace Cake.NuGet
                 throw new ArgumentNullException(nameof(path));
             }
 
-            path = path.MakeAbsolute(_environment);
-
-            var root = _fileSystem.GetDirectory(path);
-            var packagePath = path.Combine(package.Package);
-
             // Create the addin directory if it doesn't exist.
+            path = GetPackagePath(path.MakeAbsolute(_environment), package);
+            var root = _fileSystem.GetDirectory(path);
             if (!root.Exists)
             {
                 _log.Debug("Creating directory {0}", path);
                 root.Create();
             }
 
-            // Fetch available content from disc.
-            var content = _contentResolver.GetFiles(packagePath, package, type);
-            if (content.Any())
+            // Package already exist?
+            var packagePath = GetPackagePath(root);
+            if (packagePath != null)
             {
-                _log.Debug("Package {0} has already been installed.", package.Package);
-                return content;
+                // Fetch available content from disc.
+                var content = _contentResolver.GetFiles(packagePath, package, type);
+                if (content.Any())
+                {
+                    _log.Debug("Package {0} has already been installed.", package.Package);
+                    return content;
+                }
             }
 
             // Install the package.
-            _log.Debug("Installing NuGet package {0}...", package.Package);
-            var nugetPath = GetNuGetPath();
-            var process = _processRunner.Start(nugetPath, new ProcessSettings
-            {
-                Arguments = GetArguments(package, path, _config),
-                RedirectStandardOutput = true,
-                Silent = _log.Verbosity < Verbosity.Diagnostic
-            });
-            process.WaitForExit();
+            InstallPackage(package, path);
 
-            var exitCode = process.GetExitCode();
-            if (exitCode != 0)
-            {
-                _log.Warning("NuGet exited with {0}", exitCode);
-                var output = string.Join(Environment.NewLine, process.GetStandardOutput());
-                _log.Verbose(Verbosity.Diagnostic, "Output:\r\n{0}", output);
-            }
+            // Try locating the install folder again.
+            packagePath = GetPackagePath(root);
 
             // Get the files.
             var result = _contentResolver.GetFiles(packagePath, package, type);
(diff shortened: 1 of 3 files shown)


src/Cake.NuGet/NuGetPackageInstaller.cs
```diff
@@ -175,6 +160,38 @@ public IReadOnlyCollection<IFile> Install(PackageReference package, PackageType
             return result;
         }
 
+        private static DirectoryPath GetPackagePath(DirectoryPath root, PackageReference package)
+        {
+            if (package.Parameters.ContainsKey("version"))
+            {
+                var version = package.Parameters["version"].First();
+                return root.Combine($"{package.Package}.{version}");
```
devlead: Seeing issues because nuget supports partial nuget version numbers, getting this error building Cake with this PR
```
Installing NuGet package ReportGenerator...
Executing: /opt/nuget/nuget.exe install "ReportGenerator" -OutputDirectory "/pullrequest/tools" -Source "https://www.nuget.org/api/v2" -Version "2.4.5" -NonInteractive
Could not find any relevant files for tool 'ReportGenerator'. Perhaps you need an include parameter?
Error: Cake.Core.CakeException: Failed to install tool 'ReportGenerator'.
```
`ls tools` gives:
```
GitVersion.CommandLine.3.6.2
OpenCover.4.6.519
ReportGenerator.2.4.5.0
coveralls.io.1.3.4
gitreleasemanager.0.5.0 
```
Changing `#tool "nuget:https://www.nuget.org/api/v2?package=ReportGenerator&version=2.4.5"` to `#tool "nuget:https://www.nuget.org/api/v2?package=ReportGenerator&version=2.4.5.0"` in `build.cake` sorts the issue.

How smart do we need to be here? 

Checking nuspec of `ReportGenerator`
```xml
<version>2.4.5.0</version>
```
vs i.e. GitReleaseManager
```
<version>0.6.0</version>
```
but nuget omits the .0 in the UI so no good way of knowing that.

Perhaps `  return root.Combine($"{package.Package}.{version}");` should fetched with a globber and firstordefault best match.

patriksvensson: Well, there has to be some rule that governs this and it would be better to figure that one out than to guess at what the folder name should be. I can take some time tonight and read some NuGet source code.

devlead: Well rule is that nuget displays x.x.x but if version element has x.x.x.x it'll be used for folder name :)

patriksvensson: Well, not that easy. The `NuGetVersion` representation of `0.6.0` => `0.6.0.0`, but that's not the install folder. Same with `2.4.5` => `2.4.5.0`. Might be different with pre-release packages as well. I'll take a look.

devlead: https://www.nuget.org/packages/ReportGenerator/2.4.5

image


nuget install ReportGenerator -Version 2.4.5

image

It doesn't always happen, only when a package has done the x.x.x.x in nuspec
```xml
<version>2.4.5.0</version
```

patriksvensson: Yeah, but internally, NuGet CLI uses `NuGetVersion` for this that will have the same representation regardless of trailing zeros or not.

devlead: Well seems to be package dependent as GitReleaseManager 
```xml
<version>0.6.0</version>
``` 

`nuget install gitreleasemanager -Version 0.6.0`
image

patriksvensson: Well it might be. It might also be that it doesn't include revision if major version is less than zero :)

devlead: > Yeah, but internally, NuGet CLI uses NuGetVersion for this that will have the same representation regardless of trailing zeros or not.

Ok, well folder will have "same" input parameters but different outcomes, currently we won't find the correct folder.
Also `nuget install GitReleaseManager -Version 0.6.0` will generate a lowecase folder because that's the package id, so that could fail on unix.

So almost list addin folder folders and select startswith best case first then fallback to ignore case  😉

devlead: > Yeah, but internally, NuGet CLI uses NuGetVersion for this that will have the same representation regardless of trailing zeros or not.

`nuget install ReportGenerator -Version 2.5.8` yields
image

So don't think so. it's nuspec says
```xml
<version>2.5.8</version>
```

patriksvensson: Nah, we need a better solution than guessing. Will think about it some more.

devlead: Agreed, nuget.exe doesn't give much info though. hacky way would be ExcludeVersion and have ` root.Combine($"{package.Package}.{version}")` be outputdir 😆 

`nuget install ReportGenerator -Version 2.4.5 -Verbosity detailed` says it installs to 2.4.5 but that's a lie 😉

patriksvensson: @devlead Ok, I did like that. Good enough I think.

mholo65:
Relates to #1521

gep13:
This also relates to https://github.com/cake-build/cake/issues/753

devlead:
@patriksvensson  your changes have been merged, thanks for your contribution 👍

Issue #1710, opened 2017-07-31
Add alias for simple sub-directory listing


patridge:
Rather than having to learn anything about the pattern syntax used in GlobbingAliases various methods (including its own [`GetDirectories`](https://github.com/cake-build/cake/blob/2595de795c56066eee8391641f3298082b4384cd/src/Cake.Common/IO/GlobbingAliases.cs#L97) method that can be used to similar effect with the right pattern), it seems like it would be nice to have a simplified method for getting a list of all sub-directories in any given DirectoryPath.

Goal syntax:

> `var directories = GetSubDirectories(someDirectoryPath);`

[Chat discussion start](https://gitter.im/cake-build/cake?at=597f4fe1210ac269203fe970) ([approved for issue/PR](https://gitter.im/cake-build/cake?at=597f7eb3c101bc4e3ad813b1))

devlead:
Fixed by #1711

PR #1645, opened 2017-06-14
GH1540: Update to latest Roslyn

patriksvensson:
This PR will do the following:

- [x] Update all projects to target `net462` instead of `net45`
- [x] Update Roslyn to the latest version
- [x] Remove old Roslyn (regular and experimental)
- [x] Remove the Mono scripting engine
- [x] Unify script engine between full framework and .net core
- [x] Make it build on CI again
- [x] Add loading assembly from stream for debugging purposes (see AssemblyLoader).
- [x] Verify that debugging works on full framework and .NET Core
- [x] Clean up the code a bit

PR #1711, opened 2017-07-31
GH1711: GetSubDirectories alias

patridge:
As described in #1710, this offers a simple Alias to get a list of sub-directories for a given DirectoryPath.

### DIFF
--- a/src/Cake.Common/IO/DirectoryAliases.cs
+++ b/src/Cake.Common/IO/DirectoryAliases.cs
@@ -537,5 +537,45 @@ namespace Cake.Common.IO
         {
             DirectoryMover.MoveDirectory(context, directoryPath, targetDirectoryPath);
         }
+
+        /// <summary>
+        /// Gets a list of all the directories inside a directory.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// var directories = GetSubDirectories("some/dir");
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="directoryPath">The directory path.</param>
+        /// <returns>An absolute directory path.</returns>
+        [CakeMethodAlias]
+        [CakeAliasCategory("List")]
+        public static DirectoryPathCollection GetSubDirectories(this ICakeContext context, DirectoryPath directoryPath)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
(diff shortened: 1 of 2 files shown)


src/Cake.Common/IO/DirectoryAliases.cs
```diff
@@ -458,5 +458,44 @@ public static void MoveDirectory(this ICakeContext context, DirectoryPath direct
...
+        /// <code>
+        /// var directories = GetSubDirectories("some/dir");
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <param name="directoryPath">The directory path.</param>
+        /// <returns>An absolute directory path.</returns>
+        [CakeMethodAlias]
+        [CakeAliasCategory("List")]
+        public static List<DirectoryPath> GetSubDirectories(this ICakeContext context, DirectoryPath directoryPath)
```
mholo65: For consistency. This should return `DirectoryPathCollection`

mholo65: @devlead?

devlead: > For consistency. This should return DirectoryPathCollection

Agreed @mholo65 :+1:

patridge: Good catch! Fixed

patridge:
Let me know if/what I need to do about the TeamCity build failing. I wasn't totally sure what the error was describing when I logged in there as a guest.

mholo65:
@patridge the fail on TeamCity has to do with reference assemblies missing/not found. Bug was introduced in PR #1645 and has nothing to do with your changes.

@devlead @patriksvensson can we get TC Agend with `.NET 4.6.2` Targeting pack installed? Also, agent seems to use VS 2015:
```
[Step 1/1] MSBuild auto-detection: using msbuild version '14.0' from 'C:\Program Files (x86)\MSBuild\14.0\bin\amd64'.
```

mholo65:
@patridge could you squash all commits into one. Thanks!

devlead:
@mholo65 agents are maintained by JetBrains now, we could probably add a build step to install targeting pack, but best would be if they updated their agents. We'll reach out to them but meanwhile check isn't required.

devlead:
@patridge  your changes have been merged, thanks for your contribution 👍

PR #1722, opened 2017-08-03
ILRepackSettings.Libs should be DirectoryPaths

### LINKED ISSUE #1720: ILRepackSettings.Libs should be List of DirectoryPath
### What You Are Seeing?
When using the following:

```
ILRepack ("./output.dll", "first.dll", new [] { "second.dll" }, new ILRepackSettings { Libs = new List<FilePath> { "./ref.dll" } });
```

I get an error that the assembly `ref.dll` was unable to be resolved.  Which you think is weird since I'm explicitly adding it as a reference in the call.  Then when you try to run the command manually you get the same error, but upon discovering that the ILRepack arg `/lib:` is actually supposed to take a directory, not a specific file, and you try that manually and it works, you realize the original ILRepack implementation you wrote is wrong!

### What is Expected?

It should work without error.

PR coming.

Redth:
When this was originally implemented, it wasn’t very obvious, but ILRepack’s `/lib:<path>` argument should actually be a directory, not a file.

This sheds a _bit_ more light:
https://github.com/gluck/il-repack/blob/master/ILRepack/RepackOptions.cs#L221

This commit fixes up the `ILRepackSettings.Libs` to be the correct type.

This should fix #1720

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/ILRepack/ILRepackRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/ILRepack/ILRepackRunnerTests.cs
@@ -368,13 +368,13 @@ namespace Cake.Common.Tests.Unit.Tools.ILRepack
             {
                 // Given
                 var fixture = new ILRepackRunnerFixture();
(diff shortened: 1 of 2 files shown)


devlead:
@Redth  your changes have been merged, thanks for your contribution 👍

Issue #1719, opened 2017-08-02
Jenkins BRANCH_NAME is missing


Roemer:
Jenkins provides a BRANCH_NAME environment variable which is not mapped into cake.

There might be other variables missing as well.
Here's the list of variables I get from my Jenkins instance:

Variable | Description
------------ | -------------
BRANCH_NAME | For a multibranch project, this will be set to the name of the branch being built, for example in case you wish to deploy to production from master but not from feature branches; if corresponding to some kind of change request, the name is generally arbitrary (refer to CHANGE_ID and CHANGE_TARGET).
CHANGE_ID | For a multibranch project corresponding to some kind of change request, this will be set to the change ID, such as a pull request number, if supported; else unset.
CHANGE_URL | For a multibranch project corresponding to some kind of change request, this will be set to the change URL, if supported; else unset.
CHANGE_TITLE | For a multibranch project corresponding to some kind of change request, this will be set to the title of the change, if supported; else unset.
CHANGE_AUTHOR | For a multibranch project corresponding to some kind of change request, this will be set to the username of the author of the proposed change, if supported; else unset.
CHANGE_AUTHOR_DISPLAY_NAME | For a multibranch project corresponding to some kind of change request, this will be set to the human name of the author, if supported; else unset.
CHANGE_AUTHOR_EMAIL | For a multibranch project corresponding to some kind of change request, this will be set to the email address of the author, if supported; else unset.
CHANGE_TARGET | For a multibranch project corresponding to some kind of change request, this will be set to the target or base branch to which the change could be merged, if supported; else unset.
BUILD_NUMBER | The current build number, such as "153"
BUILD_ID | The current build ID, identical to BUILD_NUMBER for builds created in 1.597+, but a YYYY-MM-DD_hh-mm-ss timestamp for older builds
BUILD_DISPLAY_NAME | The display name of the current build, which is something like "#153" by default.
JOB_NAME | Name of the project of this build, such as "foo" or "foo/bar".
JOB_BASE_NAME | Short Name of the project of this build stripping off folder paths, such as "foo" for "bar/foo".
BUILD_TAG | String of "jenkins-${JOB_NAME}-${BUILD_NUMBER}". All forward slashes (/) in the JOB_NAME are replaced with dashes (-). Convenient to put into a resource file, a jar file, etc for easier identification.
[...]

PR #1724, opened 2017-08-04
GH1719: Cleaned old and added new Jenkins variables

Roemer:
See  #1719.


Roemer:
I'm not totally sure if it makes sense to keep the different JenkinsXXXInfo.cs files and classes. Wouldn't it be easier to have all information in one class? What do you think?

patriksvensson:
@Roemer It might be easier, but it would also be a breaking change so I think we should leave it as is for now.

patriksvensson:
@cake-build/cake-team Any opinions about this?

devlead:
@patriksvensson there's several breaking changes already in this PR, but I'm unsure it's better to flatten out though.

devlead:

PR #1737, opened 2017-08-07
GH1732: Unify Nuget content resolvers.

### LINKED ISSUE #1732: Remove NuGet.Core dependency in Cake.NuGet
We should do just fine with NuGet.Client. Makes life easier as we don't need separate NuGet dependency for Cake and Cake.CoreCLR

mholo65:
Use NuGet.Client on both full framework and .NET Core

Fixes #1732


devlead:

PR #1741, opened 2017-08-08
GH1739: NuGetContentResolver finds assemblies in root

### LINKED ISSUE #1739: NuGetContentResolver can't find assemblies if located in root
This is indeed an error with the package. But old V2 Content resolver could resolve packages without and TFM.

Issue here:
https://github.com/cake-build/cake/blob/develop/src/Cake.NuGet/NuGetContentResolver.cs#L106

Change that to a warning and just return assembly from root.

mholo65:
- Will treat root assemblies as AnyFramework
- Fixes #1739

### DIFF
--- a/src/Cake.NuGet/NuGetContentResolver.cs
+++ b/src/Cake.NuGet/NuGetContentResolver.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Cake.Core;
+using Cake.Core.Diagnostics;
 using Cake.Core.IO;
 using Cake.Core.Packaging;
 using NuGet.Frameworks;
@@ -17,15 +18,18 @@ namespace Cake.NuGet
         private readonly IFileSystem _fileSystem;
         private readonly ICakeEnvironment _environment;
         private readonly IGlobber _globber;
+        private readonly ICakeLog _log;
 
         public NuGetContentResolver(
             IFileSystem fileSystem,
             ICakeEnvironment environment,
-            IGlobber globber)
+            IGlobber globber,
+            ICakeLog log)
         {
             _fileSystem = fileSystem;
             _environment = environment;
             _globber = globber;
+            _log = log;
         }
 
         public IReadOnlyCollection<IFile> GetFiles(DirectoryPath path, PackageReference package, PackageType type)
@@ -87,6 +91,13 @@ namespace Cake.NuGet
                 return new List<IFile>();
             }
 
+            if (nearest == NuGetFramework.AnyFramework)
+            {
+                var framework = _environment.Runtime.TargetFramework;
+                _log.Warning("Could not find any assemblies compatible with {0} in NuGet package {1}. " +
+                             "Falling back to using root folder of NuGet package.", framework.FullName, package.Package);
+            }
+
             // Return the result.
             return mapping[nearest].Select(p => _fileSystem.GetFile(p)).ToList();
         }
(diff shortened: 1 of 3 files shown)


mholo65:
Will add some tests tomorrow

src/Cake.NuGet/NuGetContentResolver.cs
```diff
@@ -72,6 +76,11 @@ private IReadOnlyCollection<IFile> GetAddinAssemblies(DirectoryPath path, Packag
                 // Get relative path.
                 var relative = path.GetRelativePath(assembly.Path);
                 var framework = ParseFromDirectoryPath(current, relative.GetDirectory());
+                if (framework == null)
+                {
+                    _log.Warning("Unable to parse compatible framework from path {0}. Setting any framework.", assembly.Path.MakeAbsolute(_environment).FullPath);
```
mholo65: I'd like to hear @gep13's opinion on the Warning message.

gep13: @mholo65 only slight comment would be the last part `Setting any framework`.  I know what you are trying to say, but not sure if that works.  Perhaps something like `, as a result, no framework will be specified.`

patriksvensson: @mholo65 I'm not @gep13 (obviously), but I think that `Setting any framework.` sounds a little bit "leaky" in the sense that we leak internal information to the consumer which probably won't understand what we're talking about.

Perhaps we could change it to something similar to `Falling back to using root folder of NuGet package.` or something like that.

Thoughts @gep13?

mholo65: In this particular context, saying `falling back to using root folder of NuGet package.` is not the truth. When we can't parse framework from the path (be it `foo/bar/lib.dll` or just `lib.dll` we will treat it as compatible by setting framework to `AnyFramework`. By setting that, we can try loading it if we can't find a better match (e.g. `lib/net45/...`, `lib/net46/...`)

patriksvensson: @mholo65 What will the root folder for searching be with `AnyFramework`?

mholo65: It's actually the other way round. For each assembly, we will take the directory and try to parse the TFM and map that with the assemblies in that directory. After looping through all assemblies, we will find the closest matching our Target Framework (.NET Framework, Version=4.6 or .NET Standard, Version=1.6) and return all files mapped to that.

Now when looking over code once more, I found a smarter way to log warning. Now we log for each assembly found in root. Instead, we should only log if assemblies mapped to `AnyFramework` is actually returned. Let me modify this PR slightly :smile:

mholo65:
image

Example of Warning message. It's _almost_ the truth. We are now treating root folder as **AnyFramework**, but fact is that if path to assembly contains **any** (e.g. `lib/any/addin.dll`) it would also be parsed as **AnyFramework**. So *Falling back to using root folder...* might not be 100% accurate.

See **NuGetFramework.Parse(string)** Tests [here](https://github.com/NuGet/NuGet.Client/blob/4cccb13833ad29d6a0bcff055460d964f1b49cfe/test/NuGet.Core.Tests/NuGet.Frameworks.Test/SpecialFrameworkTests.cs#L12) 

Thought? @patriksvensson @gep13

patriksvensson:
@mholo65 I think it's close enough :) Haven't even seen the `any`, `agnostic` or `unsupported` convention before.

devlead:

PR #1747, opened 2017-08-12
GH1745: Change signature of methods for IScriptProcessor

### LINKED ISSUE #1745: Change parameter for InstallTools and InstallAddins in IScriptProcessor
No need to pass `ScriptAnalyzeResult` just pass a collection of `PackageReference` instead.

Will make life easier in Cake.Bakery.

mholo65:
- No need to pass the whole analyzer result.
- Fixes #1745

### DIFF
--- a/src/Cake.Core.Tests/Fixtures/ScriptProcessorFixture.cs
+++ b/src/Cake.Core.Tests/Fixtures/ScriptProcessorFixture.cs
@@ -19,10 +19,11 @@ namespace Cake.Core.Tests.Fixtures
         public FakeFileSystem FileSystem { get; set; }
         public ICakeEnvironment Environment { get; set; }
         public ICakeLog Log { get; set; }
-        public IToolLocator Tools { get; set; }
+        public IToolLocator ToolLocator { get; set; }
         public IPackageInstaller Installer { get; set; }
 
-        public ScriptAnalyzerResult Result { get; set; }
+        public List<PackageReference> Addins { get; set; }
+        public List<PackageReference> Tools { get; set; }
         public DirectoryPath InstallPath { get; set; }
 
         public ScriptProcessorFixture()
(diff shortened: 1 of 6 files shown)


src/Cake.Core/Scripting/ScriptProcessor.cs
```diff
@@ -85,10 +85,10 @@ public sealed class ScriptProcessor : IScriptProcessor
             installPath = installPath.MakeAbsolute(_environment);
 
             var result = new HashSet<FilePath>(PathComparer.Default);
-            if (analyzerResult.Addins.Count > 0)
+            if (addins.Any())
```
mholo65: `Possible multiple enumeration of IEnumerable`

Was thinking, either don't check if it contains any, pass `IReadOnlyCollection` or just let it be...

patriksvensson: My vote is for passing `IReadOnlyCollection<T>`.

src/Cake.Core.Tests/Fixtures/ScriptProcessorFixture.cs
```diff
@@ -19,28 +19,26 @@ public sealed class ScriptProcessorFixture
         public FakeFileSystem FileSystem { get; set; }
         public ICakeEnvironment Environment { get; set; }
         public ICakeLog Log { get; set; }
-        public IToolLocator Tools { get; set; }
+        public IToolLocator ToolLocator { get; set; }
         public IPackageInstaller Installer { get; set; }
 
-        public ScriptAnalyzerResult Result { get; set; }
+        public IEnumerable<PackageReference> Addins { get; set; }
```
patriksvensson: Make this an `List<PackagerReferences>` since it's in the fixture.

src/Cake.Core.Tests/Fixtures/ScriptProcessorFixture.cs
```diff
@@ -19,28 +19,26 @@ public sealed class ScriptProcessorFixture
...
+        public IEnumerable<PackageReference> Tools { get; set; }
```
patriksvensson: Make this an `List<PackagerReferences>` since it's in the fixture.

mholo65:
@patriksvensson fixed!

src/Cake.Core/Scripting/ScriptProcessor.cs
```diff
@@ -85,10 +84,10 @@ public sealed class ScriptProcessor : IScriptProcessor
...
```
patriksvensson: Should be `addins.Count > 0` here.

PR #1748, opened 2017-08-13
GH1746: ScriptAnalyzer will not throw on error

### LINKED ISSUE #1746: ScriptAnalyzer.Analyze() should not throw, instead return list of errors
Will make life easier with Cake.Bakery

mholo65:
- ScriptAnalyzer will not throw on error. Instead the result will
indicate if analysis succeeded or not. A list of errors is also present
in the analyzer result.
- Fixes #1746

### DIFF
--- a/src/Cake.Core/Scripting/Analysis/ScriptAnalyzerError.cs
+++ b/src/Cake.Core/Scripting/Analysis/ScriptAnalyzerError.cs
@@ -1,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Cake.Core.IO;
+
+namespace Cake.Core.Scripting.Analysis
+{
+    /// <summary>
+    /// Represents a script analysis error.
+    /// </summary>
+    public sealed class ScriptAnalyzerError
+    {
+        /// <summary>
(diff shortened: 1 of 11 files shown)


src/Cake.Core/Scripting/ScriptRunner.cs
```diff
@@ -131,6 +131,17 @@ public void Run(IScriptHost host, FilePath scriptPath, IDictionary<string, strin
             _log.Verbose("Analyzing build script...");
             var result = _analyzer.Analyze(scriptPath.GetFilename());
 
+            // Log all errors and throw
+            if (!result.Succeeded)
```
patriksvensson: Would be good to have a test for this execution path.

src/Cake.Core/Scripting/Analysis/ScriptAnalyzerError.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="file">The file containing the error.</param>
+        /// <param name="line">The line number for the error.</param>
+        /// <param name="message">The error message.</param>
+        public ScriptAnalyzerError(FilePath file, int line, string message)
+        {
+            if (file == null)
+            {
+                throw new ArgumentNullException(nameof(file));
+            }
+
```
patriksvensson: We should check message for null here.

mholo65:
@patriksvensson @devlead Fixed!

PR #1753, opened 2017-08-16
GH1752: Add AutomaticDecompression to HttpClientHandler

### LINKED ISSUE #1752: Extend DownloadFile to allow AcceptEncoding gzip
DownloadFile should be extended (probably the ```DownloadFileSettings``` class) to allow setting the ```AcceptEncoding: gzip``` header. This allows to optimize the traffic in some cases (like when getting xmls or other texts).

Should not be a big deal.
It's probably enough to add ```AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate``` to the ```HttpClientHandler``` in the ```GetHttpClient``` method.
We could even think to add this anyway as it should be able to not use compression if the webserver does not support it. What do you think?

Roemer:
Fixes #1752.

### DIFF
--- a/src/Cake.Common/Net/HttpAliases.cs
+++ b/src/Cake.Common/Net/HttpAliases.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


gep13:

Issue #1754, opened 2017-08-16
VSWhere not returning "prerelease" versions


olevett:
### What You Are Seeing?
VSWhere not picking up pre-release VS2017 installs

### What is Expected?

VSWhere has flag for prerelease versions

### Are you running on a CI Server?  If so, which one?

AppVeyor - https://ci.appveyor.com/project/dotnetfoundation/reactiveui/build/1.0.21

### How Did You Get This To Happen? (Steps to Reproduce)

On machine with only latest VS2017 version (e.g. AppVeyor Visual Studio 2017 Preview builds)

I think the solution is to add an extra option that maps to "-prerelease" (see https://blogs.msdn.microsoft.com/heaths/2017/06/24/vswhere-version-2-0-released/) which is ignored on older versions of VSWhere but wanted to check before sending a PR in

devlead:
This sounds like a great addition to Cake, PR would be welcome.

devlead:
Fixed by #1755

PR #1755, opened 2017-08-16
GH1754 - add prerelease flag to VSWhere

olevett:
As discussed in #1754, this adds a "prerelease" flag.

I think it's more appropriate as an additional flag on the existing VSWhereXXX functionality, but happy to refactor to a VSWherePrerelease or something if that's preferred.


devlead:
@olevett  your changes have been merged, thanks for your contribution 👍

PR #1756, opened 2017-08-16
GHxxx: Removed unnecessary parameter

Roemer:
Replaces PR #1649 as that one seems dead.


patriksvensson:
@Roemer Great! Could you also add a test to verify that it indeed is dead? :)

Roemer:
By dead I mean the original PR is dead.
The change itself is already tested in Should_Remove_Trailing_Slashes.

devlead:
@Roemer thanks for your contribution and support to the Cake project! as the PR is same as #1649,  we've chosen to close this in favor of that, and attribute original issuer.

PR #1758, opened 2017-08-17
GH1757: Remove installation folder if nuget install fails.

### LINKED ISSUE #1757: Cake.NuGet should remove intermediate directory if installation fails
Avoid situations like this when using with OmniSharp

image

patriksvensson:
Fixes #1757

### DIFF
--- a/src/Cake.NuGet/NuGetPackageInstaller.cs
+++ b/src/Cake.NuGet/NuGetPackageInstaller.cs
@@ -46,7 +46,7 @@ namespace Cake.NuGet
         /// <param name="toolResolver">The NuGet tool resolver.</param>
         /// <param name="contentResolver">The content resolver.</param>
         /// <param name="log">The log.</param>
-        /// <param name="config">the configuration</param>
+        /// <param name="config">The configuration.</param>
         public NuGetPackageInstaller(
             IFileSystem fileSystem,
             ICakeEnvironment environment,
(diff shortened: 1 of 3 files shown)


src/Cake.NuGet/NuGetPackageInstaller.cs
```diff
@@ -56,38 +56,13 @@ public sealed class NuGetPackageInstaller : INuGetPackageInstaller
...
-            _contentResolver = contentResolver;
-            _log = log;
-            _config = config;
+            _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _processRunner = processRunner ?? throw new ArgumentNullException(nameof(processRunner));
+            _toolResolver = toolResolver ?? throw new ArgumentNullException(nameof(toolResolver));
+            _contentResolver = contentResolver ?? throw new ArgumentNullException(nameof(contentResolver));
+            _log = log ?? throw new ArgumentNullException(nameof(log));
+            _config = config ?? throw new ArgumentNullException(nameof(config));
```
Roemer: This wasn't tested for null before. Is it correct that it now throws when it's null?

src/Cake.NuGet/NuGetPackageInstaller.cs
```diff
@@ -149,7 +126,16 @@ public IReadOnlyCollection<IFile> Install(PackageReference package, PackageType
...
             // Install the package.
-            InstallPackage(package, path);
+            if (!InstallPackage(package, path))
+            {
+                _log.Warning("An error occured while installing package {0}.", package.Package);
+                if (createdDirectory)
+                {
+                    _log.Debug("Deleting package directory {0}...", path);
+                    root.Delete(true);
+                    return new List<IFile>(0);
```
Roemer: I think in .Net 4.6 you could use ```Array.Empty<IFile>();``` which could reduce the allocations.

src/Cake.NuGet/NuGetPackageInstaller.cs
patriksvensson: Yes. It was an oversight that it was never checked for null.

src/Cake.NuGet/NuGetPackageInstaller.cs
patriksvensson: We're not extremely concerned about allocations. Especially since this is not a hot path.

src/Cake.NuGet/NuGetPackageInstaller.cs
```diff
@@ -46,7 +46,7 @@ public sealed class NuGetPackageInstaller : INuGetPackageInstaller
         /// <param name="toolResolver">The NuGet tool resolver.</param>
         /// <param name="contentResolver">The content resolver.</param>
         /// <param name="log">The log.</param>
-        /// <param name="config">the configuration</param>
+        /// <param name="config">the configuration.</param>
```
patriksvensson: Good catch. Will update!

src/Cake.NuGet/NuGetPackageInstaller.cs
patriksvensson: But you're right. Will change this as well. Thanks!

src/Cake.NuGet/NuGetPackageInstaller.cs
```diff
@@ -149,7 +126,16 @@ public IReadOnlyCollection<IFile> Install(PackageReference package, PackageType
             }
 
             // Install the package.
-            InstallPackage(package, path);
+            if (!InstallPackage(package, path))
+            {
+                _log.Warning("An error occured while installing package {0}.", package.Package);
+                if (createdDirectory)
```
mholo65: Could we have a test for this path?

patriksvensson: I've added a test.

devlead:
@patriksvensson could you address feedback and rebase against latest develop.

patriksvensson:
@devlead Done

devlead:

PR #1760, opened 2017-08-19
GH1759: Omit xml declaration on XmlPoke it it didn't exist before

### LINKED ISSUE #1759: XmlPoke always writes the xmldeclaration even if the original file didn't have one
As the title says, when using XmlPoke, it always writes the declaration.

There are two possibilities:
1. Add an OmitXmlDeclaration flag to XmlPokeSettings to define if the declaration should be written or not.
2. Check in XmlPoke if the original one has a declaration and either write it or not.

Which one should we prefer?

Roemer:
… some issues with test resources.

Fixes #1759 (2nd version).


devlead:
@Roemer  your changes have been merged, thanks for your contribution 👍

PR #1765, opened 2017-08-25
Build behind proxy

aateeque:
This enables running the script behind a corporate proxy


gep13:
@aateeque is this something that should be added here: https://github.com/cake-build/resources

aateeque:
@gep13 thanks for pointing in the right direction. I'll close it here and open it there

Issue #1761, opened 2017-08-20
DoesForEach() extension method


david-driscoll:
I ran into a scenario where I wanted to run a set of actions over a collection, even in the event that one of more fails.  The main one is running unit tests, but there are other scenarios like handling many files with a common action, that I've found it handy.

For my code it's replaced anywhere I would use a `foreach` at the top of my `Does()` statement, hence calling the extension method `DoesForEach()`.

```csharp
public static CakeTaskBuilder<ActionTask> DoesForEach<T>(this CakeTaskBuilder<ActionTask> builder, IEnumerable<T> items, Action<T> action)
{
    return DoesForEach(builder, () => items, (i, c) => action(i));
}

public static CakeTaskBuilder<ActionTask> DoesForEach<T>(this CakeTaskBuilder<ActionTask> builder, IEnumerable<T> items, Action<T, ICakeContext> action)
{
    return DoesForEach(builder, () => items, action);
}
public static CakeTaskBuilder<ActionTask> DoesForEach<T>(this CakeTaskBuilder<ActionTask> builder, Func<IEnumerable<T>> items, Action<T> action)
{
    return DoesForEach(builder, items, (i, c) => action(i));
}

public static CakeTaskBuilder<ActionTask> DoesForEach<T>(this CakeTaskBuilder<ActionTask> builder, Func<IEnumerable<T>> items, Action<T, ICakeContext> action)
{
    builder.Does(c =>
    {
        var exceptions = new List<Exception>();
        foreach (var item in items())
        {
            try
            {
                action(item, c);
            }
            catch (Exception e)
            {
                exceptions.Add(e);
            }
        }
        if (exceptions.Any()) throw new AggregateException("Task failed with following exceptions", exceptions);
    });
    return builder;
}
```

devlead:
Sounds like a reasonable addition to cake, would you be willing to send in a PR for review?

david-driscoll:
Closing merged in #1767

PR #1767, opened 2017-08-27
Added new extension method DoesForEach

david-driscoll:
This adds two new extension methods for `ActionTask`
* `DoesForEach`
* `DeferOnError`

`DoesForEach` allows a collection of items, or a delegate that returns a collection of items, to be added as actions to a given task.  In the case of the delegate returning a collection, the delegate is only executed when the task is executed.  This allows the task to look up items that don't exist when the script is first run, such as build artifacts.

`DeferOnError` allows a given task the ability to defer all exceptions until the end of the task execution.  This way tasks can run all actions to completion before failing.  This can have some value when you work with something like multiple unit test projects, or multiple publishes.  You can see the output of the entire task, and not just where it failed.

Also makes some changes to `ActionTask`, I'm not sure if I should or not.  The reason I did make the changes was so that these two new extension methods work correctly in more scenarios.  One such case would have been given:

```csharp
Task("Awesome")
    .Does(() => { throw new Exception(); })
    .DoesForEach(GetFiles("**/*.txt"), (file) => { // My cool file action })
    .DeferOnError();
```

This way all the actions would execute, not just the action before the actions contained in `DoesForEach`.

Related: #1761

### DIFF
--- a/src/Cake.Core.Tests/Unit/ActionTaskTests.cs
+++ b/src/Cake.Core.Tests/Unit/ActionTaskTests.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using Cake.Core.Tests.Fixtures;
 using Xunit;
 
 namespace Cake.Core.Tests.Unit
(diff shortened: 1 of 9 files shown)


devlead:
@david-driscoll  your changes have been merged, thanks for your contribution 👍

PR #1768, opened 2017-08-29
GH1743: Implements functionality for downloading NuGet packages

### LINKED ISSUE #1743: Implement functionality in Cake.NuGet for downloading packages
Implement functionality in Cake.NuGet for downloading packages instead of shelling out to `nuget.exe`

Benefits:
* Package installation works in `Cake.CoreCLR`. Today we do not allow since we can't be sure mono is actually installed on Linux.
* Allows us to work around nasty MAX_PATH issues with `nuget.exe` when path becomes too long
* More control over actual name of folder where package is extracted

mholo65:
- Fixes #1743

TODO:
- [x] Make this opt-in through configuration param.
- [ ] Add integration tests.
- [x] Opt-in Load Dependencies (both config and directive)

### DIFF
--- a/src/Cake.NuGet.Tests/Unit/NuGetModuleTests.cs
+++ b/src/Cake.NuGet.Tests/Unit/NuGetModuleTests.cs
@@ -6,6 +6,7 @@ using Cake.Core.Composition;
 using Cake.Core.Packaging;
 using Cake.Core.Scripting.Processors.Loading;
 using Cake.NuGet.Tests.Fixtures;
+using Cake.Testing.Xunit;
 using NSubstitute;
 using Xunit;
 
@@ -20,7 +21,7 @@ namespace Cake.NuGet.Tests.Unit
             {
                 // Given
                 var fixture = new NuGetModuleFixture<NuGetContentResolver>();
-                var module = new NuGetModule();
+                var module = fixture.CreateModule();
 
                 // When
                 module.Register(fixture.Registrar);
@@ -31,13 +32,12 @@ namespace Cake.NuGet.Tests.Unit
                 fixture.Builder.Received(1).Singleton();
             }
 
-#if !NETCORE
-            [Fact]
+            [RuntimeFact(TestRuntime.Clr)]
             public void Should_Register_The_NuGet_Load_Directive_Provider()
             {
                 // Given
                 var fixture = new NuGetModuleFixture<NuGetLoadDirectiveProvider>();
-                var module = new NuGetModule();
+                var module = fixture.CreateModule();
 
                 // When
                 module.Register(fixture.Registrar);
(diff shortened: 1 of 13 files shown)


src/Cake.NuGet/Install/NuGetSourceRepositoryProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+
+namespace Cake.NuGet.Install
+{
+    internal sealed class NuGetSourceRepositoryProvider : ISourceRepositoryProvider
+    {
+        private static readonly string[] _defaultSources =
+        {
+            "https://api.nuget.org/v3/index.json"
```
gep13: @mholo65 should this come from the `cake.config` file?

devlead: Yes, I think it should.

mholo65: Yes it should!

src/Cake.NuGet.Tests/Unit/NuGetModuleTests.cs
```diff
@@ -65,6 +98,26 @@ public void Shouls_Register_The_NuGet_Package_Installer()
...
+                var fixture = new NuGetModuleFixture<NuGetPackageInstaller>();
+                fixture.Configuration.SetValue(Constants.NuGet.UseInProcessClient, bool.TrueString);
+                var module = fixture.CreateModule();
+
+                // When
+                module.Register(fixture.Registrar);
+
+                // Then
+                fixture.Registrar.Received(1).RegisterType<Install.NuGetPackageInstaller>();
+                fixture.Builder.Received(1).As<INuGetPackageInstaller>();
```
mholo65: @cake-build/cake-team can anyone tell me why this test is failing here?

mholo65:
@devlead rebased.

src/Cake.NuGet.Tests/Unit/NuGetModuleTests.cs
```diff
@@ -65,6 +98,26 @@ public void Shouls_Register_The_NuGet_Package_Installer()
...
                 fixture.Builder.Received(1).Singleton();
             }
+
+#pragma warning disable xUnit1004 // Test methods should not be skipped
+            [Fact(Skip = "Don't know why this is failing...")]
+#pragma warning restore xUnit1004 // Test methods should not be skipped
+            public void Should_Register_The_In_Process_NuGet_Package_Installer_If_Set_In_Configuration()
+            {
+                // Given
+                var fixture = new NuGetModuleFixture<NuGetPackageInstaller>();
```
devlead: This should be
```csharp
var fixture = new NuGetModuleFixture<Install.NuGetPackageInstaller>();
```

src/Cake.NuGet.Tests/Unit/NuGetModuleTests.cs
devlead: see above comment

src/Cake.NuGet.Tests/Unit/NuGetModuleTests.cs
mholo65: Ofc, stupid me looking at wrong place! Thank you!

mholo65:
@devlead fixed

devlead:
@mholo65 
Found one thing on Windows, the nuget client libs doesn't seem to be able to use sources with encrypted passwords. 

And it doesn't respect if you place a NuGet.config in repo root/tools folder/or cake folder. 

Interestingly the source it complains about is disabled, so it shouldn't be used, if i remove the disabled sources and only have 
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="NuGetV3" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```

Feels like an issue with the NuGet client libs.

I don't get the same errors as I got on Linux so I'll rerun the tests there if it was something temporary.

mholo65:
@devlead exception should be fixed... Will re-investigate re-download. If nothing helps, best would be to set `PackageSaveMode` to `NuPkg`. Any objections, @gep13?

devlead:
@mholo65 I can confirm exception is sorted and it now works for me the first time.
But the second time I execute Cake with tools & addins restored in place it fails with:
```Error: System.AggregateException: One or more errors occurred. ---> System.InvalidOperationException: Package 'gitreleasemanager 0.5.0' is not found in the following primary source(s): '/pullrequest/tools'. Please verify all your online package sources are available (OR) package id, version are specified correctly.```

mholo65:
@devlead I blame NuGet client libs and awful xplat support when it comes to file system. This breaks support for nuspec package save mode since nuspec is stored to disk as normalized (e.g. filename to lower invariant). But then when it checks if package is installed, it expects to find nuspec in non-normalized form.

mholo65:
@devlead @gep13 rebased. Please take it for a spin.

devlead:

PR #1774, opened 2017-08-31
GH1773: Fix NuGetHasSource ArgumentCustomization & add ConfigFile to NuGetSourcesSettings

### LINKED ISSUE #1773: NuGetHasSource call do not take care of ArgumentCustomization in NuGetSourcesSettings
### What You Are Seeing?
When I am using NuGetHasSource with NuGetSourcesSettings including ArgumentCustomization like in the example below the ArgumentCustomization will not applied to the call of the nuget.exe.

```
var nugetSourceSettings = new NuGetSourcesSettings
{
	UserName = "VssSessionToken",
	Password = accessToken,
	IsSensitiveSource = true,
	Verbosity = NuGetVerbosity.Detailed,
	ArgumentCustomization = args => args.Append("-ConfigFile " + solutionDirectory + "/NuGet.config")
};

if (!NuGetHasSource(internalFeed, nugetSourceSettings))
{
	Information("Add internal feed '{0}' as nuget source", internalFeed);
	NuGetAddSource("Internal", internalFeed , nugetSourceSettings);
}
```

**Console Output:**
> 
Executing: "C:/Daimler/FB/fb-timemgmt-service/Build/tools/NuGet.exe" sources List
Registered Sources:

  1.  nuget.org [Enabled]
      https://api.nuget.org/v3/index.json
  2.  Microsoft Visual Studio Offline Packages [Enabled]
      C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\
  3.  Microsoft Azure Service Fabric SDK [Enabled]
      C:\Program Files\Microsoft SDKs\Service Fabric\packages

### What is Expected?
What I expected is that the nuget.exe will be called like this:

> 
Executing: "C:/Daimler/FB/fb-timemgmt-service/Build/tools/NuGet.exe" sources List -ConfigFile C:/[Path to Repository]/[solutionDirectory]/NuGet.config
Registered Sources:
...

I already checked the implementation of [NuGetSources class](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/NuGet/Sources/NuGetSources.cs) and it seems the Run function call (L141+142) in the method 'HasSource' will be called with no valid ProcessArgumentBuilder (is null).

0.21.1
64-Bit
Windows
VSTS
See above
### Output Log
No log

devlead:
* Addresses issue with ArgumentCustomization not being used with NuGetHasSource 
* Adds ConfigFile to NuGetSourcesSettings so one can work against custom NuGet.config
* Adds NuGetSourcesSettings  ConfigFile unit tests
* Adds ArgumentCustomization unit test for NuGetHasSource 
* Adds NuGet sources integration tests (_.NET Core on posix excluded as nuget.exe won't run ootb there_)
* Adds NonInteractive which was missing from NuGetHasSource (user input no good in build scripts 😎 )
* Fixes #1773

### DIFF
--- a/src/Cake.Common/Tools/NuGet/Sources/NuGetSourcesSettings.cs
+++ b/src/Cake.Common/Tools/NuGet/Sources/NuGetSourcesSettings.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Cake.Core.IO;
 using Cake.Core.Tooling;
 
 namespace Cake.Common.Tools.NuGet.Sources
@@ -44,5 +45,10 @@ namespace Cake.Common.Tools.NuGet.Sources
         /// <c>true</c> if password is stored as unencrypted; otherwise, <c>false</c>.
         /// </value>
         public bool StorePasswordInClearText { get; set; }
+
+        /// <summary>
+        /// Gets or sets the location of the NuGet configuration file. If not specified, file %AppData%\NuGet\NuGet.config is used as configuration file.
+        /// </summary>
+        public FilePath ConfigFile { get; set; }
     }
 }
--- a/src/Cake.Common.Tests/Unit/Tools/NuGet/Sources/NuGetSourcesTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NuGet/Sources/NuGetSourcesTests.cs
@@ -271,6 +271,21 @@ namespace Cake.Common.Tests.Unit.Tools.NuGet.Sources
                 Assert.Equal("sources Add -Name \"name\" -Source \"source\" " +
                              "-NonInteractive -StorePasswordInClearText", result.Args);
             }
+
+            [Fact]
+            public void Should_Add_ConfigFile_To_Arguments_If_Set()
+            {
+                // Given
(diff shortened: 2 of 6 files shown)


src/Cake.Common/Tools/NuGet/Sources/NuGetSourcesSettings.cs
```diff
@@ -44,5 +45,10 @@ public sealed class NuGetSourcesSettings : ToolSettings
         /// <c>true</c> if password is stored as unencrypted; otherwise, <c>false</c>.
         /// </value>
         public bool StorePasswordInClearText { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value the NuGet configuration file. If not specified, file %AppData%\NuGet\NuGet.config is used as configuration file.
```
gep13: @devlead that wording here is a little off.  Perhaps `Gets or sets the location of the NuGet configuration file`

devlead: agreed I'll fix.

devlead: this should be addressed now.

gep13:

Issue #1751, opened 2017-08-15
Cake on .Net core


trondhindenes:
According to the documentation, Cake can run on .Net core. However, when running the build.sh package for linux, it requires mono.
How can I run a cake script without relying on mono on Linux?

patriksvensson:
@trondhindenes You will need to install the `Cake.CoreClr` package instead of the `Cake`. We currently do not have any bootstrappers for Cake.CoreClr that doesn't use NuGet to fetch packages so you will need to download the packages by some other means.

I think @devlead have written scripts for this.

devlead:
Here's an example bootstrapper for running only on .NET Core 
https://github.com/devlead/BitbucketPipelinesShield/blob/master/build.sh

trondhindenes:
Thanks!
I'm getting this: The specified framework 'Microsoft.NETCore.App', version '1.0.4' was not found, I guess it's caused by outdated references in the Cake.CoreCLR. Nuget.org doesnt list any dependencies for the package, so it's hard to see if the newest versions are updated either. It would be great if you could publish more metadata to nuget.

devlead:
Which version of the .NET CLI do you have installed? Package has no dependencies as everything but the .NET CLI is in the package.

trondhindenes:
The script which the bootstrap installer downloads is not compatible with my OS (Ubuntu 17)  it seems, so I commented it out - hoping that I could use one of the installed .net core versions already on my computer instead - which didn't work. But no worries, I can install some older CoreFX runtimes/SDKs tomorrow and try again then. Will report back. I guess the script could be a made to check if a framework download is needed before it goes ahead and downloads it, but no biggie. I'll report back once I've looked more at this in the morning.

devlead:
The example supplied script downloads 0.19.3,  latest version of Cake is 0.21.1 so you could try getting latest version if you haven't already. 
Which version of the .NET CLI & SDK do you have installed? Cake currently compiled with .NET CLI 1.1.0 and SDK 1.0.4.

trondhindenes:
```
The specified framework 'Microsoft.NETCore.App', version '1.0.4' was not found.
  - Check application dependencies and target a framework version installed at:
      /usr/share/dotnet/shared/Microsoft.NETCore.App
  - The following versions are installed:
      1.1.2
      1.1.1
```
and
```
~/Documents/projects/cakecorefx  dotnet --version
1.0.4
```

trondhindenes:
[...]

PR #1778, opened 2017-09-04
GH1779: Update xUnit to latest version

ErikSchierboom:
While working on #1751, I found out that older versions of xUnit don't play well with building with .NET CLI 2.0. This PR updates xUnit to the latest version, in which this problem is fixed. 

In the process, the xunit.analyzers dependency was also updated, which lead to a bunch of compiler warnings which were converted to errors when running the build script. This PR fixes all of those xUnit warnings, thereby making the build pass again. Each fix is in its own, separate commit, to allow for easier reviewing. All fixes were made automatically using the code fix provided by the xunit analyzer.

The one oddity is the xUnit1025 rule, which looks for duplicate theory parameters. It appears that I have found a bug in the analyzer, as it erronously marks different enum values with the same underlying (int) value as duplicates. I've added a `#pragma` to disable this warning in the (one) affected file.

Refs #1779

### DIFF
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -22,10 +22,10 @@
   <!-- Global packages -->
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
-    <PackageReference Include="xunit" Version="2.3.0-beta2-build3683" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta2-build1317" />
+    <PackageReference Include="xunit" Version="2.3.0-beta5-build3769" />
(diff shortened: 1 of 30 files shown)


ErikSchierboom:
@devlead I'll look into it! I've created [an issue](https://github.com/cake-build/cake/issues/1779).

ErikSchierboom:
@devlead With the help of @bradwilson (see [this thread](https://github.com/xunit/xunit/issues/1443#issuecomment-330073069)), I've managed to fix the issue. The build now passes on all platforms.

devlead:
@ErikSchierboom what the current status? ready for review? If so could you rebase against latest develop and squash commits.

ErikSchierboom:
@devlead Yes this is ready for review! I've squashed and rebased against the latest develop.

devlead:
@ErikSchierboom  your changes have been merged, thanks for your contribution 👍

PR #1783, opened 2017-09-06
[WIP review request] GH1770: Module attributes adding script namespaces and assemblies

### LINKED ISSUE #1770: Provide a way for modules to add default assemblies and namespaces for the script
I'd like to be able to have the module add itself to IScriptConventions.GetDefaultAssemblies. You can't do that without registering as the _only_ IScriptConventions instance, and you can't register as the only IScriptConventions instance without losing access to the previous instance of IScriptConventions that you need to delegate to.

That's a general frustration with the registration model, no way to just decorate instances, but what would be nice is some way to supplement the default assemblies and namespaces.

jnm2:
Please do not merge yet.

Closes #1770. I offered to add the assembly attributes to add namespaces and assembly references to the script from a module, as @patriksvensson wanted.

Implementation was very easy, but writing tests is not and I'd like guidance on the best way to do that.
Walking through the commits:

 - First three commits created the API shape discussed in #1770

 - Fourth commit added a test ensuring the **existing** addin namespace import was working. You didn't seem to have a test for this. Please especially review this one because I want feedback on the style of testing. (https://github.com/cake-build/cake/commit/5098fb7f610bb6bdefe976e0eab05a989d27f434)

 - Fifth commit decouples the module loader from the module searcher ahead of time for testing purposes- I wanted to be able to test code in the module loader without involving the searcher to make sure namespaces and assemblies would be imported

 - Sixth commit actually implements the new API.

 - Now I'm stuck. I want to test loading a module in the context of a `ScriptRunnerFixture` so that I can duplicate the test in https://github.com/cake-build/cake/commit/5098fb7f610bb6bdefe976e0eab05a989d27f434. But that requires referencing both Cake and Autofac from Cake.Core.Tests. If I put the test in Cake.Tests, I lose access to `ScriptRunnerFixture`. How would you go about testing this?

Thanks!

### DIFF
--- a/src/Cake.Core.Tests/Unit/Annotations/CakeNamespaceImportAttributeTests.cs
+++ b/src/Cake.Core.Tests/Unit/Annotations/CakeNamespaceImportAttributeTests.cs
@@ -17,7 +17,7 @@ namespace Cake.Core.Tests.Unit.Annotations
             public void Should_Throw_If_Namespace_Is_Null()
             {
                 // Given, When
-                var result = Record.Exception(() => new CakeNamespaceImportAttribute(null));
+                var result = Record.Exception(() => new CakeNamespaceImportAttribute((string)null));
 
                 // Then
                 AssertEx.IsArgumentNullException(result, "namespace");
--- a/src/Cake.Core.Tests/Unit/Annotations/CakeNamespaceImportAttributeTests.cs
+++ b/src/Cake.Core.Tests/Unit/Annotations/CakeNamespaceImportAttributeTests.cs
@@ -3,14 +3,61 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Reflection;
 using Cake.Core.Annotations;
+using Cake.Core.IO;
+using Cake.Core.Packaging;
+using Cake.Core.Scripting;
+using Cake.Core.Tests.Fixtures;
+using Cake.Testing;
+using NSubstitute;
 using Xunit;
+
+[assembly: CakeNamespaceImport("assembly-level namespace import 1")]
+[assembly: CakeNamespaceImport("assembly-level namespace import 2")]
 
 namespace Cake.Core.Tests.Unit.Annotations
 {
-    public sealed class CakeNamespaceImportAttributeTests
+    [CakeNamespaceImport("class-level namespace import 1")]
+    [CakeNamespaceImport("class-level namespace import 2")]
+    public static class CakeNamespaceImportAttributeTests
     {
+        [CakeMethodAlias]
+        [CakeNamespaceImport("method-level namespace import 1")]
+        [CakeNamespaceImport("method-level namespace import 2")]
+        public static void DummyAlias(this ICakeContext context)
+        {
+        }
+
+        [Theory]
+        [InlineData("method")]
+        [InlineData("class")]
+        [InlineData("assembly")]
+        public static void Imports_from_addin_into_session(string level)
+        {
+            var fixture = new ScriptRunnerFixture();
+            fixture.AliasFinder = new ScriptAliasFinder(fixture.Log);
+
+            // Necessary because the ScriptConventions class starts returning null in the assemblies list
+            // when it uses a test double assembly loader and tries to load System.Runtime
+            fixture.ScriptConventions = Substitute.For<IScriptConventions>();
+            fixture.ScriptConventions.GetDefaultAssemblies(Arg.Any<DirectoryPath>()).Returns(Array.Empty<Assembly>());
+
+            var thisAssembly = typeof(CakeNamespaceImportAttributeTests).GetTypeInfo().Assembly;
+            var thisAssemblyPath = new FilePath(thisAssembly.Location);
+            fixture.FileSystem.CreateFile(thisAssemblyPath);
+            fixture.AssemblyLoader.Load(Arg.Is<FilePath>(v => v.FullPath == thisAssemblyPath.FullPath), Arg.Any<bool>()).Returns(thisAssembly);
+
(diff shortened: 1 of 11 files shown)


src/Cake.Core/Scripting/ScriptConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <param name="namespace">The namespace to import in the script.</param>
+        public void AddScriptNamespace(string @namespace) => scriptNamespaces.Add(@namespace);
+
+        /// <summary>
+        /// Adds an assembly name to reference in the script.
+        /// </summary>
+        /// <param name="assemblyName">The assembly name to reference in the script.</param>
+        public void AddScriptAssembly(string assemblyName) => scriptAssemblyNames.Add(assemblyName);
+
+        IReadOnlyList<string> IScriptConfiguration.Namespaces => scriptNamespaces;
```
patriksvensson: Properties goes before methods (in Cake).

src/Cake.Core.Tests/Unit/Annotations/CakeNamespaceImportAttributeTests.cs
```diff
@@ -3,21 +3,68 @@
...
+        [CakeNamespaceImport("method-level namespace import 2")]
+        public static void DummyAlias(this ICakeContext context)
+        {
+        }
+
+        [Theory]
+        [InlineData("method")]
+        [InlineData("class")]
+        [InlineData("assembly")]
+        public static void Imports_from_addin_into_session(string level)
```
patriksvensson: Should be `Imports_From_Addin_Into_Session`.

patriksvensson: Test should be divided into `Given`, `When`, `Then`.

src/Cake.Core/Scripting/IScriptConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Provides the ability to configure the script via IoC.
+    /// </summary>
+    public interface IScriptConfiguration
+    {
+        /// <summary>
+        /// Gets additional namespaces to be imported in the script.
+        /// </summary>
+        IReadOnlyList<string> Namespaces { get; }
```
patriksvensson: Since order isn't important here, can we change this to `IReadOnlyCollection<string>`?

src/Cake.Core/Scripting/IScriptConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        /// <summary>
+        /// Gets additional assembly names to be referenced by the script.
+        /// </summary>
+        IReadOnlyList<string> AssemblyNames { get; }
```
patriksvensson: Since order isn't important here, can we change this to `IReadOnlyCollection<string>`?

src/Cake.Core/Scripting/ScriptConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Collections.Generic;
+
+namespace Cake.Core.Scripting
+{
+    /// <summary>
+    /// Provides the ability to configure the script via IoC.
+    /// </summary>
+    public sealed class ScriptConfiguration : IScriptConfiguration
+    {
+        private readonly List<string> scriptNamespaces = new List<string>();
```
patriksvensson: * Members start with `_`.
* Could we move initialization to constructor?

src/Cake.Core/Scripting/ScriptConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+        private readonly List<string> scriptAssemblyNames = new List<string>();
```
patriksvensson: * Members start with `_`.
* Could we move initialization to constructor?

src/Cake.Core/Scripting/ScriptConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Adds an assembly name to reference in the script.
+        /// </summary>
+        /// <param name="assemblyName">The assembly name to reference in the script.</param>
+        public void AddScriptAssembly(string assemblyName) => scriptAssemblyNames.Add(assemblyName);
+
+        IReadOnlyList<string> IScriptConfiguration.Namespaces => scriptNamespaces;
+
+        IReadOnlyList<string> IScriptConfiguration.AssemblyNames => scriptAssemblyNames;
```
patriksvensson: Properties goes before methods (in Cake).

src/Cake/Composition/ModuleLoader.cs
```diff
@@ -4,36 +4,55 @@
...
-using Cake.Core.IO;
+using Cake.Core.Scripting;
 
 namespace Cake.Composition
 {
-    internal sealed class ModuleLoader
+    /// <summary>
+    /// Loads module types by applying them to a container and script configuration.
+    /// </summary>
+    public sealed class ModuleLoader
```
patriksvensson: Any reason for this to be public?

src/Cake/Program.cs
```diff
@@ -68,8 +69,9 @@ public static int Main()
                     builder.Update(container);
 
                     // Load all modules.
-                    var loader = container.Resolve<ModuleLoader>();
-                    loader.LoadModules(container, options);
+                    var moduleSearcher = container.Resolve<ModuleSearcher>();
+                    var moduleLoader = container.Resolve<ModuleLoader>();
+                    moduleLoader.LoadModules(moduleSearcher.Search(options), container, container.Resolve<ScriptConfiguration>());
```
patriksvensson: What's the reason for this change (searching for modules outside of the ModuleLoader)?

jnm2:
I apologize, I won't have time to finish this.

PR #1786, opened 2017-09-06
GH1785: Bump LatestBreakingChange to 0.22.0

### LINKED ISSUE #1785: Bump LatestBreakingChange to 0.22.0

mholo65:
- Fixes #1785

### DIFF
--- a/src/Cake.Core/Constants.cs
+++ b/src/Cake.Core/Constants.cs
@@ -8,7 +8,7 @@ namespace Cake.Core
 {
     internal static class Constants
     {
-        public static readonly Version LatestBreakingChange = new Version(0, 16, 0);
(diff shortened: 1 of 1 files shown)


devlead:
Interestingly this will mean we can't build Cake with Cake 😉 
```
Error: The assembly 'Cake.Coveralls, Version=0.4.0.0, Culture=neutral, PublicKeyToken=null'
is referencing an older version of Cake.Core (0.17.0). This assembly need to reference at least Cake.Core version 0.22.0.
Another option is to downgrade Cake to an earlier version.
```
Launching with `mono ./tests/integration/tools/Cake/Cake.exe --settings_skipverification=true` works though. But we'll need to communicate this asap.

@cake-build/cake-team Also I'm think we should add this information to the error message.

devlead:

PR #1789, opened 2017-09-08
GH1787: Detailed assembly version verification message

### LINKED ISSUE #1787: Add opt-out config information to assembly version verification error message
Currently we get an error like
```
Error: The assembly 'Cake.Coveralls, Version=0.4.0.0, Culture=neutral, PublicKeyToken=null'
is referencing an older version of Cake.Core (0.17.0). This assembly need to reference at least Cake.Core version 0.22.0.
Another option is to downgrade Cake to an earlier version.
```
A third option is opt out on own risk using Skip Verification settings
```
--settings_skipverification=true
```
we should add this to the message,

devlead:
* Adds opt-out configuration information in error message
* Outputs detailed debug information about assemblies that verifier would throw on if enabled
* fixes #1787

If opt-out you see something like this so you know which assemblies are referencing the wrong Cake version
image

On error it'll just as before fail & throw on first assembly
image

### DIFF
--- a/src/Cake.Core/Reflection/AssemblyVerifier.cs
+++ b/src/Cake.Core/Reflection/AssemblyVerifier.cs
@@ -23,11 +23,7 @@ namespace Cake.Core.Reflection
 
         public void Verify(Assembly assembly)
         {
-            if (_skipVerification)
-            {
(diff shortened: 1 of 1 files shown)


gep13:

gep13:
@patriksvensson sorry, didn't see your comment on my phone before meeting :-(

Issue #1625, opened 2017-06-01
Comma in msbuild commands are not escaped 


Julien-Mialon:
### What You Are Seeing?
MSBuild generated commande line is invalid when property contains a comma.

### What is Expected?
Valid generated command line

### What version of Cake are you using?
v0.19.5

### Are you running on a 32 or 64 bit system?
64 bits

### What environment are you running on?  Windows? Linux? Mac?
Mac

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)
I made a simple cake script available there https://gist.github.com/Julien-Mialon/1918cc8c9db470558eaa8e0f9fdfb77b to reproduce the problem

### Output Log
Here is the generated command

~~~sh
Executing: /Library/Frameworks/Mono.framework/Versions/Current/Commands/msbuild /v:normal /p:CodesignKey='iPhone Distribution: Company, SA (XXXXXXXXXX)' /target:Build "/Users/Julien/Documents/Source/tmp/cake/test.sln"

Microsoft (R) Build Engine version 15.2.0.0 (xplat-2017-02/c2edfeb Thu May 18 13:58:03 EDT 2017)
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1006: Property is not valid.
Switch:  SA (XXXXXXXXXX)

For switch syntax, type "MSBuild /help"
~~~

Jenkins has a similar issue and resolve it by replacing every comma by %2c (link to issue for jenkins : https://issues.jenkins-ci.org/browse/JENKINS-5552?focusedCommentId=143769&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-143769)

I would be happy to propose a PR, if the solution to replace comma by %2c is accepted.

devlead:
@Julien-Mialon your suggestion sounds valid, so a PR sounds reasonable, there's a few more special values which probably also could be covered by the PR
* https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-special-characters
* https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-escape-special-characters-in-msbuild

also looks like uppercase hex is preferred.

Julien-Mialon:
@devlead I'm not sure all of specials characters should be escaped because they could be valid values for msbuild properties. The only two characters that really cause issues are comma (,) and semicolon (;)

[...]

PR #1793, opened 2017-09-09
GH1625 Escape comma and semicolon in msbuild property values

Julien-Mialon:
Replace comma and semicolon in msbuild properties values by hex equivalent to fix issue #1625


devlead:
@Julien-Mialon your changes have been merged, thanks for your contribution 👍

PR #1819, opened 2017-09-21
GH1818: Support task dependees (reverse dependencies)

### LINKED ISSUE #1818: Support task dependees (reverse dependencies)
It would be nice to let tasks specify that they are dependencies of other tasks. It would also be good if we could tell whether dependencies are required or not. This would open up for better extensibility and specialized build systems like `Cake.Recipe`.

We could call this method `IsDependeeOf` (suggestions for other names are greatly appreciated).
I would also suggest that we add support for optional dependencies while we're at it.

**Example:**

```csharp
Task("A");
Task("B").IsDependentOn("A", required: true);
Task("C").IsDependentOn("B").IsDependeeOf("D");
Task("D").IsDependeeOf("E", required: false);

RunTarget("D");
```

Woud result in the following task traversal.

`A` -> `B` -> `C` -> `D`

**Example 2:**

This would allow for custom build systems distributed via NuGet let their consumers add custom targets and tell Cake when to run them.

```csharp
#load nuget:?package=supercoolbuildsystem&version=1.0.0

Task("My-Unit-Tests")
  .IsDependeeOf("Run-Unit-Tests")
  .Does(() =>
{
  // Do stuff here
});

SuperCoolBuildSystem.Run();
```

Would result in the following hypothetical task traversal.

`Clean` -> `Build` -> `My-Unit-Tests` -> `Run-Unit-Tests` -> `RunStuff`

Using an optional dependency, `SuperCoolBuildSystem` could do the same thing, by adding an optional dependency on `My-Unit-Tests` and thus make it convention based.

patriksvensson:
* Added support for reverse dependencies.
* Dependencies can now be optional.

```csharp
Task("A");
Task("B").IsDependentOn("A", required: true);
Task("C").IsDependentOn("B").IsDependeeOf("D");
Task("D").IsDependeeOf("E", required: false);

RunTarget("D");
```

Fixes #1818

### DIFF
--- a/src/Cake.Core/ICakeTaskInfo.cs
+++ b/src/Cake.Core/ICakeTaskInfo.cs
@@ -27,6 +27,12 @@ namespace Cake.Core
         /// Gets the task's dependencies.
         /// </summary>
         /// <value>The task's dependencies.</value>
-        IReadOnlyList<string> Dependencies { get; }
+        IReadOnlyList<CakeTaskDependency> Dependencies { get; }
+
+        /// <summary>
+        /// Gets the tasks that the task want to be a dependency of.
+        /// </summary>
+        /// <value>The tasks that the task want to be a dependency of.</value>
+        IReadOnlyList<CakeTaskDependency> Dependees { get; }
     }
 }
--- a/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeEngineTests.cs
@@ -150,15 +150,19 @@ namespace Cake.Core.Tests.Unit
                 engine.RegisterTask("A").Does(() => result.Add("A"));
                 engine.RegisterTask("B").IsDependentOn("A").Does(() => result.Add("B"));
                 engine.RegisterTask("C").IsDependentOn("B").Does(() => result.Add("C"));
+                engine.RegisterTask("D").IsDependentOn("C").IsDependeeOf("E").Does(() => { result.Add("D"); });
+                engine.RegisterTask("E").Does(() => { result.Add("E"); });
 
                 // When
-                engine.RunTarget(fixture.Context, fixture.ExecutionStrategy, "C");
+                engine.RunTarget(fixture.Context, fixture.ExecutionStrategy, "E");
 
                 // Then
-                Assert.Equal(3, result.Count);
+                Assert.Equal(5, result.Count);
                 Assert.Equal("A", result[0]);
                 Assert.Equal("B", result[1]);
                 Assert.Equal("C", result[2]);
+                Assert.Equal("D", result[3]);
+                Assert.Equal("E", result[4]);
             }
 
             [Fact]
--- a/src/Cake.Core.Tests/Unit/CakeTaskBuilderExtensionsTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeTaskBuilderExtensionsTests.cs
@@ -25,70 +25,88 @@ namespace Cake.Core.Tests.Unit
                 // Then
                 Assert.Equal(1, task.Dependencies.Count);
             }
+
+            public sealed class OnMethodTaskBuilder
+            {
+                [Fact]
+                public void Should_Add_Dependency_To_Task()
+                {
+                    // Given
+                    var parentTask = new ActionTask("parent");
+                    var childTask = new ActionTask("child");
+                    var builder = new CakeTaskBuilder<ActionTask>(parentTask);
+                    var cakeTaskBuilder = new CakeTaskBuilder<ActionTask>(childTask);
+
+                    // When
+                    builder.IsDependentOn(cakeTaskBuilder);
+
+                    // Then
+                    Assert.Equal(1, parentTask.Dependencies.Count);
+                }
+
+                [Fact]
+                public void Should_Add_Dependency_To_Task_With_Correct_Name()
+                {
+                    // Given
+                    var parentTask = new ActionTask("parent");
+                    var childTask = new ActionTask("child");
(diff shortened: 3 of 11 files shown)


src/Cake.Common/Tools/MSBuild/MSBuildBinaryLogSettings.cs
```diff
@@ -16,17 +16,17 @@ namespace Cake.Common.Tools.MSBuild
     public class MSBuildBinaryLogSettings
```
patriksvensson: For some reason, I got StyleCop errors for this file so I updated it. I have no clue why...

src/Cake.Common/Tools/OpenCover/OpenCoverSettings.cs
```diff
@@ -67,7 +67,7 @@ public sealed class OpenCoverSettings : ToolSettings
         public bool MergeOutput { get; set; }
 
         /// <summary>
-        /// Gets or sets a list of directories where assemblies being loaded from will be ignored.
+        /// Gets a list of directories where assemblies being loaded from will be ignored.
```
patriksvensson: For some reason, I got StyleCop errors for this file so I updated it. I have no clue why...

src/Cake.Common/Tools/MSBuild/MSBuildBinaryLogSettings.cs
mholo65: Those errors are correct. However, don't know why they haven't been caught in CI before. Could we possibly keep this PR focused and move the StyleCop issues to a separate PR?

patriksvensson: I can remove the changes.

src/Cake.Core/CakeEngine.cs
```diff
@@ -59,6 +59,7 @@ public CakeTaskBuilder<ActionTask> RegisterTask(string name)
                 const string format = "Another task with the name '{0}' has already been added.";
                 throw new CakeException(string.Format(CultureInfo.InvariantCulture, format, name));
             }
+
```
mholo65: This is probably also a StyleCop issue not related to this PR?

patriksvensson: No, I just tidied up a little :D

patriksvensson: Will skip this change.

devlead:
@patriksvensson this branch is out of date and has conflicts with develop, could you squash and rebase for final review?

patriksvensson:
@devlead Done

devlead:

src/Cake.Core/ICakeTaskInfo.cs
```diff
@@ -27,6 +27,12 @@ public interface ICakeTaskInfo
         /// Gets the task's dependencies.
         /// </summary>
         /// <value>The task's dependencies.</value>
-        IReadOnlyList<string> Dependencies { get; }
+        IReadOnlyList<CakeTaskDependency> Dependencies { get; }
```
AlgorithmsAreCool: Just as a note, this change broke our build. But it was an easy fix once we figured it out.

We have a very non-standard usage where we dynamically generate CakeTask objects as a pre-build step.

gep13: Sorry to hear that you had issues. Could you provide an example of what you are doing, and the change you had to make?

AlgorithmsAreCool: TL;DR : 
```c#
   if(!taskB.Dependencies.Contains(taskA.Name))
```
This line broke due to the highlighted line in the diff.

Once we ruled out a couple of other possible causes, the fix was simply:
```c#
   if(!taskB.Dependencies.Select(dep => dep.Name).Contains(taskA.Name))
```
in retrospect we could have also used
```c#
   if(!taskB.Dependencies.Any(dep => dep.Name == taskA.Name))
```

**Longer version**
Without going into extreme detail, we a have bunch of builds that are structurally very similar. Rather than defining N separate cake builds, we wrote a 'meta' layer in cake that takes a couple of params and then generates the needed cake tasks at build time. It actually works really well since Cake takes care of the dependency graph ordering for us. 

The crux of the operation looked like this before the update.
```c#
delegate CakeTemplate<TBuildContext>[] BuildFlowSelector<TBuildContext>(TBuildContext context) where TBuildContext : IBuildContext;

CakeTask ComposeFlow<TBuildContext>(BuildFlowSelector<TBuildContext> selector, IEnumerable<TBuildContext> contexts)
        where TBuildContext : IBuildContext
{
    var headOfTaskChain = contexts
        .SelectMany(ctx => selector(ctx).Select(template => GenerateTask(template,ctx)))
        .Aggregate((taskA,taskB) => {
            //Information($"{taskA.Name} -> {taskB.Name}");
            if(!taskB.Dependencies.Contains(taskA.Name))
                taskB.AddDependency(taskA.Name);
            return taskB;
        });

    return headOfTaskChain;
}
```
I know that is a bit dense but the issue was on this line 

```c#
   if(!taskB.Dependencies.Contains(taskA.Name))
```

This ensures the task isn't already linked and links if needed. Note that the inferred signature of `Contains` is `bool Contains<T>(this IEnumerable<T> enumerable, T item)` where `T` is string.

After Cake 0.23 went out `taskB.Dependencies` was no longer an `IReadOnlyList<string>` but now an ` IReadOnlyList<CakeTaskDependency>` as noted in the diff above. But the second argument of `Contains`  was still `taskA.Name` which was a string. The type mismatch caused the overload resolution to fail and make a bogus suggestion about IQueryable being the closest match. Which sent me down a short goose chase before i realized exactly what was going on.  

No real harm done, just figured i would bring it up.

Issue #1831, opened 2017-09-25
CleanDirectories Throws NullReferenceException When Token Is Null


jpreese:
### What You Are Seeing?
When calling CleanDirectories() with an invalid Glob pattern (in this case a trailing slash), a NullReferenceException is thrown.

### What is Expected?
The method should behave the same as unable to locate the directory.

### What version of Cake are you using?
0.22.2

### Are you running on a 32 or 64 bit system?
64 bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)
Run the following cake script

```
var target = Argument("target", "Default");

Task("Default")
  .Does(() =>
  {
    CleanDirectories("somedir/");
  });

RunTarget(target);
```

### Output Log

```
An error occurred when executing task 'Default'.
Error: System.AggregateException: One or more errors occurred. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Cake.Core.IO.Globbing.GlobParser.ParseSegment(GlobParserContext context)
   at Cake.Core.IO.Globbing.GlobParser.Parse(GlobParserContext context)
   at Cake.Core.IO.Globber.Match(String pattern, Func`2 predicate)
   at Cake.Common.IO.GlobbingAliases.GetDirectories(ICakeContext context, String pattern)
   at Cake.Common.IO.DirectoryAliases.CleanDirectories(ICakeContext context, String pattern)
   at Submission#0.CleanDirectories(String pattern)
   at Submission#0.<<Initialize>>b__0_0()
   at Cake.Core.ActionTask.Execute(ICakeContext context)
   at Cake.Core.DefaultExecutionStrategy.Execute(CakeTask task, ICakeContext context)
   at Cake.Core.CakeEngine.ExecuteTask(ICakeContext context, IExecutionStrategy strategy, Stopwatch stopWatch, CakeTask task, CakeReport report)
   at Cake.Core.CakeEngine.RunTarget(ICakeContext context, IExecutionStrategy strategy, String target)
   at Cake.Scripting.BuildScriptHost.RunTarget(String target)
   at Submission#0.<<Initialize>>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1832, opened 2017-09-25
GH1831: Verify CurrentToken is not null before checking Kind property

jpreese:
For #1831 

Discussed briefly in the CAKE gitter. This should be a better place for any discussions if they are needed.


devlead:
@jpreese your changes have been merged, thanks for your contribution 👍

jpreese:
Awesome! :tada:

PR #1837, opened 2017-09-26
GH1821: Implement NuGet list

### LINKED ISSUE #1821: Missing Cake method for nuget list
Although most NuGet cli commands are supported, the command `list` (https://docs.microsoft.com/en-us/nuget/tools/nuget-exe-cli-reference#list) is not.
Adding this one, will allow the investigate/list what packages a nuget server offers. All suboptions, as per nuget 4.3, are to be implemented offering support for semver 2.0

bravecobra:
- fixes #1821 
Added an implementation for the NuGet list method. Only the non-verbose output is used.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NuGet/List/NuGetListTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NuGet/List/NuGetListTests.cs
@@ -1,0 +1,250 @@
+﻿using Cake.Common.Tests.Fixtures.Tools.NuGet.List;
+using Cake.Testing;
+using Cake.Testing.Xunit;
+using Xunit;
+
+namespace Cake.Common.Tests.Unit.Tools.NuGet.List
+{
+    public sealed class NuGetListTests
+    {
+        public sealed class TheListMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Target_Package_Id_Is_Null()
+            {
+                // Given
+                var fixture = new NuGetListFixture
+                {
+                    PackageId = null
+                };
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                AssertEx.IsArgumentNullException(result, "packageId");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Settings_Are_Null()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.Settings = null;
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                AssertEx.IsArgumentNullException(result, "settings");
+            }
+
+            [Fact]
+            public void Should_Throw_If_NuGet_Executable_Was_Not_Found()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.GivenDefaultToolDoNotExist();
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                AssertEx.IsCakeException(result, "NuGet: Could not locate executable.");
+            }
+
+            [Theory]
+            [InlineData("/bin/nuget/nuget.exe", "/bin/nuget/nuget.exe")]
+            [InlineData("./tools/nuget/nuget.exe", "/Working/tools/nuget/nuget.exe")]
+            public void Should_Use_NuGet_Executable_From_Tool_Path_If_Provided(string toolPath, string expected)
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.Settings.ToolPath = toolPath;
+                fixture.GivenSettingsToolPathExist();
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal(expected, result.Path.FullPath);
+            }
+
+            [WindowsTheory]
+            [InlineData("C:/nuget/nuget.exe", "C:/nuget/nuget.exe")]
+            public void Should_Use_NuGet_Executable_From_Tool_Path_If_Provided_On_Windows(string toolPath, string expected)
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.Settings.ToolPath = toolPath;
+                fixture.GivenSettingsToolPathExist();
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal(expected, result.Path.FullPath);
+            }
+
+            [Fact]
+            public void Should_Throw_If_Process_Was_Not_Started()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.GivenProcessCannotStart();
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                AssertEx.IsCakeException(result, "NuGet: Process was not started.");
+            }
+
+            [Fact]
+            public void Should_Throw_If_Process_Has_A_Non_Zero_Exit_Code()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.GivenProcessExitsWithCode(1);
+
+                // When
+                var result = Record.Exception(() => fixture.Run());
+
+                // Then
+                AssertEx.IsCakeException(result, "NuGet: Process returned an error (exit code 1).");
+            }
+
+            [Fact]
+            public void Should_Find_NuGet_Executable_If_Tool_Path_Not_Provided()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("/Working/tools/NuGet.exe", result.Path.FullPath);
+            }
+
+            [Fact]
+            public void Should_Add_Mandatory_Arguments()
+            {
+                // Given
+                var fixture = new NuGetListFixture
+                {
+                    PackageId = "Cake"
+                };
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("list \"Cake\" -Verbosity Normal -NonInteractive", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_AllVersions_To_Arguments_If_True()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.Settings.AllVersions = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("list \"Cake\" -AllVersions -Verbosity Normal -NonInteractive", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_IncludeDelisted_To_Arguments_If_True()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.Settings.IncludeDelisted = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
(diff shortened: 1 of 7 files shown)


src/Cake.Common/Tools/NuGet/List/NuGetListItem.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Cake.Common.Tools.NuGet.List
+{
+    /// <summary>
+    /// An item as returned by NuGetList
+    /// </summary>
+    public class NuGetListItem
```
patriksvensson: Should be sealed.

src/Cake.Common/Tools/NuGet/NuGetAliases.cs
```diff
@@ -1100,5 +1101,33 @@ public static void NuGetInit(this ICakeContext context, string source, string de
...
+        /// <summary>
+        /// List packages on available from source using specified settings
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="packageId">The package Id</param>
+        /// <param name="settings">The settings.</param>
+        /// <returns>List of packages with their version</returns>
+        /// <example>
+        /// <code>
+        /// NuGetList("Cake", new NuGetListSettings {
```
patriksvensson: Would be good if the example showed that there are items being returned.

```csharp
var result = NuGetList("Cake", new NuGetListSettings {
    AllVersions = true
});

File: src/Cake.Common/Tools/NuGet/List/NuGetListItem.cs (hunk 1 above)
Comment (bravecobra) [reply src/Cake.Common/Tools/NuGet/List/NuGetListItem.cs]: Makes sense

File: src/Cake.Common/Tools/NuGet/NuGetAliases.cs (hunk 2 above)
Comment (bravecobra) [reply src/Cake.Common/Tools/NuGet/NuGetAliases.cs]: I've updated the example code with something more descriptive

File: src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs [hunk 3]
```diff
@@ -0,0 +1,47 @@
...
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether list all versions of a package.By default, only the latest package version is displayed
+        /// </summary>
+        public bool AllVersions { get; set; }
+
+        /// <summary>
+        /// Gets or sets the amount of details in the output: normal, quiet, detailed.
+        /// </summary>
+        public NuGetVerbosity? Verbosity { get; set; }
```
mholo65: Correct me if I'm wrong, but wont setting Verbosity  to either Quiet or Detailed mess up parsing? The parser expects one package per line with `id` and `version` separated by whitespace. Quiet verbosity won't output anything, while detailed will place them on separate rows plus additionally output some other noise.

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to allow unlisted packages to be shown.
+        /// </summary>
+        public bool IncludeDelisted { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to force the application to run using an invariant, English-based culture.
+        /// </summary>
+        public bool ForceEnglishOuput { get; set; }
```
mholo65: Is this needed?

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Gets or sets the NuGet configuration file.If not specified, file %AppData%\NuGet\NuGet.config is used as configuration file.
```
mholo65: Missing whitespace  `file.If`

src/Cake.Common.Tests/Unit/Tools/NuGet/List/NuGetListTests.cs
```diff
@@ -0,0 +1,250 @@
...
+                fixture.Settings.ConfigFile = "./nuget.config";
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("list \"Cake\" -ConfigFile \"/Working/nuget.config\" " +
+                             "-NonInteractive", result.Args);
+            }
+        }
```
mholo65: Could we also have some test for parsing the output?

src/Cake.Common/Tools/NuGet/List/NuGetList.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        /// <summary>
+        /// Lists available packages with their versions
+        /// </summary>
+        /// <param name="packageId">The source package id.</param>
+        /// <param name="settings">The settings</param>
+        /// <returns>A list of available packages</returns>
+        public IEnumerable<NuGetListItem> List(string packageId, NuGetListSettings settings)
+        {
+            if (packageId == null)
```
mholo65: This check is redundant. Next check will also catch `null`

src/Cake.Common/Tools/NuGet/List/NuGetList.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                throw new ArgumentNullException(nameof(packageId));
+            }
+            if (string.IsNullOrWhiteSpace(packageId))
```
mholo65: Not sure about this. Omitting the `packageId` is fully acceptable for `nuget.exe list`. However, calling it using a public feed like nuget.org might take a while. I can however see a use case when calling it using a small private feed with only a couple of packages. @cake-build/cake-team  any thoughts?

src/Cake.Common/Tools/NuGet/NuGetAliases.cs
```diff
@@ -1100,5 +1101,38 @@ public static void NuGetInit(this ICakeContext context, string source, string de
...
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+            var resolver = new NuGetToolResolver(context.FileSystem, context.Environment, context.Tools);
+            var runner = new NuGetList(context.FileSystem, context.Environment, context.ProcessRunner, context.Tools, resolver);
+            return runner.List(packageId, settings);
+        }
     }
```
mholo65: Could we also have  `IEnumerable<NuGetListItem> NuGetList(this ICakeContext context, string packageId)` overload.

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
devlead: No and if it's needed for parsing to be consistent across cultures then it should always be explicitly set.

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
devlead: yip, for consistency one should explicitly be set.

src/Cake.Common/Tools/NuGet/List/NuGetList.cs
bravecobra: Implemented this as that would indeed seem a valid use case.

src/Cake.Common/Tools/NuGet/List/NuGetList.cs
bravecobra: Removed the second check instead (see implementation of omiting the packageId)

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
bravecobra: Fixed that one

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
bravecobra: ForceEnglishOuput has been removed

src/Cake.Common/Tools/NuGet/List/NuGetListSettings.cs
bravecobra: Explicitly setting the Verbosity as that indeed broke the parsing.

src/Cake.Common.Tests/Unit/Tools/NuGet/List/NuGetListTests.cs
bravecobra: I've added a test that would check the output of the parsing. However I'm not yet completely happy, since there is no real error checking in the current implementation either although neither is there in the other code.

bravecobra: If the NuGet server returns an error (processExitCode != 0), then the script throws a Cake exception, before the parsing even begins. 
```
Executing: "C:/repo/tools/NuGet.exe" list "Cake" -Source "LocalServer" -Verbosity Normal -NonInteractive
The specified source 'LocalServer' is invalid. Please provide a valid source.
An error occurred when executing task 'Default'.
Error: System.AggregateException: One or more errors occurred. ---> Cake.Core.CakeException: NuGet: Process returned an error (exit code 1).
```
Should that exception be handled?

patriksvensson:
@bravecobra Can you squash the commits and rebase against develop?

src/Cake.Common.Tests/Unit/Tools/NuGet/List/NuGetListTests.cs
```diff
@@ -0,0 +1,250 @@
...
+                // Then
+                Assert.Equal("list \"Cake\" -Prerelease -Verbosity Normal -NonInteractive", result.Args);
+            }
+
+            [Fact]
+            public void Should_Add_Sources_To_Arguments_If_Set()
+            {
+                // Given
+                var fixture = new NuGetListFixture();
+                fixture.Settings.Source = new[] { "A;B;C" };
```
mholo65: Shouldn't this be `new[] { "A", "B", "C" };` in order to test that sources are concatenated corrected?

bravecobra: @mholo65 If it is wrong here, then it's wrong in the same test in https://github.com/cake-build/cake/blob/develop/src/Cake.Common.Tests/Unit/Tools/NuGet/Install/NuGetInstallerTests.cs as well (there might be other places as well), since I shamelessly copied it from there. It would have to be tested in this test class as well since it's a different property of a different class. I depended on the Cake team for their expertise in this case.
That being said, I did some quick tests with the cli. The dot-comma separated syntax is apparently perfectly valid. So is adding multiple -Source arguments. So basically, this test only checks whether the value is appended to the -Source argument. The value just happens to be valid syntax for multiple repositories. We need additional tests to validate the concatenation, here and in the other nuget tests of the other commands. Good spotting though! 👍 
I'd suggest opening a separate issue to fix it in the multiple places at the same time. That should indeed be a quick improvement. The test in itself is not invalid, it just doesn't cover the concatenation part, but that goes for the other ones as well.  If you want, you can assign me immediately to that new issue to improve the tests regarding the concatenation. I'd be happy to add them. Suggestions?

bravecobra: Note that multiple -Source arguments works, but it is not referenced/documented in https://docs.microsoft.com/en-us/nuget/tools/nuget-exe-cli-reference.

devlead:
@bravecobra your changes have been merged, thanks for your contribution 👍

PR #1839, opened 2017-09-26
GH1838: Add reference to nuget dependencies

### LINKED ISSUE #1838: Dependencies are installed but have no references added when using `LoadDependencies=true` with in process nuget client
### What You Are Seeing?

When using the new in process nuget client, and setting `nuget_loaddependencies=true`, the dependencies of addins are downloaded/installed but they are not actually referenced by the executing cake script causing errors about missing references.

### What is Expected?

Installed/Downloaded dependencies should also be referenced.

0.22.2

64-bit

Windows + Mac

Take this following script
```
#addin nuget:?package=Cake.Yaml&loaddependencies=true

Task ("OK").Does (() => {
    var yaml = "Key: value";
    var y = DeserializeYaml<YamlData> (yaml);
    Information ("Yaml key: {0}", y.Key);
});

public class YamlData
{
    public string Key { get;set;}
}

RunTarget ("OK");
```

### Output Log
~~~sh
An error occurred when executing task 'OK'.
Error: System.AggregateException: One or more errors occurred. ---> System.IO.FileNotFoundException: Could not load file or assembly 'YamlDotNet, Version=4.2.1.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.
  at Submission#0.DeserializeYaml[T] (System.String yaml) [0x0000c] in <e6597acf13834a3b9c292cd41029a93b>:0 
  at Submission#0.<<Initialize>>b__0_0 () [0x00007] in <e6597acf13834a3b9c292cd41029a93b>:0 
  at Cake.Core.CakeTaskBuilderExtensions+<>c__DisplayClass5_0.<Does>b__0 (Cake.Core.ICakeContext context) [0x00000] in...
~~~

Pull request coming...

Redth:
This fixes #1838

I think that simply the wrong package identity was used here.  Instead of `package` which is the identity of the package we are evaluating dependencies for, the `dependency` instance should be used instead which will give us the correct path to the actual libraries in the dependency package.

Using `dependency` just caused the actual main package to have multiple duplicate references added.

### DIFF
--- a/src/Cake.NuGet/Install/NuGetFolderProject.cs
+++ b/src/Cake.NuGet/Install/NuGetFolderProject.cs
@@ -105,7 +105,7 @@ namespace Cake.NuGet.Install
                         continue;
                     }
 
-                    var dependencyInstallPath = new DirectoryPath(_pathResolver.GetInstallPath(package));
+                    var dependencyInstallPath = new DirectoryPath(_pathResolver.GetInstallPath(dependency));
 
                     if (!_fileSystem.Exist(dependencyInstallPath))
                     {


Redth:
After a bit more debugging I'm also curious about why `_installedPackages` is used to find dependencies.  The behaviour of this is that each call to `GetFiles` that's made will look at _every single_ installed package and so we'll end up returning files that we returned in previous calls to this method.  Ultimately it seems to not matter as these duplicates seem to be resolved in the HashSet they end up in.  Just not the _fastest_ way to do this.

Redth:
@mholo65 this should fix what we were discussing in Gitter.

mholo65:
@Redth your changes have been merged. Thank you for your contribution!

PR #1844, opened 2017-09-26
Fixes StyleCop issues

### LINKED ISSUE #1841: StyleCop issues doesn't break build when running through Cake
StyleCop analyzer issues doesn't break build. They are only present as warnings in Visual Studio.

patriksvensson:
StyleCop should now run when running MSBuild from the commandline. Not sure why it didn't work but updating the StyleCop.Analyzers package solved it.

Also had to fix the `SA1623` and `SA1624` errors to make the CI build pass, so proving that this works is kind of difficult...

Fixes #1841


devlead:

PR #1857, opened 2017-10-05
GH1856: Adds support for MSBuild warnaserror & warnasmessage

### LINKED ISSUE #1856: Support MSBuild warnaserror and warnasmessage arguments
MSBuild as command line arguments to threat all errors as warnings, just some and to NOT treat given warnings as errors. I'll submit a PR shortly with a proposal to add support for this.

devlead:
* fixes #1856

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
@@ -946,6 +946,66 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
                 // Then
                 Assert.Equal(expected, result.Args);
             }
+
+            [Fact]
+            public void Should_Use_WarnAsError_If_Specified()
+            {
+                // Given
+                var expected = "/v:normal /target:Build /warnaserror \"C:/Working/src/Solution.sln\"";
+                var fixture = new MSBuildRunnerFixture(false, PlatformFamily.Windows);
+                fixture.Settings.WithWarningsAsError();
+
+                // When
+                var result = fixture.Run();
+
+                // Then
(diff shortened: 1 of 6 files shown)


src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
```diff
@@ -946,6 +946,51 @@ public void Should_Escape_Special_Characters_In_Property_Value(string propertyVa
...
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal(expected, result.Args);
+            }
+
+            [Fact]
+            public void Should_Use_WarnAsMessage_Codes_If_Specified()
```
gep13: @devlead do we need another test for the combination of both error and message?

devlead: Do we? "separate" things, happy to add if you think it's needed though.

gep13: Not saying that it is needed, but just asking the question, as it occurred to me as I was reading the code.  My thought process was that since these properties are intended to work together, a final overall test to show them working at the same time would be a good idea.  Perhaps overkill though.

devlead: @gep13 added `Should_Use_WarnAsError_And_WarnAsMessage_Codes_If_Specified()`

patriksvensson:
@devlead Your changes have been merged. Thank you for your contribution! 😉

devlead:
@gep13 addressed feedback and rebased.

Issue #1547, opened 2017-03-27
Escaping curly braces in log messages


DixonDs:
It seems that is not possible to escape curly braces properly in log messages. 

1. `Information("{} {0}", 123);` leads to error "Error: Input string was not in a correct format."
2. `Information("{{}} {0}", 123);` outputs "{{}} 123". The same input to `string.Format` yields "{} 123" which is what I want to get.

This is related to https://github.com/SharpeRAD/Cake.Powershell/issues/25 when the powershell output contains this particular string "{}"

kcamp:
I've made a first pass over this bug and think the changes I've made at 88b2f9d represent the appropriate solution.  I've added the tests that cover the two use-cases referenced above.

The one thing that I think this exposes is a bug in the `LiteralToken` that is being returned for escaped curlies - 
> Information("{{}} {0}", 123); outputs "{{}} 123"

@DixonDs is right in that this should probably output `{} 123` to be consistent with expectations, namely the implementation of `string.Format`.  I have adjusted the parser to return a single '{' or '}' in those cases and updated the tests to match.

If this seems reasonable, I will create a PR with the changes.

devlead:
@kcamp Can you submit a PR we can review?

kcamp:
@devlead Sure, will have that put together in the next day or so.

devlead:
@kcamp excellent! 👍

devlead:
Fixed by #1863

PR #1863, opened 2017-10-06
GH1547: Fix issue with parsing escaped curly braces in FormatParser

kcamp:
This addresses the issues described in GH-1547 regarding escaping curlies.

As detailed in the original issue, testing this fix uncovered an additional bug (inconsistency with `string.Format` implementation)

> The one thing that I think this exposes is a bug in the LiteralToken that is being returned for escaped curlies 
>`Information("{{}} {0}", 123); outputs "{{}} 123"`
>@DixonDs is right in that this should probably output {} 123 to be consistent with expectations, namely the implementation of string.Format. I have adjusted the parser to return a single '{' or '}' in those cases and updated the tests to match.


devlead:

Issue #1548, opened 2017-03-28
LogExtension colorizes output incorrectly


kcamp:
### What You Are Seeing
The new `LogExtension` overloads for `(string)` and `(object)` break the token paletting

### What is Expected?
The console output to be colorized consistent with 0.17 and lower.

### What version of Cake are you using?
0.19.1

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)
Use build script content such as:
```
_context.Information("  Updating assembly [FileVersion] metadata...");
_context.Information("========================================");
_context.Information("  Updating from {0} to {1}...", oldVersion, newVersion);
_context.Information("  Successs!");
```

Because of the way the methods were overloaded/cascaded, the console output now colorizes the `string` or `object` parameters as though they were formatting arguments.  In the left-hand capture, you can see that the entire line is colorized with forecolor/backcolor instead of just the replacements.

![image](https://cloud.githubusercontent.com/assets/649593/24419743/34a526ae-13be-11e7-9dcc-945d6a6c6e8f.png)

kcamp:
I think the easiest solution is to simply change the implementation for the new `string` and `object` overloads for the Debug/Verbose/Warning/Information/Error methods.

```
public static void Debug(this ICakeLog log, object value)
{
-  log.Debug("{0}", value);
+  log.Debug(string.Format("{0}", value));   // call the (string) overload
}

public static void Debug(this ICakeLog log, string value)
{
-  log.Debug("{0}", value);
+  log.Debug(value, null);  // cascade this call into the (string, params oject[]) overload
}
```

I have created a local build and this seems to be a reasonable solution.  The unit tests would need to be updated to meet the new expectations, as they currently fail.  Thoughts?  I can take on that work if it seems like a reasonable approach.

devlead:
@kcamp the overloads were added because issues with sending strings  with special characters, many people log generated code and json which fails if sent.
So they were added specifically if you only sent a string it wouldn't be passed to the tokenizer.
So this behavior is per design so object/strings wouldn't break logging which they did on several occasions  pre v0.18.0, and they were added in a way to not break ABI.

kcamp:
@devlead 
[...]

PR #1864, opened 2017-10-06
GH1548: Add logic to escape colorization for single {0} token

kcamp:
This addresses the regression identified in GH-1548.

Because of the changes that were made as part of 0.18.0, the `CakeBuildLog` will ultimately be consuming _some_ strings that consist solely of `{0}`.  In those cases, it doesn't make sense to change the `BackgroundColor` and `ForegroundColor` for the entire string (treating the entire string as a token argument)

These cases come into play in the `LogExtensions` methods like 
```
public static void Warning(this ICakeLog log, object value)
{
    log.Warning("{0}", value);
}
public static void Warning(this ICakeLog log, string value)
{
    log.Warning("{0}", value);
}
```

By introducing an extra check for the entire raw string being `{0}`, we can allow the output to be colorized in the same way it was prior to 0.18.0


devlead:

Issue #1855, opened 2017-10-03
SignTool is not found with latest windows 10 SDK


bravecobra:
With the latest Windows 10 SDK installed on a Windows 10 mahcine, the signtool.exe is no longer found by Cake. The new installation path is `C:\Program Files (x86)\Windows Kits\10\bin\10.0.15063.0\(x86|x64|arm|arm64)`. Note the extra version number and the platform folders.  there is an extra signtool under `C:\Program Files (x86)\Windows Kits\10\App Certification Kit`. None of these paths as covered in `src/Cake.Common/Tools/SignTool/SignToolResolver.cs`.

phillipsj:
@devlead @gep13 @bravecobra  happy to take this one.

patriksvensson:
@phillipsj Don't let me stop you! 😉

phillipsj:
@patriksvensson Any thoughts on the arm folders?  Seems like that would be a separate platform choice and right now only x86 and x64 are supported.

devlead:
Fixed by  #1868 and #1925

dskipper99:
I believe there is a recurrence of this particular issue.

Signing via ClickOnce is fine in Visual Studio, but when triggered from cake i receive the following:

  C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(4279,5): error MSB3482: An error occurred while signing: SignTool.exe was
 not found at path C:\Users\AG49101\OneDrive - Anthem\Documents\fep-m1cd-command-center\signtool.exe. [C:\Users\AG49101\OneDrive - Anthem\Documents\fep-m1cd-command-center\M1CD.vbproj]

Is there a workaround for this issue?

PR #1868, opened 2017-10-08
GH1855: Adding support for additional locations.

phillipsj:
Relates to #1855.

I did not add any ARM paths.  I didn't see the best way to handle that as there isn't any explicit ARM flags inside of the environment that I could find.

### DIFF
--- a/src/Cake.Common/Tools/SignTool/SignToolResolver.cs
+++ b/src/Cake.Common/Tools/SignTool/SignToolResolver.cs
@@ -68,6 +68,8 @@ namespace Cake.Common.Tools.SignTool
             {
                 // 64-bit specific paths.
                 files.Add(programFilesPath.Combine(@"Windows Kits\10\bin\x64").CombineWithFilePath("signtool.exe"));
+                files.Add(programFilesPath.Combine(@"Windows Kits\10\bin\10.0.15063.0\x64").CombineWithFilePath("signtool.exe"));
+                files.Add(programFilesPath.Combine(@"Windows Kits\10\bin\10.0.14393.0\x64").CombineWithFilePath("signtool.exe"));
                 files.Add(programFilesPath.Combine(@"Windows Kits\8.1\bin\x64").CombineWithFilePath("signtool.exe"));
                 files.Add(programFilesPath.Combine(@"Windows Kits\8.0\bin\x64").CombineWithFilePath("signtool.exe"));
                 files.Add(programFilesPath.Combine(@"Microsoft SDKs\Windows\v7.1A\Bin").CombineWithFilePath("signtool.exe"));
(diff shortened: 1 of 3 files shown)


patriksvensson:
@phillipsj You're on fire with all your PR:s 😄 👍 
Would it be possible to add some tests for this as well?

phillipsj:
Yep i will get some extra ones added.

src/Cake.Common/Tools/SignTool/SignToolResolver.cs
```diff
@@ -67,7 +67,9 @@ private FilePath GetFromDisc()
             if (_environment.Platform.Is64Bit)
             {
                 // 64-bit specific paths.
-                files.Add(programFilesPath.Combine(@"Windows Kits\10\bin\x64").CombineWithFilePath("signtool.exe"));
+                files.Add(programFilesPath.Combine(@"Windows Kits10\bin\x64").CombineWithFilePath("signtool.exe"));
```
mholo65: Was this intentional? Is the folder really named `Windows Kits10`? If yes then same thing should be done on 32-bit side, right?

src/Cake.Common/Tools/SignTool/SignToolResolver.cs
```diff
@@ -67,7 +67,9 @@ private FilePath GetFromDisc()
...
+                files.Add(programFilesPath.Combine(@"Windows Kits\10\bin\10.0.15063.0\x64").CombineWithFilePath("signtool.exe"));
+                files.Add(programFilesPath.Combine(@"Windows Kits\10\bin\10.0.14393.0\x64").CombineWithFilePath("signtool.exe"));
```
mholo65: Instead of hardcoding versions (which would make us add additional path everytime a new version of Win10 SDK is released) could we instead search using pattern `Windows Kits/10/bin/*/x64/signtool.exe`?

phillipsj:
Nice suggestions on both.  Thoughts on ARM support?

phillipsj:
@patriksvensson I have added tests.  @mholo65 I have fixed the typo and I tried adding the wildcard, however, the GetFile on the Filesystem object is throwing an error.  I would have to use something like GetFiles, then pull out the paths that way.  That didn't feel like the best way to handle it.

phillipsj:
@mholo65 @patriksvensson Does it look good?  Any changes that I need to make?  See my previous comment about using wildcards.

patriksvensson:
Sorry for not getting back to you about this. Been super busy lately but will set aside some time asap for this.

phillipsj:
@patriksvensson no worries, just making sure I am not holding it up.

Thealexbarney:
Hard-coding the SDK versions doesn't seem like the best method. This PR would already be out-of-date with the latest SDK release. The installed Windows 10 SDK versions are are listed as subkeys of `HKLM\SOFTWARE\Microsoft\Windows Kits\Installed Roots`. Those could be used instead of hard-coded versions.

Thealexbarney:
I made my own solution with PR #1925. I wasn't sure if I should opened a new PR or somehow attach it to this one, given that the methods of resolving the SDK are different between the two PRs.

devlead:
@Thealexbarney agreed, I'll merge this PR and then you can rebase your PR and I'll review that one.

devlead:
@phillipsj your changes have been merged, thanks for your contribution 👍

Issue #1908, opened 2017-11-03
Octopus Deploy tool does not support list-deployments call for octo.exe


CRuppert:
### What You Are Seeing?
missing feature for list-deployments implementation for octo.exe
### What is Expected?
implementation with filter support for octo.exe list-deployments
### What version of Cake are you using?
Current
### Are you running on a 32 or 64 bit system?
All
### What environment are you running on?  Windows? Linux? Mac?
Windows
### Are you running on a CI Server?  If so, which one?
Jenkins

### How Did You Get This To Happen? (Steps to Reproduce)

used the product. ;)

Feature request, See Above
Discussed in the gitter w/ @gep13 

PR forthcoming.

gep13:
Gitter discussion was here: https://gitter.im/cake-build/cake?at=59fcd032e44c43700ab46105

devlead:
@CRuppert I've added this to the v0.27.0 milestone. The next release v0.26.0 will basically be focused on .NET Core 2.0, asap after that's released I'll review your PR and if all's ok pull it in.

PR #1909, opened 2017-11-03
GH1908: Octo improvements - list-deployments implementation for octo.exe

CRuppert:
reference feature request #1908 

I am guessing you might want something cleaner for the stuff for the string in src/Cake.Common.Tests/Unit/Tools/OctopusDeploy/OctopusDeploymentQueryTests.cs line 260+.

I'm open to ideas as to how you approach that, since most of the source I read was write-heavy and doesn't do too much parsing, relying mostly on return codes when you have to shell out. I figured that I'd leave it until you said otherwise, since it's pretty darn specific and going the whole embedded resource route or similar seemed a bit overkill.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/OctopusDeploymentQueryierFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/OctopusDeploymentQueryierFixture.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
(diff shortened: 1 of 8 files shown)


CRuppert:
You should find I have signed the CLA at this point. :)

gep13:
@dnfclas @dnfgituser can you please confirm that this is the case?

@CRuppert sorry, just need to make sure

gep13:
@CRuppert we have been seeing some issues with the CLA bot that we are using.

CRuppert:
I was preemptive on the license. my bad. :) now it's done for sure.

devlead:
@CRuppert  your changes have been merged, thanks for your contribution 👍

Issue #1915, opened 2017-11-10
Only set working directory on process runner if set in settings


patriksvensson:
We should not explicitly set a working directory if none have been specified in the settings.

So [this assignment](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessRunner.cs#L78) should only be made if `ProcessSettings.WorkingDirectory` have been set.

devlead:
fixed by #1935

PR #1916, opened 2017-11-13
Set working directory only when set in settings

AbderrahmaneLaribi:
We should not explicitly set a working directory if none have been specified in the settings. #1915

### DIFF
--- a/src/Cake.Core/IO/ProcessRunner.cs
+++ b/src/Cake.Core/IO/ProcessRunner.cs
@@ -69,17 +69,18 @@ namespace Cake.Core.IO
 
             // Get the working directory.
             var workingDirectory = settings.WorkingDirectory ?? _environment.WorkingDirectory;
-            settings.WorkingDirectory = workingDirectory.MakeAbsolute(_environment);
 
             // Create the process start info.
             var info = new ProcessStartInfo(fileName)
             {
                 Arguments = arguments.Render(),
-                WorkingDirectory = workingDirectory.FullPath,
+                WorkingDirectory = settings.WorkingDirectory != null ? workingDirectory.FullPath : null,
                 UseShellExecute = false,
                 RedirectStandardError = settings.RedirectStandardError,
                 RedirectStandardOutput = settings.RedirectStandardOutput
             };
+
+            settings.WorkingDirectory = workingDirectory.MakeAbsolute(_environment);
 
             // Add environment variables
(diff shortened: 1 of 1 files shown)


src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -69,18 +69,19 @@ public IProcess Start(FilePath filePath, ProcessSettings settings)
...
             {
                 Arguments = arguments.Render(),
-                WorkingDirectory = workingDirectory.FullPath,
+                WorkingDirectory = settings.WorkingDirectory != null ? workingDirectory.FullPath : null,
                 UseShellExecute = false,
                 RedirectStandardError = settings.RedirectStandardError,
                 RedirectStandardOutput = settings.RedirectStandardOutput
             };
 
+            settings.WorkingDirectory = workingDirectory.MakeAbsolute(_environment);
```
patriksvensson: Doesn't reallty make sense to set this after it's been assigned to the `ProcessStartInfo`. This could be removed.

src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -69,18 +69,19 @@ public IProcess Start(FilePath filePath, ProcessSettings settings)
 
             // Get the working directory.
             var workingDirectory = settings.WorkingDirectory ?? _environment.WorkingDirectory;
```
patriksvensson: This could be changed to `var workingDirectory = settings.WorkingDirectory?.MakeAbsolute(_environment)`.

src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -69,18 +69,19 @@ public IProcess Start(FilePath filePath, ProcessSettings settings)
...
-            settings.WorkingDirectory = workingDirectory.MakeAbsolute(_environment);
 
             // Create the process start info.
             var info = new ProcessStartInfo(fileName)
             {
                 Arguments = arguments.Render(),
-                WorkingDirectory = workingDirectory.FullPath,
+                WorkingDirectory = settings.WorkingDirectory != null ? workingDirectory.FullPath : null,
```
patriksvensson: I think we should move the assignment and only assign if `settings.WorkingDirectory` is not null so we don't `touch it` unless we want to set it.

patriksvensson:
Closing this due to inactivity.

PR #1921, opened 2017-11-17
GH-1910: MSBuild property values should escape carriage return and line feed.

### LINKED ISSUE #1910: MSBuild property values should escape carriage return and line feed
Without escaping these, there's no way to set a multiline property value such as `PackageReleaseNotes` for the MSBuild [Pack target](https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets).

I saw that you're doing URL encoding when semicolons are present and indeed, the same trick works for CRLF:

```cake
var releaseNotes = ParseReleaseNotes("./ReleaseNotes.md").Notes;

MSBuild(BuildParameters.ProjectDir, new MSBuildSettings()
    .WithTarget("Pack")
    .WithProperty("PackageReleaseNotes", string.Join("%0D%0A", releaseNotes)));
```

However, I shouldn't be URL-encoding myself- what if one of the lines contains a semicolon? Then it probably gets URL-encoded again by Cake, breaking the manually-URL-encoded newlines.

We should be able to do `.WithProperty("PackageReleaseNotes", string.Join("\r\n", releaseNotes))` and have automatic URL-encoding just like we get for semicolons.

johankson:
Fixes #1910

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
@@ -1007,6 +1007,24 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
                 // Then
                 Assert.Equal(expected, result.Args);
             }
+
+            [Theory]
+            [InlineData("\r", "%0D")]
+            [InlineData("\n", "%0A")]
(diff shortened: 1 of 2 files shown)


patriksvensson:
Fixes #1910

patriksvensson:
@johankson Your changes have been merged. Thank you for your contribution! 😄

jnm2:
Yes, thanks for solving my problem! 😄

PR #1925, opened 2017-11-22
GH1855: Resolve Signtool in newer Windows 10 SDKs

Thealexbarney:
Addresses #1855 

Uses registry keys to find installed Windows 10 SDK versions.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/SignToolResolverFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/SignToolResolverFixture.cs
@@ -54,27 +54,15 @@ namespace Cake.Common.Tests.Fixtures.Tools
             }
         }
 
-        public void GivenThatToolExistInKnownPathWindows10SpecificSDK()
-        {
-            if (_is64Bit)
-            {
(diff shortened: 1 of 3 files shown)


Thealexbarney:
@devlead 
Regarding your comment [here](https://github.com/cake-build/cake/pull/1868#issuecomment-347528278), after rebasing, should I remove the hard-coded SDK versions added by #1868 in the resolver since they would be found via the registry keys?

devlead:
@Thealexbarney yes, sounds reasonable, but keep unit tests (if possible add any known missing cases). And make sure newest prioritized (might need sorting).

Thealexbarney:
Rebased on to latest `develop`. 

Updated changes:
- Use registry to find specific versions of Windows 10 SDKs instead of hard-coded version strings.
- Correct "App Certification Kit" path.
- Add a "temporary" fix to address #1926. The resolver still returns the first `signtool.exe` found, but I've rearranged the order it searches in. There are still no guarantees the the newest version will be returned. A more proper solution might check each file found and then return the newest.

devlead:
@Thealexbarney your changes have been merged, thanks for your contribution 👍

Issue #1930, opened 2017-11-27
The "out" parameters are not compiled properly.


LaoR-UA:
### What You Are Seeing?
When I create alias that accepts out parameters - cake will generate not working code.
For example, the method:

```
public interface IVersion
{
 public int Build {get;}
}

public class CusomVersion : IVersion
{
     public CustomVersion(int build)
     {
            Build = build;
     }
     public int Build {get;}
}

[CakeMethodAlias]
public static bool TryParse(this ICakeContext context, string versionToParse, out IVersion version)
{
.....
}
```

Will generate wrapper:
```
public System.Boolean TryParse(System.String versionToParse, out MyCake.Common.IVersion& version)
{
    return MyCake.Common.VersionHelper.TryParse(Context, versionToParse, out version);
}
```
And this leads to a lot of errors due to "&" symbol at the end of out parameter type declaration.

### What is Expected?
I expect to see properly generated wrapper and able to use "out" parameters.

### What version of Cake are you using?
Cake version is 0.23.0

### Are you running on a 32 or 64 bit system?
I am running on 64 bit Windows 10 system (client version, local machine).

jnm2:
Now that Cake is on Roslyn, what if you generated syntax nodes in code rather than generating text?

LaoR-UA:
I do not generate anything, it was done by Roslin on compile stage.

jnm2:
@LaoR-UA Sorry I wasn't clear; I was addressing the Cake team.

devlead:
Fixed by #1934

PR #1934, opened 2017-11-28
GH-1930: Fix ParameterEmitter to correctly handle out params

kcamp:
This adjusts the logic in the `ParameterEmitter` to correctly handle the case described in GH-1930 and adds test coverage for the same.

If there's additional test coverage desired, please let me know.


devlead:

PR #1936, opened 2017-11-28
Move to signing service v2

onovotny:
Note: you'll find the updated username and password in your lastpass. both should be treated as secure variables in your build system.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,15 +1,15 @@
 // Install addins.
-#addin "nuget:https://www.nuget.org/api/v2?package=Cake.Coveralls&version=0.7.0"
-#addin "nuget:https://www.nuget.org/api/v2?package=Cake.Twitter&version=0.6.0"
-#addin "nuget:https://www.nuget.org/api/v2?package=Cake.Gitter&version=0.7.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Cake.Coveralls&version=0.7.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Cake.Twitter&version=0.6.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Cake.Gitter&version=0.7.0"
 
 // Install tools.
-#tool "nuget:https://www.nuget.org/api/v2?package=gitreleasemanager&version=0.5.0"
-#tool "nuget:https://www.nuget.org/api/v2?package=GitVersion.CommandLine&version=3.6.2"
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -313,8 +313,13 @@ Task("Sign-Binaries")
...
         throw new InvalidOperationException("Could not resolve signing secret.");
     }
+    // Get the user.
+    var user = EnvironmentVariable("SIGNING_USER");
+    if(string.IsNullOrWhiteSpace(user)) {
+        throw new InvalidOperationException("Could not resolve signing user.");
+    }
 
-    var client = File("./tools/SignClient/tools/SignClient.dll");
+    var client = File("./tools/SignClient/tools/netcoreapp2.0/SignClient.dll");
```
devlead: Currently we're building using the 1.0.7 SDK(2.0 is on the way just not yet), so I suspect this version requires .NET Core 2.0? So we should use the netcoreapp1.1 moniker instead right?

onovotny: k

devlead: Excellent 👍  , I'll take this for a spin.

build.cake
```diff
@@ -329,12 +338,13 @@ Task("Sign-Binaries")
 
         // Build the argument list.
         var arguments = new ProcessArgumentBuilder()
-            .AppendQuoted(MakeAbsolute(client.Path).FullPath)
-            .Append("zip")
+            .AppendQuoted(signClientPath.FullPath)
+            .Append("sign")
```
onovotny: Yes, thanks missed this one, thanks for catching it!

mholo65:
@onovotny should this repo also be updated https://github.com/cake-build/bakery?

onovotny:
@mholo65 Yes, didn't see that, but should be the same set of changes. Anywhere you are signing with the existing service/config should be moved over.

BTW, the service does support VSIX's if you have any addin's/templates.

mholo65:
@onovotny thanks! I’ll update that one then.

Do we sign VSIX packages? @gep13 @agc93?

gep13:
@mholo65 no, we aren't currently signing VSIX packages.  We probably should.  Both the Visual Studio and Visual Studio Code packages.

onovotny:
Sign all the things! :)

devlead:
@onovotny your changes have been merged, thanks for your contribution 👍

gep13:
@onovotny @devlead was there a linked issue for this PR?

Issue #1939, opened 2017-11-30
Bug - TypeExtensions.GetFullName doesn't handle nested types correctly


kcamp:
While fixing GH-1930 the other night, some of the tests for `ParameterEmitter` caught my eye (seemed to be asserting the wrong "expected" values).  With fresh eyes, I wrote some red/green tests against the `MethodAliasGenerator` to prove what I was seeing was in fact wrong.

### What You Are Seeing?
The code generation for _some_ nested types is being emitted with the + sign in the type name.  i.e., `My.Namespace.ParentType+NestedType`.

Upon reviewing the code, I see that some of the work I did for GH-1566 accounted for this (presence of `+`) in the `GenericParameterConstraintEmitter` implementation;  this should have rightfully been made part of the `TypeExtensions.GetFullName` extension method, instead.

### What is Expected?
Valid, compilable code to be emitted for all types via the `TypeExtensions.GetFullName` method.

### What version of Cake are you using?
0.23

I will have a PR for this shortly.

devlead:
Fixed by #1940

PR #1940, opened 2017-11-30
Fix GH1939: Corrects the implementation of TypeExtensions.GetFullName

kcamp:
This addresses the issue described in GH-1939.

* Fixed the `GetFullName` method to account for nested types
* Remove now-extraneous applications of the same fix appearing in the `GenericParameterConstraintEmitter` type.
* Add test coverage to isolate nested type behavior


mholo65:
We’d probably need to port this to Bakery also.

PR #1942, opened 2017-12-05
GH1950: Allow Cake modules to be bootstrapped by Cake in a pre-processing phase.

patriksvensson:
I was reading @Redth's [blog post](https://redth.codes/cake-load-url/) and was thinking that it would be cool if modules could be installed via Cake itself as part of a pre-processing phase (while also honoring the `cake.config` conventions), so I put together this PR as a proof of concept and for discussion.

```
#module nuget:?package=Cake.UrlLoadDirective.Module&version=1.0.2

Information("Hello World!");
```

To install modules declared in a script, simply execute cake with the `--bootstrap` argument like this:

```
./cake.exe build.cake --bootstrap
```


src/Cake/Commands/BootstrapCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        public bool Execute(CakeOptions options)
+        {
+            // Get the script path and the root.
+            var path = options.Script.MakeAbsolute(_environment);
+            var root = path.GetDirectory();
+
+            // Analyze the script.
+            var result = _analyzer.Analyze(path);
```
gep13: @patriksvensson the result variable is not being used.  Should it be?

mholo65: @gep13 it is used. See next call, `result.Modules`

patriksvensson:
@devlead @gep13 @agc93 @mholo65 Should we also download NuGet as part of the bootstrapping process if we're not using the in-proc nuget installer?

gep13:
@patriksvensson as in, have the Cake.exe/Cake.CoreCLR.dll do that work, rather than have that in the bootstrapper files?

patriksvensson:
@gep13 Yes, exactly. That way we (in the new bootstrapper scripts) download Cake from GitHub directly and nothing more.

mholo65:
@patriksvensson I don’t think it’s Cake’s responsibility to download NuGet.exe and we’d like to get rid of that dependencybin Cake.NuGet.

patriksvensson:
@mholo65 True, but it's still widely used, and would be nice to reduce the logic in the bootstrapper scripts. Even with the in-proc installer, the need for nuget.exe won't disappear. We could make this an opt-in behavior as well that can be configured via the cake.config conventions.

patriksvensson:
The day where nuget.exe isn't needed anymore, we could simply deprecate this.

gep13:
@patriksvensson I am not against this.  The same checks that are being done in the bootstrapper script needs to also be done here though, i.e. check for the presence of NuGet.exe on the path, and not download if there.

mholo65:
@patriksvenssen, well it’s only Cake.NuGet which requires NuGet.exe to be available at bootstrap. For NuGetAliases, NuGet.exe can be downloaded with `#tool` directive (https://www.nuget.org/packages?q=Nuget.commandline).

As @gep13 said, the nuget.exe logic from bootstrapper should be moved into Cake, and honestly I don’t think it’s worth it since we’d want to remove the dependency long-term.

gep13:
>@mholo65 said...
For NuGetAliases, NuGet.exe can be downloaded with #tool directive (https://www.nuget.org/packages?

That's a good point, I hadn't considered that side of it.  This is something that I never do, as the bootstrapper always resolves it for me.  Perhaps I need to start being more explicit in this area.

phillipsj:
I like this being discussed as it is going to make what I am working on
easier.

devlead:
@cake-build/cake-team Code factor complains about duplication, code indeed looks very similar, is this code that we should extract and reuse in an internal matter?
image
https://github.com/cake-build/cake/blob/3e96871bb24d6257bba05fc7a09604ae49b2abc8/src/Cake/Composition/ModuleLoader.cs#L53-L73
https://github.com/cake-build/cake/blob/3e96871bb24d6257bba05fc7a09604ae49b2abc8/src/Cake/Commands/BootstrapCommand.cs#L52-L72

devlead:
@cake-build/cake-team would it make sense to have an config to opt-in to boostrap before executing script so one could just point Cake at a script and it would bootstrap and execute it in one go? If not should we validate if module doesn't exist? If module directive is specified should we limit what other modules are loaded?

Just trying to think of potential issues users might have / report.

gep13:
@devlead I like the idea of a config option for doing both.

patriksvensson:
I would prefer not to at this point. Would need to rewrite big parts of Cake.exe to allow it. Perhaps after a rewrite?

mholo65:
@patriksvensson no rewrite needed, just a simple `goto` :smile:
https://xkcd.com/292/

patriksvensson:
Updated PR based on feedback.

patriksvensson:
@cake-build/cake-team Any way of disabling the last codefactor warning since it won't happen atm? 😄

gep13:
@patriksvensson no, none that i am aware of.

mholo65:
@patriksvensson @gep13, we just silently ignore it and merge anyway.

gep13:
@mholo65 or, we could create an issue, in order to remember to come back and look at this :-)

gep13:
@patriksvensson was there an issue opened to go along with this PR?

patriksvensson:
@gep13 No issue for this. I'll create one for you.

patriksvensson:
@gep13 Added issue #1950

src/Cake/Commands/BootstrapCommand.cs
devlead: It's used but it should also check `result.Succeeded` if it's false it shouldn't continue to install modules. But it should abort, error log `result.Errors` and exit with a non zero code.

Something like:
```csharp
if (!result.Succeeded)
{
    throw new AggregateException(
        $"Bootstrap command failed to analyze script: {path}",
        result.Errors.Select(
            error => new CakeException($"File: {error.File}, Line: {error.Line}, Message: {error.Message}")));
}
```

Currently it just silently fails and reports success, which Isn't what users would expect.

patriksvensson:
@devlead Updated PR now.

devlead:

jnm2:
🎉

pitermarx:
thanks for this!

Issue #1943, opened 2017-12-06
Chocolatey package dependencies cannot be set using the ChocolateyPackSettings


RichardJFoster:
This problem affects Cake 0.23.0

The ChocolateyPackSettings class does not contain any mechanism to add package dependencies.

The expected operation is that the ChocolateyPackSettings has a Dependencies property that mirrors the Dependencies property already present in the NuGetPackSettings.

This problem was first observed on a 64 bit Windows system, but an examination of the source suggests this issue is not dependent on OS or build server type.

RichardJFoster:
So... I was looking at implementing this so I could submit a PR. The implementation is easy, but I am having difficulty working out how to unit test the change. It looks as if the unit tests are based on the abstract ChocolateyFixture (by way of the ChocolateyPackerFixture), but neither the ChocolateyPackerWithNuSpecFixture or the ChocolateyPackerWithoutNuSpecFixture currently appear to contain any test assertions at all. Am I looking in the right place?

gep13:
@RichardJFoster The Chocolatey Pack tests are here:

https://github.com/cake-build/cake/blob/develop/src/Cake.Common.Tests/Unit/Tools/Chocolatey/Pack/ChocolateyPackerTests.cs

Seems like you are looking in the Fixtures folder, which contains the setup logic for running the actual tests.

gep13:
If you look at this file:

https://github.com/cake-build/cake/blob/develop/src/Cake.Common.Tests/Unit/Tools/NuGet/Pack/NuGetPackerTests.cs

You will find the where the dependencies are tested in the NuGet version.

RichardJFoster:
Aha! Excellent. I was looking in the wrong place. :-) Thanks for the pointers.

gep13:
@RichardJFoster you are very welcome!  Let us know if you have any questions!

RichardJFoster:
No questions for now. Just waiting on the contribution license email, but if there are comments on the PR in the mean time I'm all ears. :-)

devlead:
Fixed by #1945

PR #1945, opened 2017-12-07
GH1943: Allow ChocolateyPackSettings to specify dependent packages

RichardJFoster:
This PR adds a Dependencies property, of type ICollection<ChocolateyNuSpecDependency>, to the ChocolateyPackSettings class (mirroring the existing behavior in the NuGet pack processing). If package dependencies are specified using this mechanism, the ChocolateyNuSpecTransformer has been modified to place those dependency records into the transformed NuSpec. The ChocolateyPackerTests have also been modified to verify that the transformed output matches expectations.

### DIFF
--- a/src/Cake.Common.Tests/Properties/Resources.resx
+++ b/src/Cake.Common.Tests/Properties/Resources.resx
@@ -143,6 +143,10 @@
     &lt;releaseNotes&gt;&lt;![CDATA[Line #1
 Line #2
 Line #3]]&gt;&lt;/releaseNotes&gt;
+    &lt;dependencies&gt;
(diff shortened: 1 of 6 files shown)


gep13:
@RichardJFoster I went ahead and rebased your branch against the latest upstream develop branch, which has the side effect of removing the merge commit that you had in there.  I will wait for the CI builds to finish, and if all goes well, I will get this merged in.

Thanks again for taking the time to do this work!

gep13:
Relates to #1943

RichardJFoster:
You're welcome @gep13. Hopefully it'll end up being be the first PR of many. :-)

Issue #1952, opened 2017-12-15
Add support for class/namespace/method arguments for XUnit2


kcamp:
It would be useful to have a way to natively support targeting namespace/class/method tests via the XUnit2 console runner.

```
  -method "name"         : run a given test method (should be fully specified;       
                         : i.e., 'MyNamespace.MyClass.MyTestMethod')                 
                         : if specified more than once, acts as an OR operation      
  -class "name"          : run all methods in a given test class (should be fully    
                         : specified; i.e., 'MyNamespace.MyClass')                   
                         : if specified more than once, acts as an OR operation      
  -namespace "name"      : run all methods in a given namespace (i.e.,               
                         : 'MyNamespace.MySubNamespace')                             
                         : if specified more than once, acts as an OR operation      
```

I've made a pass over this functionality; will submit PR shortly

devlead:
fixed by #1953

PR #1953, opened 2017-12-15
GH1952: Add support for targeting namespace/class/method via xunit.console runner

kcamp:
This addresses GH-1952.

This adds the properties to `XUnit2Settings` + correct extension methods to `XUnit2SettingsExtensions`, and the implementation to `XUnit2Runner` to pass these switches into the runner.


kcamp:
I see that CodeFactor is unhappy about method complexity.  I can certainly refactor or extract those bits of logic that were added, but I believe that to be a superficial change that does little to alter the underlying intent of the method.   Let me know the best way to proceed with changes, if any.

devlead:

Issue #1987, opened 2018-01-06
Confusing Error from Bad Format String to `Information()`


WalkerCodeRanger:
### What You Are Seeing?

`Information("    Line {1} didn't match", i);`
Prints:
```
    Line
An error occurred when executing task 'Test-Current'.
Error: One or more errors occurred.
        Index was outside the bounds of the array.
```
### What is Expected?
A format error like from `string.Format` making it clear that there was something wrong with the format string.
### What version of Cake are you using?
0.23.0
### Are you running on a 32 or 64 bit system?
64 Bit
### What environment are you running on?  Windows? Linux? Mac?
Windows
### Are you running on a CI Server?  If so, which one?
No
### How Did You Get This To Happen? (Steps to Reproduce)
In your `build.cake` file, have a line where the format string passed to `Information()` is invalid.  Like:
`Information("    Line {1} didn't match", i);`

devlead:
Fixed by #1990

PR #1990, opened 2018-01-09
GH-1987: Normalize exception handling in PropertyToken for invalid parameter index

kcamp:
This addresses the concern outlined in GH-1987 and uses the same exception type + message as `string.Format` for consistency.

Consider the following:
`var s = string.Format("Testing message {1}", "parameter formatting");`
 
and 

```
Task("Default")
    .Does(() => {
	Information("Testing message {1}", "parameter formatting");
    });

RunTarget("Default");
```

In both cases, the exception is `FormatException` with message _"Index (zero based) must be greater than or equal to zero and less than the size of the argument list."_

It is worth noting that while this is cleaner than _"Index was outside the bounds of the array."_, it will still (partially) render the content by nature of the implementation of the token parsing/rendering.

```
Default                                  
Executing task: Default                  
Testing message                          
```

If we want to short-circuit the partial-rendering, the only way I can see in the way of an "all or nothing" approach would be to effectively perform a pre-flight check of the formatting string, something like this in [CakeBuildLog](https://github.com/cake-build/cake/blob/3b9aeced6d4d73f1fe8dde4485a31bec85b01036/src/Cake.Core/Diagnostics/CakeBuildLog.cs#L47), which I don't think is necessary or desirable, but can amend the PR accordingly if that's something we want to move forward with.
```
public void Write(Verbosity verbosity, LogLevel level, string format, params object[] args)
{
    if (verbosity > Verbosity)
    {
        return;
    }

    // make sure the format string + args align
    try
    {
       var output = string.Format(format, args);
    }
    catch(FormatException) 
    { 
        throw; 
    }

    // continue normal implementation
```

### DIFF
--- a/src/Cake.Core.Tests/Unit/Diagnostics/Formatting/PropertyTokenTests.cs
+++ b/src/Cake.Core.Tests/Unit/Diagnostics/Formatting/PropertyTokenTests.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using Cake.Core.Diagnostics.Formatting;
+using Xunit;
+
+namespace Cake.Core.Tests.Unit.Diagnostics.Formatting
+{
+    public sealed class PropertyTokenTests
+    {
+        public sealed class TheRenderMethod
+        {
+            [Fact]
+            public void Should_Throw_FormatException_When_Index_And_Args_Are_Mismatched()
+            {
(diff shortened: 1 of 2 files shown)


devlead:

Issue #1997, opened 2018-01-10
Add the --trace option to the NUnit3Settings class.


WayneHiller:
NUnit3 - Can we add in the --trace option to the NUnit3Setting class and to the Runner? Now that --verbose is gone there is no way to turn on detailed logging.

kcamp:
I can make a pass over this & have a PR in the next day or so.

gep13:
@kcamp thank you very much!

devlead:
Fixed by #2001

PR #2001, opened 2018-01-12
GH-1997: Add support for --trace command line option for NUnit3 settings and runner

kcamp:
This addresses the improvement identified in GH-1997.

A couple of things to note -
- Please feel free to comment on the name of type `NUnitInternalTraceLevel`.  I think it could conceivably be shortened (`NUnitTraceLevel`, maybe?); I don't feel strongly that we necessarily _need_ to disambiguate the type as "Internal", as there's really only one notion of "trace".
- I took this implementation in the direction of an `enum` over the `string Trace { get; set; }` implementation in the v2 implementation of `NUnitSettings`.  I believe that the usage of enums here creates a stronger API that forces users into a successful end-state by constraining their choices.  However, I acknowledge that it creates a divide between the v2 and v3 Cake API that we may not wish to introduce.  That said, the changes between NUnit v2 and v3 were already significant enough that I don't know that we need to be too particularly concerned with this.

If there are changes to be made, please let me know.

### DIFF
--- a/src/Cake.Common/Tools/NUnit/NUnitInternalTraceLevel.cs
+++ b/src/Cake.Common/Tools/NUnit/NUnitInternalTraceLevel.cs
@@ -1,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Cake.Common.Tools.NUnit
+{
+    /// <summary>
+    /// Represents the level of detail at which NUnit should set internal tracing
+    /// </summary>
+    public enum NUnitInternalTraceLevel
+    {
+        /// <summary>
+        /// Do not display any trace messages
+        /// </summary>
+        Off = 0,
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


src/Cake.Common/Tools/NUnit/NUnitInternalTraceLevel.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Cake.Common.Tools.NUnit
+{
+    /// <summary>
+    /// Represents the level of detail at which NUnit should set internal tracing
+    /// </summary>
+    public enum NUnitInternalTraceLevel
```
patriksvensson: Can we assign values for these? `Off = 0` etc.

kcamp: Certainly. Change pushed.

devlead:

PR #2015, opened 2018-01-20
GH2014: Allow multiple NugetFeeds for InProcess-NugetClient

### LINKED ISSUE #2014: In-process NuGet don’t support multiple feeds through config
Simple fix, put a `string.Split(’;’,..)` here https://github.com/cake-build/cake/blob/develop/src/Cake.NuGet/Install/NuGetSourceRepositoryProvider.cs#L37

Insire:
- added basic argument validation in NuGetSourceRepositoryProvider ctor
- added support for multiple feeds as suggested here #2014 
- added basic unit tests testing argument validation
- added unit tests for NuGetSourceRepositoryProvider testing added functionality

should fix #2014

### DIFF
--- a/src/Cake.NuGet/Install/NuGetSourceRepositoryProvider.cs
+++ b/src/Cake.NuGet/Install/NuGetSourceRepositoryProvider.cs
@@ -19,6 +19,21 @@ namespace Cake.NuGet.Install
 
         public NuGetSourceRepositoryProvider(ISettings settings, ICakeConfiguration config, PackageReference package)
         {
+            if (settings == null)
+            {
+                throw new ArgumentNullException(nameof(settings));
+            }
+
+            if (config == null)
+            {
+                throw new ArgumentNullException(nameof(config));
+            }
+
+            if (package == null)
+            {
+                throw new ArgumentNullException(nameof(package));
+            }
+
             // Create the package source provider (needed primarily to get default sources)
             PackageSourceProvider = new PackageSourceProvider(settings);
 
@@ -34,10 +49,18 @@ namespace Cake.NuGet.Install
                 CreateRepository(package.Address.AbsoluteUri);
             }
 
-            var nugetSource = config.GetValue(Constants.NuGet.Source);
-            if (!string.IsNullOrWhiteSpace(nugetSource))
+            var nugetSources = config.GetValue(Constants.NuGet.Source);
+            if (string.IsNullOrEmpty(nugetSources))
             {
-                CreateRepository(nugetSource);
+                return;
+            }
+
+            foreach (var nugetSource in nugetSources.Split(';'))
+            {
+                if (!string.IsNullOrWhiteSpace(nugetSource))
+                {
+                    CreateRepository(nugetSource);
+                }
             }
 
             foreach (var source in PackageSourceProvider.LoadPackageSources())
--- a/src/Cake.NuGet.Tests/Unit/NuGetSourceRepositoryProviderTests.cs
+++ b/src/Cake.NuGet.Tests/Unit/NuGetSourceRepositoryProviderTests.cs
@@ -1,0 +1,97 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Cake.Core.Configuration;
+using Cake.Core.Packaging;
+using Cake.NuGet.Install;
+using NuGet.Configuration;
+using Xunit;
+
+namespace Cake.NuGet.Tests.Unit
+{
+    public sealed partial class NuGetSourceRepositoryProviderTests
+    {
+        public sealed class TheConstructor
+        {
+            [Fact]
+            public void Should_Throw_If_Settings_Is_Null()
+            {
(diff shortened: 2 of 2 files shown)


src/Cake.NuGet/Install/NuGetSourceRepositoryProvider.cs
```diff
@@ -34,10 +49,13 @@ public NuGetSourceRepositoryProvider(ISettings settings, ICakeConfiguration conf
                 CreateRepository(package.Address.AbsoluteUri);
             }
 
-            var nugetSource = config.GetValue(Constants.NuGet.Source);
-            if (!string.IsNullOrWhiteSpace(nugetSource))
+            var nugetRawSources = config.GetValue(Constants.NuGet.Source);
```
mholo65: I’d say `nugetSources` instead of `nugetRawSources`.

mholo65:
@devlead is bitrise Ubuntu failing due to your NET Core 2 tests?

devlead:
@mholo65 yes

Insire:
Is there anything i can do to help with that?

mholo65:
@Insire, at this point, no. We’ll just need to wait for the NET Core 2 PR to get merged.

What you can do in the meantime is to squash those two commits.

Insire:
This should have worked. I'm not too familiar with git yet. Let me know if there is anything to fix.

src/Cake.NuGet/Install/NuGetSourceRepositoryProvider.cs
```diff
@@ -34,10 +49,13 @@ public NuGetSourceRepositoryProvider(ISettings settings, ICakeConfiguration conf
                 CreateRepository(package.Address.AbsoluteUri);
             }
 
-            var nugetSource = config.GetValue(Constants.NuGet.Source);
-            if (!string.IsNullOrWhiteSpace(nugetSource))
+            var nugetSources = config.GetValue(Constants.NuGet.Source);
+            foreach (var nugetSource in nugetSources.Split(';'))
```
mholo65: Just curious, what will happen if `Constant.NuGet.Source` is not set in configuration? Will `ICakeConfiguration` return null?

mholo65: Seems like it will return `null`... We should guard for `null` here, otherwise we'll get a `NullReferenceException`
https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Configuration/CakeConfiguration.cs#L35

Insire: What if the nugetSource is actually null? Should we throw an exception for that, or is simply returning enough? My Hunch is, that returning should be ok, i am not sure though.

aateeque:
@mholo65 is there anyting left to do on this PR?

mholo65:
No, should be ok! I’ll try finding time to take a final look soon!

mholo65:
@Insire I've rebased this. Will merge as soon as CI passes.

mholo65:
@Insire your changes have been merged. Thank you for your contribution.

nikhilagrawaldotnet:
What are your thoughts on adding `-nuget_sources` parameter (along with existing `-nuget-source`). This use case is already in cake in the form of `version` or `ver`?

I'll pick it in case you are wondering. :)

gep13:
>@nikhilagrawaldotnet said...
his use case is already in cake in the form of version or ver?

Can you point me at where you are seeing this?

nikhilagrawaldotnet:
In `Cake.Arguments.ArgumentParser`

    if (name.Equals("version", StringComparison.OrdinalIgnoreCase) || name.Equals("ver", StringComparison.OrdinalIgnoreCase))

gep13:
@nikhilagrawaldotnet hmm, that is one specific use case for the arguments for Cake.exe itself.  You are talking about adding an alias for one of the aliases that is executed by Cake, which isn't really the same thing.  In this instance, I don't think there is a need to do this.

@cake-build/cake-team what are your thoughts?

mholo65:
@gep13 @nikhilagrawaldotnet 

Well there are differences. `--version` and `--ver` is an argument that `Cake.exe` itself knows how to handle, while [`--nuget_source`](https://github.com/cake-build/cake/blob/develop/src/Cake.NuGet/Constants.cs) is a configuration that only `Cake.NuGet.dll` knows how handle. Please not that configurations can also be set via environment variables or `cake.config`. I do feel that `nuget_source` as singular feels wrong, but it might be too late to change that, and we'll end up with two different configuration keys with same value for a long time. However, even in [`nuget.exe` ](https://docs.microsoft.com/en-us/nuget/tools/cli-ref-install) this is calles `source` even though it accepts a list of sources concatenated with `;`.

nikhilagrawaldotnet:
In that case this discussed stands closed. #MicDrop

gep13:
>@mholo65 said...
However, even in nuget.exe this is calls source even though it accepts a list of sources concatenated with `;`.

This was my exact thought too.

PR #2020, opened 2018-01-25
[Proof of concept] Expand environment variables during preprocess

Kemyke:
Expanding environment vaariables in `UriDirectiveProcessor`.

### DIFF
--- a/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
+++ b/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
@@ -69,6 +69,25 @@ namespace Cake.Core.Tests.Unit.Scripting.Processors
             Assert.Equal("/Working/other.cake", result.Script.Includes[1].Path.FullPath);
         }
 
+        [Theory]
+        [InlineData("#l \"%CAKE_TEST_SCRIPT_PATH%/utils.cake\"")]
+        public void Should_Process_Environment_Variable_Found_In_Source(string source)
+        {
+            // Given
+            System.Environment.SetEnvironmentVariable("CAKE_TEST_SCRIPT_PATH", "/Working");
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("/Working/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/Working/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
         [Fact]
         public void Should_Insert_Line_Directives_When_Processing_Load_Directives()
         {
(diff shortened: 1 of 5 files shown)


tests/integration/build.ps1
```diff
@@ -50,6 +50,7 @@ Function GetCakeCoreCLRPath([string]$ScriptRoot)
 # PREPARATION
 #####################################################################
 
+$Env:CAKE_INTEGRATION_TEST_ROOT = "../.."
```
devlead: this should be after 
https://github.com/Kemyke/cake/blob/d8dd72d4ce3bdd6cd68bcf9c4db0655ff508b5e6/tests/integration/build.ps1#L174

where integration test variables are set,

devlead: Also we would need to do the same in `build.sh`

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
```diff
@@ -69,6 +69,25 @@ public void Should_Process_Multiple_Script_References_Found_In_Source(string sou
             Assert.Equal("/Working/other.cake", result.Script.Includes[1].Path.FullPath);
         }
 
+        [Theory]
+        [InlineData("#l \"%CAKE_TEST_SCRIPT_PATH%/utils.cake\"")]
+        public void Should_Process_Environment_Variable_Found_In_Source(string source)
+        {
+            // Given
+            System.Environment.SetEnvironmentVariable("CAKE_TEST_SCRIPT_PATH", "/Working");
```
devlead: This fails on posix systems. Haven't checked why, this should only affect process, but could be something with how mono handles processes.

patriksvensson: To be honest I'm not a fan of setting environment variables like this in a test. It works but is more or less the same thing as a shared mutable state.

devlead: Agree doesn't "feel" ok. Don't know what would be better, picking existing known to exist? That doesn't feel 100% either.

Kemyke:
I was thinking about to do this in `ScriptAnalyzer` but it looked quite risky to me. Maybe it is not a good idea to expand the env variables in all lines because we don't know the context. I mean it would replace variables in string, comments, etc. Furthermore this environment variable expansion is only important in the preprocessor directives. In the cake script you can easily access env vars with `EnvironmentVariable` function. But in the preprocess step it is not possible to use these functions or variables of course so my only idea was the environment variable to access various values during the preprocess step. 
That's why i implemented in `UriDirectiveProcessor` so the expansion happens only in URI-s. 

Or is it possible somehow to use variables in this directives? Like `$"#load {repoRoute}/utils.cake"` or `#addin "nuget:?package=Cake.Some.Nuget&version={someNugetVersion}"`. Would this be a better way?

patriksvensson:
@Kemyke I accidentally merged this one to develop and I will have to revert it. Sorry for the inconvenience.

patriksvensson:
@Kemyke As soon as the PR is reverted you should be able to rebase it against develop and submit it again.

Again, sorry about the inconvenience. First time this happened to me 😞

devlead:
@Kemyke I've reverted so should just need to open a new pr.

Kemyke:
No problem guys. I made a new [PR](https://github.com/cake-build/cake/pull/2023) and copied my thoughts and questions.

Kemyke:
> Cause it doesn't change the source line, only the line passed to to the directives, if no directive accepts the line it'll just add the source line

Ok, now i understand and i agree it looks much better there.

Issue #85, opened 2014-08-17
Support environment variables in paths


patriksvensson:
Paths should support environment variables (on Mono and Windows).

Segments surrounded by `%` tokens are not technically valid paths (on Windows), but we're going to make an exception for this. This means that we should no longer throw an exception when creating an illegal path, but mark the path as invalid. The `$` token is valid on *nix systems, so a path like that will still be considered valid.

A path containing environment variables will be considered invalid. Tools and script aliases that work with paths will have to take this into consideration and act accordingly.

**Path**
- `bool Path.ContainsEnvironmentVariables()`
- `bool IsValidPath { get; }`

**FilePath**
- `FilePath ExpandEnvironmentVariables(ICakeEnvironment environment)`

**DirectoryPath**
- `DirectoryPath ExpandEnvironmentVariables(ICakeEnvironment environment)`

fleed:
Any update on supporting string interpolation (`$` sign in front of strings)?

devlead:
@fleed It's already supported if you pass the `-Experimental` flag to Cake or use the CoreCLR version of Cake.

devlead:
@fleed CoreCLR was recently released along the v0.16.0 version, there's been some fixes required and we're today releasing v0.16.2 

Biggest limitation is that addins need to be .NET Core compatible assemblies, tools work usual as long as their dependencies exist on the machine they're executing, but if you're i.e. running in a CoreCLR only environment like Nano Server tools requiring desktop .NET framework won't work.

In the v0.17.0 time frame we'll focus on delivering documentation, examples and guidance around CoreCLR, An really basic example using Cake CoreCLR can be found [here](https://github.com/devlead/BitbucketPipelinesShield)

patriksvensson:
Superseded by #2312

PR #2023, opened 2018-01-26
GH2078: Expand environment variables in Uri

Kemyke:
This is the second PR after the first was accidentally merged. Here are my thoughts from the previous conversation:

I was thinking about to do this in `ScriptAnalyzer` but it looked quite risky to me. Maybe it is not a good idea to expand the env variables in all lines because we don't know the context. I mean it would replace variables in string, comments, etc. Furthermore this environment variable expansion is only important in the preprocessor directives. In the cake script you can easily access env vars with `EnvironmentVariable` function. But in the preprocess step it is not possible to use these functions or variables of course so my only idea was the environment variable to access various values during the preprocess step. 
That's why i implemented in `UriDirectiveProcessor` so the expansion happens only in URI-s. 

Or is it possible somehow to use variables in this directives? Like `$"#load {repoRoute}/utils.cake"` or `#addin $"nuget:?package=Cake.Some.Nuget&version={someNugetVersion}"`. Would this be a better way?

### DIFF
--- a/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
+++ b/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
@@ -69,6 +69,45 @@ namespace Cake.Core.Tests.Unit.Scripting.Processors
             Assert.Equal("/Working/other.cake", result.Script.Includes[1].Path.FullPath);
         }
 
+        [Theory]
+        [InlineData("#l \"%CAKE_TEST_SCRIPT_PATH%/utils.cake\"")]
+        public void Should_Process_Environment_Variable_Found_In_Source(string source)
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Environment.SetEnvironmentVariable("CAKE_TEST_SCRIPT_PATH", "test");
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("/Working/test/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/Working/test/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
+        [Theory]
+        [InlineData("#l \"%CAKE_TEST_SCRIPT_BASE_PATH%/%CAKE_TEST_SCRIPT_PATH%/utils.cake\"")]
+        public void Should_Process_Multiple_Environment_Variable_Found_In_Source(string source)
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Environment.SetEnvironmentVariable("CAKE_TEST_SCRIPT_BASE_PATH", "test");
+            fixture.Environment.SetEnvironmentVariable("CAKE_TEST_SCRIPT_PATH", "scripts");
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("/Working/test/scripts/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/Working/test/scripts/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
         [Fact]
         public void Should_Insert_Line_Directives_When_Processing_Load_Directives()
         {
--- a/src/Cake.Core/ICakeEnvironmentExtensions.cs
+++ b/src/Cake.Core/ICakeEnvironmentExtensions.cs
@@ -1,0 +1,41 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Cake.Core
+{
+    /// <summary>
+    /// Extension methods for ICakeEnvironment
+    /// </summary>
+    public static class ICakeEnvironmentExtensions
+    {
+        /// <summary>
+        /// Expands the environment variables in the text. Variables shoud be quoted with %.
+        /// </summary>
+        /// <param name="environment">The cake environment interface</param>
+        /// <param name="text">A string containing the names of zero or more environment variables.</param>
+        /// <returns>A string with each environment variable replaced by its value.</returns>
+        public static string ExpandEnvironmentVariables(this ICakeEnvironment environment, string text)
+        {
+            var environmentVariables = environment.GetEnvironmentVariables();
+            Regex r = new Regex("%(.*?)%");
+            var matches = r.Matches(text);
+            foreach (Match m in matches)
+            {
+                string varName = m.Groups[1].Value;
+                if (environmentVariables.ContainsKey(varName))
+                {
+                    text = text.Replace(m.Value, environmentVariables[varName]);
+                }
+            }
+            return text;
+        }
+    }
+}
--- a/src/Cake.Core/Scripting/Analysis/ScriptAnalyzer.cs
+++ b/src/Cake.Core/Scripting/Analysis/ScriptAnalyzer.cs
@@ -117,7 +117,7 @@ namespace Cake.Core.Scripting.Analysis
             {
                 string replacement = null;
 
-                if (!_lineProcessors.Any(p => p.Process(context, line, out replacement)))
+                if (!_lineProcessors.Any(p => p.Process(context, _environment.ExpandEnvironmentVariables(line), out replacement)))
                 {
                     context.AddScriptLine(line);
                 }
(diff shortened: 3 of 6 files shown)


devlead:
> Maybe it is not a good idea to expand the env variables in all lines because we don't know the context.

The below change shouldn't expand on all lines

https://github.com/cake-build/cake/blob/6011c9b67e6d686ef85fa9bc3c7c79e519ed0ab4/src/Cake.Core/Scripting/Analysis/ScriptAnalyzer.cs#L120

```diff
-if (!_lineProcessors.Any(p => p.Process(context, line, out replacement)))
+if (!_lineProcessors.Any(p => p.Process(context, Environment.ExpandEnvironmentVariables(line), out replacement)))
```

Cause it doesn't change the source line, only the line passed to to the directives, if no directive accepts the line it'll just add the source line
https://github.com/cake-build/cake/blob/6011c9b67e6d686ef85fa9bc3c7c79e519ed0ab4/src/Cake.Core/Scripting/Analysis/ScriptAnalyzer.cs#L122
but if the the line is processed any directive then a replacement line is added instead
https://github.com/cake-build/cake/blob/6011c9b67e6d686ef85fa9bc3c7c79e519ed0ab4/src/Cake.Core/Scripting/Analysis/ScriptAnalyzer.cs#L127

We would still need integration tests for existing directives, but it should only be directives that are affected, also expand environment variables will only expand if the environment variable exists.

The reason I'm contemplating this is that it would be cleaner(no change of base classes or method signatures), and make sure it works consistent across directives.

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
```diff
@@ -60,14 +60,17 @@ public sealed override bool Process(IScriptAnalyzerContext context, string line,
         private Uri ParseUriFromTokens(string[] tokens)
         {
             Uri uri;
-            if (!Uri.TryCreate(tokens[1].UnQuote(), UriKind.Absolute, out uri))
+
+            var expandedTokens = tokens.Select(t => Environment.ExpandEnvironmentVariables(t));
```
patriksvensson: Perhaps we could add an `Expand` method to `ICakeEnvironment` instead of using System.Environment directly? That way it should be pretty straight forward mocking it during tests.

Kemyke: There is the `GetEnvironmentVariables` function in the `ICakeEnvironment`. Wouldn't it be better to use that method and do the variable substitution manually?

Kemyke:
> Cause it doesn't change the source line, only the line passed to to the directives, if no directive accepts the line it'll just add the source line

Ok, now i understand and it looks better to me also.

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
patriksvensson: @Kemyke It would be OK to call `Environment.ExpandEnvironmentVariables(t)` from within the `CakeEnvironment` class. Putting it there would give us a way of mocking the behavior in tests.

Kemyke: @patriksvensson  Yeah, i understand, i just look through the `ICakeEnvironment` interface and it looks a little bit redundant to me to have and `ExpandEnvironmentVariable` method and the `GetEnvironmentVariables` dictionary. You could do the mocking with that dictionary as well. 
It was just an idea to keep the `ICakeEnvironment` interface slim. I will do these changes and modify the PR.

Kemyke:
@devlead @patriksvensson i moved the logic to ScriptAnalyzer and changed the unit test. 
If it is looking better for you i will do all the other tests for all directives.

gep13:
@Kemyke was there an issue created to go along with this PR? Thanks

Kemyke:
@gep13 there is an improvement tagged issue which is slightly related to this PR i think https://github.com/cake-build/cake/issues/85

gep13:
@Kemyke agreed.  I don't think it would close that issue, but I do agree that it is related. :+1:

Kemyke:
@gep13 please let me know if there is anything i can do to move forward with this PR!

devlead:
Relates to #2078

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
```diff
@@ -69,6 +69,25 @@ public void Should_Process_Multiple_Script_References_Found_In_Source(string sou
             Assert.Equal("/Working/other.cake", result.Script.Includes[1].Path.FullPath);
         }
 
+        [Theory]
+        [InlineData("#l \"%CAKE_TEST_SCRIPT_PATH%/utils.cake\"")]
+        public void Should_Process_Environment_Variable_Found_In_Source(string source)
```
devlead: This tests fails on MacOS / Linux

```
[xUnit.net 00:00:02.7407740]     Cake.Core.Tests.Unit.Scripting.Processors.LoadDirectiveProcessorTests.Should_Process_Environment_Variable_Found_In_Source(source: "#l \"%CAKE_TEST_SCRIPT_PATH%/utils.cake\"") [FAIL]
[xUnit.net 00:00:02.7444020]       Assert.Equal() Failure
[xUnit.net 00:00:02.7445470]       Expected: 1
[xUnit.net 00:00:02.7446970]       Actual:   0
[xUnit.net 00:00:02.7463720]       Stack Trace:
[xUnit.net 00:00:02.7488290]         /Users/vagrant/git/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs(84,0): at Cake.Core.Tests.Unit.Scripting.Processors.LoadDirectiveProcessorTests.Should_Process_Environment_Variable_Found_In_Source(String source)
Failed   Cake.Core.Tests.Unit.Scripting.Processors.LoadDirectiveProcessorTests.Should_Process_Environment_Variable_Found_In_Source(source: "#l \"%CAKE_TEST_SCRIPT_PATH%/utils.cake\"")
Error Message:
 Assert.Equal() Failure
Expected: 1
Actual:   0
Stack Trace:
   at Cake.Core.Tests.Unit.Scripting.Processors.LoadDirectiveProcessorTests.Should_Process_Environment_Variable_Found_In_Source(String source) in /Users/vagrant/git/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs:line 84
```

Kemyke: I fixed the test. Now there is a CodeFactor issue which is not really connected to my PR, i think.

src/Cake.Core/ICakeEnvironment.cs
```diff
@@ -40,6 +40,13 @@ public interface ICakeEnvironment
         /// <returns>The value of the environment variable.</returns>
         string GetEnvironmentVariable(string variable);
 
+        /// <summary>
+        /// Expands the environment variables in the text. Variables shoud be quoted with %.
+        /// </summary>
+        /// <param name="text">A string containing the names of zero or more environment variables.</param>
+        /// <returns>A string with each environment variable replaced by its value.</returns>
+        string ExpandEnvironmentVariables(string text);
```
devlead: Been trying this out, and while I really like the feature, we're really hesitant to introduce a breaking change in Cake.Core 0.27.0 - which adding an interface member is.

Been giving some thought on how not to introduce a breaking change, and only possibility would probably be to switch from an interface member to an extension method, something like

`public static string ExpandEnvironmentVariables(this ICakeEnvironment environment, string text)`

Would make it harder to differentiate between Fake and non fake environments, one naive way would be `if (environment is CakeEnvironment)` or just use the same implementation for both (_that way unit test would run same code as scripts_).

`_environmentVariables` could be replaced with `ICakeEnvironment.GetEnvironmentVariables()` in that method.

Kemyke: Can you give me some insight why is it a breaking change to add a method to an interface? Meanwhile i will check how to do the same with extension method.

devlead: An interface is an contract, any addin / module / assembly inheriting that would break as they would no longer uphold the contract.

Kemyke: I didn't know that the addins can implement the ICakeEnvironment interface. When is this a valid use case?
I thought ICakeEnvironment is something what is used by the addins and a new feature won't be a breaking change. And you have to update the Cake reference in the addin only if you would like to use that new feature.

Kemyke: For the extension method:
what if i don't use the `Environment.ExpandEnvironmentVariables`method but we use the `GetEnvironmentVariables()` of the `ICakeEnvironment` so it will work for normal and fake environment as well. The implementation can be almost the same what is in the `FakeEnvironment.xpandEnvironmentVariables` now.
What do you think? I think this is a better approach than casting the interface to exact types.

devlead: Issue is that it's a public interface, as such we can't know who inherits from it. Someone might replace it with a module or i.e. proxy it within in a addin or their unit tests etc. 

We try really hard to keep both source code and binary compatibility, and that's why we're very careful about changes in Cake.Core, sometimes the feature justifies a breaking change, but in this case I think it's worth the extra effort to not break existing things.

There's some good guidance on the topic provided by the .NET team at:
https://github.com/dotnet/corefx/tree/master/Documentation/coding-guidelines

devlead: > For the extension method:
what if i don't use the Environment.ExpandEnvironmentVariablesmethod but we use the GetEnvironmentVariables() of the ICakeEnvironment so it will work for normal and fake environment as well. The implementation can be almost the same what is in the FakeEnvironment.xpandEnvironmentVariables now.
What do you think? I think this is a better approach than casting the interface to exact types.

Sounds reasonable to me.

Kemyke:
@devlead i create the extension method. I create a new `ICakeEnvironmentExtensions.cs` file and put the logic in it. Is it a good place for it?
Please check how you like this implementation!

devlead:
@Kemyke your changes have been merged, thanks for your contribution 👍

Kemyke:
Thanks!

Issue #2039, opened 2018-02-14
XUnit2Runner doesn't respect `ParallelismOption.None` 


kcamp:
### What You Are Seeing?
XUnit2 is not respecting `ParallelismOption.None`

### What is Expected?
The XUnit tool command should include the `-parallel none` when it's specified.

### What version of Cake are you using?
0.25

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
N/A

### How Did You Get This To Happen? (Steps to Reproduce)

In consideration of this sample set of tests..
```
public sealed class SomeClass
{
    [Fact]
    public void OneShouldBeOne()
    {
        Thread.Sleep(5000);
        "one".Should().Be("one");
    }

    [Fact]
    public void TwoShouldBeTwo()
    {
        Thread.Sleep(5000);
        2.Should().Be(2);
    }
}

public sealed class AnotherClass
{
    [Fact]
    public void ThreeShouldBeThree()
    {
        Thread.Sleep(5000);
        3.Should().Be(3);
    }
}
```

When running this assembly with the task (following) I would expect the elapsed time of the tests to be 15 seconds.

```
XUnit2(
    assm,
    new XUnit2Settings
    {
        NoAppDomain = false,
        ShadowCopy = false,
        Parallelism = ParallelismOption.None
    });
```

Instead, it executes in ~10 seconds, indicating some degree of parallelism is at play.

Reading the docs at https://xunit.github.io/docs/running-tests-in-parallel.html
```
-parallel option 
Allows the user to specify which kinds of parallelization should be allowed for the test run. The valid option values are:
**none** Turns off all parallelization
**collections** Parallelizes collections, not assemblies
**assemblies** Parallelizes assemblies, not collections
**all** Parallelizes both collections and assemblies

The default value is collections. 
```

I think this last bit (defaulting to `collections`)is where our bug is coming into play; by suppressing the `-parallel none`, we're accepting the runner default.

Running that same set of tests again, manually from the CLI (below) to explicitly add the `none` flag gives execution time of ~15 seconds, as expected.

```
[...]

PR #2040, opened 2018-02-14
GH2039: Fix XUnit2Settings + XUnit2Runner to honor ParallelismOption.None

kcamp:
Here is the first pass over what I believe to be the most appropriate fix for GH-2039.
I've made the property on the `XUnit2Settings` class nullable.
By default, assuming a user hasn't set the value, it will not be emitted as an argument.
This maintains parity with the previous implementation where `ParallelismOptions.None`, the CLR default, would suppress the emission.  In either case, no explicit argument will allow the runner to use the default.

The _other_ option that I entertained that we could explore further would be setting the default on the `XUnit2Settings` class to `Collections` to match the runner, but I think this has the potential to create regressions for people.

Please feel free to suggest other implementations if they're more desirable.


devlead:

PR #2041, opened 2018-02-18
GH1213: Improve NuGetPushSettings.Source documentation

### LINKED ISSUE #1213: NuGetPushSettings.Source: incorrect documentation
Regarding NuGetPushSettings.Source: documentation says `If not specified, nuget.org is used`. However, when this value is null I am getting the following error: `Source parameter was not specified`.

The documentation in question is here: http://cakebuild.net/api/cake.common.tools.nuget.push/bcd6313b
and also here: http://cakebuild.net/api/cake.common.tools.nuget.push/bcd6313b/83747a07

By the way, since there is no 'default' value, we should document the most common values (e.g.: nuget.org, myget, etc.).

Jericho:
Document the fact that NuGetPushSettings.Source is mandatory since nuget 3.4.2 and also document the url for two of the most popular public nuget servers (nuget.org and myget).

Resolves #1213

### DIFF
--- a/src/Cake.Common/Tools/NuGet/Push/NuGetPushSettings.cs
+++ b/src/Cake.Common/Tools/NuGet/Push/NuGetPushSettings.cs
@@ -14,12 +14,21 @@ namespace Cake.Common.Tools.NuGet.Push
     public sealed class NuGetPushSettings : ToolSettings
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


Jericho:
Done. Encoded the `<` and `>` in the XML comments

devlead:
@Jericho  your changes have been merged, thanks for your contribution 👍

PR #2069, opened 2018-03-06
GH2070: Specify multiple nuget source

aateeque:
Specifying multiple nuget sources in `nuget_source` cmd line arg currently fails. Believe we introduced this regression when we moved to internal nuget as previously this would be passed to nuget.exe.

note: nuget cli delimits source parameter with ";" so I've stuck to that


mholo65:
I believe this is a duplicate of https://github.com/cake-build/cake/pull/2015

PR #2072, opened 2018-03-08
Support for setting environment variable on Bitrise

stefandevo:
I think it make sense that setting an environment variable is part of the common Bitrise support in Cake, instead of making EnvMan an add-in. 
Checked the AppVeyor provider and saw extra methods on the same level, so I think it makes sense.


stefandevo:
It seems that the team did not want this.

gep13:
>@stefandevo said...
It seems that the team did not want this.

I would suggest that you leave this open for review a bit longer.

Also, was there an issue opened for this, as per our guidelines?

stefandevo:
No issue was created. I will open one.

devlead:
@stefandevo your changes have been merged, thanks for your contribution 👍

PR #2080, opened 2018-03-13
GH2079: Set default nuget sources

### LINKED ISSUE #2079: Default sources not loaded if nuget_source is empty
#2014 introduced a bug where default sources would not be loaded if `nuget_source` config was null or empty.

Providing PR...

mholo65:
- Fixes #2079


devlead:
@mholo65 your changes have been merged, thanks for your contribution 👍

PR #2089, opened 2018-03-16
GH2088: Don't quote VSWhere -requires and -products arguments

### LINKED ISSUE #2088: VSWhere -requires and -products argument values are quoted but VSWhere doesn't support multiple values in quotes
### What You Are Seeing?
Cake's VSWhere aliases put quotes around the `-requires` and `-products` argument values, which are supposed to allow multiple workloads/products to be specified, but per Microsoft/vswhere#141 VSWhere doesn't support passing multiple workloads/products within one quoted argument value.

### What is Expected?
Cake's VSWhere aliases should not put quotes around the `-requires` and `-products` argument values in order to allow multiple workloads/products to be specified.

0.26.1

Windows

This code:
```C#
DirectoryPath vs = VSWhereLatest(new VSWhereLatestSettings
{
    Requires = "Microsoft.Component.MSBuild Microsoft.VisualStudio.PackageGroup.TestTools.Core"
});
```
currently ends up running something like
```
vswhere -latest -requires "Microsoft.Component.MSBuild Microsoft.VisualStudio.PackageGroup.TestTools.Core"
```
which will never return anything due to VSWhere treating the entire `-requires` argument value as a single workload ID.

Similarly, code like `VSWhereProducts("X Y")` currently ends up running something like `vswhere -products "X Y"` which will never return anything because VSWhere looks for the product ID "X Y" rather than "X" and "Y" separately.

sean-rose:
Changes VSWhere `-requires` and `-products` arguments to not be quoted, since VSWhere doesn't support passing multiple workloads/products within one quoted argument value.

Fixes #2088.


gep13:
@sean-rose thank you for making this contribution.  This has been rebased and merged into the develop branch here: https://github.com/cake-build/cake/commit/8afc366cba2f988f45603362d96305e55d4821de

Issue #2094, opened 2018-03-21
Add support for executing a single task without dependencies


patriksvensson:
Sometimes you just want to execute a single task, without executing its dependencies. It would be nice to be able to tell Cake to just execute the task as it is.

```
> Cake.exe --target Foo --exclusive
```

gep13:
@patriksvensson this sounds like a good idea to me.  I "think" @jnm2 might have created a related issue.

patriksvensson:
@gep13 I looked for an issue, but couldn't find one when I wrote it.

I assume you're thinking about https://github.com/cake-build/cake/issues/1341, but that's not the same thing.

gep13:
@patriksvensson yip, that is the one that I was thinking of.  Certainly related to this one.  Could both be tackled at the same time?

patriksvensson:
@gep13 I don't know. I see them as two different features.

jnm2:
You know, I think Patrik's idea would give me most of the benefit I was hoping for.
Especially if this is coupled with the ability to specify multiple tasks to run, which I hope we get to use with or without the --exclusive switch.

PR #2098, opened 2018-03-23
GH-2094: Allow execution of single task (without dependencies)

patriksvensson:
Related to #2094

### DIFF
--- a/src/Cake.Core/ICakeEngine.cs
+++ b/src/Cake.Core/ICakeEngine.cs
@@ -81,9 +81,9 @@ namespace Cake.Core
         /// </summary>
         /// <param name="context">The context.</param>
         /// <param name="strategy">The execution strategy.</param>
-        /// <param name="target">The target to run.</param>
+        /// <param name="settings">The execution settings.</param>
         /// <returns>The resulting report.</returns>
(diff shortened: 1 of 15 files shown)


src/Cake.Core/ICakeEngine.cs
```diff
@@ -55,9 +55,9 @@ public interface ICakeEngine
         /// </summary>
         /// <param name="context">The context.</param>
         /// <param name="strategy">The execution strategy.</param>
-        /// <param name="target">The target to run.</param>
+        /// <param name="settings">The execution settings.</param>
         /// <returns>The resulting report.</returns>
-        Task<CakeReport> RunTargetAsync(ICakeContext context, IExecutionStrategy strategy, string target);
+        Task<CakeReport> RunTargetAsync(ICakeContext context, IExecutionStrategy strategy, ExecutionSettings settings);
```
mholo65: Could we keep the old method and obsolete that? Don’t think much will break due to this, but still.

patriksvensson: No, sadly not possible since it then would rely on side-effects.

patriksvensson:
PR have been updated and rebased.

devlead:

HebaruSan:
Is it documented somewhere how to use this?

PR #2100, opened 2018-03-25
GH2095: add   skipnontestassemblies funcionality to nunit3 runner

pavzaj:
### What You Are Seeing?
NUnit3Settings class does not have property SkipNonTestAssemblies
### What is Expected?
NUnit3Settings class does have property SkipNonTestAssemblies and features behind it as described here: https://github.com/nunit/docs/wiki/Console-Command-Line
### What version of Cake are you using?
v0.26.1
### Are you running on a 32 or 64 bit system?
64
### What environment are you running on?  Windows? Linux? Mac?
Windows
### Are you running on a CI Server?  If so, which one?
NO

PR #2107, opened 2018-03-30
GH2106: Obsolete removed NUnit3 --err option

### LINKED ISSUE #2106: Remove NUnit3Settings.ErrorOutputFile property
See https://github.com/nunit/nunit-console/pull/381

The `--err` flag was recently removed from nunit3-console. It turned out it had never been implemented, and nobody noticed for the first four years. 😬 Given implementation complications and lack of demand, it was removed rather than implemented.

Cake should be updated accordingly. I'll submit a PR shortly. 😄

ChrisMaddock:
Fixes #2106.

When I got to the code, I saw that the `Full` and `Verbose` options, which were in the same position, have both just been Obsoleted rather than fully removed. For consistency, I did the same for `ErrorOutputFile`.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
@@ -217,8 +217,8 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
                 fixture.Settings.StopOnError = true;
                 fixture.Settings.Work = "out";
                 fixture.Settings.OutputFile = "stdout.txt";
-                fixture.Settings.ErrorOutputFile = "stderr.txt";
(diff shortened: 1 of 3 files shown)


pavzaj:
Hello, I've noticed that I have the same problem with CodeFactor as here:
https://github.com/cake-build/cake/pull/2100

Is there any way to get rid of this warning?
How does this validation actually work? Is this some pre-CodeFactor code issue revealed now because CodeFactor has been triggered by source code file change?

ChrisMaddock:
I assumed the issues would just be bypassed on merge. Like you say - the lines causing issues aren't touched by this PR. 🙂

devlead:
@ChrisMaddock your changes have been merged, thanks for your contribution 👍

ChrisMaddock:
Thanks!

PR #2110, opened 2018-04-02
GH1106: Multiple values passed to WithProperty are overridden

### LINKED ISSUE #1106: Multiple values passed to WithProperty are overridden
### What You Are Seeing?

MSBuildSettingsExtensions​.WithProperty(MSBuildSettings, ​string, ​String[]) does not work as intended for multiple values of some properties (Maybe all, I don't know). Each value seems to overwrite, so only the last supplied parameter will be applied.

Calling .WithProperty("WarningsNotAsErrors", "1573", "1591") results in /warnaserror-:1591

It looks like method which is used to parse the property dictionary inside MSBuildRunner: 
private static IEnumerable<string> GetPropertyArguments(IDictionary<string, IList<string>> properties)
{
	foreach (var propertyKey in properties.Keys)
	{
		foreach (var propertyValue in properties[propertyKey])
		{
			yield return string.Concat("/p:", propertyKey, "=", propertyValue);
		}
	}           
}

Will return "/p:WarningsNotAsErrors=1573", "/p:WarningsNotAsErrors=1591". Which may override when applied sequentially. 

Possibly the method could be amended to 

private static IEnumerable<string> GetPropertyArguments(IDictionary<string, IList<string>> properties)
{
	foreach (var propertyKey in properties.Keys)
	{
		var values = string.Join(",", properties[propertyKey]);
		yield return string.Concat("/p:", propertyKey, "=", values);
	}           
}

Which would return "/p:WarningsNotAsErrors=1573,1591". I'm not sure how this would affect other properties that are used, if a switch must be declared multiple times with single values it would break that.

### What is Expected?

Calling .WithProperty("WarningsNotAsErrors", "1573", "1591") would then become /warnaserror-:1573,1591 when run.

0.14.0.0

64

Windows 7

No, running from powershell script locally

Add .WithProperty("WarningsNotAsErrors", "1573", "1591") to any MSBuildSettings.

Run in Diagnostic.

Check the value sent to CoreCompile vs the Executing line.

### Output Log

Run with -Mono and -Experimental:

[log.txt](https://github.com/cake-build/cake/files/380631/log.txt)

mschlaffer:
Changed MSBuildExtension's WithProperty to forward multiple parameters as a comma separated list surrounded by quotes. Single parameter behavior is unchanged to before and does not have the added quotes.

This should fix https://github.com/cake-build/cake/issues/1106. Check the comments there for more details.


pascalberger (changes requested):
@Caopi Thanks you for your contribution and sorry for the late answer. We're now at a point where we can change this behavior in an upcoming release.

We would prefer to use semicolon instead of comas for separating the values as mentioned in the comment. Additionally there should be some [integration tests](https://github.com/cake-build/cake/tree/develop/tests/integration) for this, which makes sure that this actually works together with MsBuild.

Are you still in a position to continue with this PR or prefer for someone else to take over?

Caopi:
Hey Pascal, sorry for the late reply as well. I've quickly accepted your suggestion of course, but I would prefer if someone else could take over the integration test part. Haven't worked with C# and Cake in a while nowadays and don't have a lot of time to commit to this right now, sorry!

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -298,9 +298,14 @@ private static IEnumerable<string> GetPropertyArguments(IDictionary<string, ILis
...
-                foreach (var propertyValue in properties[propertyKey])
+                if (properties[propertyKey].Count > 1)
                 {
-                    yield return string.Concat("/p:", propertyKey, "=", propertyValue.EscapeMSBuildPropertySpecialCharacters());
+                    var commaSeparatedValues = string.Join(';', properties[propertyKey].Select(x => x.EscapeMSBuildPropertySpecialCharacters()));
+                    yield return string.Concat("/p:", propertyKey, "=", '"', commaSeparatedValues, '"');
+                }
+                else if (properties[propertyKey].Count == 1)
+                {
+                    yield return string.Concat("/p:", propertyKey, "=", properties[propertyKey].First().EscapeMSBuildPropertySpecialCharacters());
```
paulomorgado, line 308: Consider using string interpolation for readbility instead:


```suggestion
                    yield return $"/p:{propertyKey}={properties[propertyKey].First().EscapeMSBuildPropertySpecialCharacters())}";
```

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -298,9 +298,14 @@ private static IEnumerable<string> GetPropertyArguments(IDictionary<string, ILis
         {
             foreach (var propertyKey in properties.Keys)
             {
-                foreach (var propertyValue in properties[propertyKey])
+                if (properties[propertyKey].Count > 1)
                 {
-                    yield return string.Concat("/p:", propertyKey, "=", propertyValue.EscapeMSBuildPropertySpecialCharacters());
+                    var commaSeparatedValues = string.Join(';', properties[propertyKey].Select(x => x.EscapeMSBuildPropertySpecialCharacters()));
```
paulomorgado, line 303: Consider using string interpolation for readbility instead:


```suggestion
                    var commaSeparatedValues = $";{properties[propertyKey].Select(x => x.EscapeMSBuildPropertySpecialCharacters()))}";
```

src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs
```diff
@@ -298,9 +298,14 @@ private static IEnumerable<string> GetPropertyArguments(IDictionary<string, ILis
...
+                    yield return string.Concat("/p:", propertyKey, "=", '"', commaSeparatedValues, '"');
```
paulomorgado, line 304: Consider using string interpolation for readbility instead:


```suggestion
                    yield return $"/p:{propertyKey}=\"{commaSeparatedValues}\"";
```

paulomorgado (commented):
Can tests be added that have MSBuild escaped characters?

PR #2120, opened 2018-04-07
GH2119: Added missing .NET Core publish parameters

### LINKED ISSUE #2119: DotNetCore Publish misses Force / Self contained / Sources settings

devlead:
Missing DotNetCorePublishSettings added

* Force
* SelfContained
* Sources
* Fixes #2119

### DIFF
--- a/src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublishSettings.cs
+++ b/src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublishSettings.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
 using Cake.Common.Tools.DotNetCore.MSBuild;
 using Cake.Core.IO;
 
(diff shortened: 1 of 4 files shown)


src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublishSettings.cs
```diff
@@ -54,6 +55,30 @@ public sealed class DotNetCorePublishSettings : DotNetCoreSettings
...
+        /// </remarks>
+        public bool SelfContained { get; set; }
+
+        /// <summary>
+        /// Gets or sets the specified NuGet package sources to use during the restore.
+        /// </summary>
+        /// <remarks>
+        /// Requires .NET Core 2.x or newer.
+        /// </remarks>
+        public ICollection<string> Sources { get; set; }
```
patriksvensson: I think it would be a good idea to initialize this in the constructor to avoid null reference exceptions if the user assumes it's not null.

devlead: That would make it inconsistent with i.e.
https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/DotNetCore/Restore/DotNetCoreRestoreSettings.cs#L19

https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/NuGet/Restore/NugetRestoreSettings.cs#L32

though.

patriksvensson: But not inconsistent with https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs#L149

tests/integration/Cake.Core/Scripting/AddinDirective.cake
```diff
@@ -11,7 +11,13 @@ Task("Cake.Core.Scripting.AddinDirective.LoadNetStandardAddin")
...
     ";
 
-    CakeExecuteExpression(script);
+    CakeExecuteExpression(script,
+    	new CakeSettings {
+    		EnvironmentVariables = new Dictionary<string, string>{
+    			{"CAKE_NUGET_SOURCE", $"{Paths.Temp}/tools/Addins"},
+    			{"CAKE_PATHS_TOOLS", $"{Paths.Temp}/tools"},
+    			{"CAKE_PATHS_MODULES", $"{Paths.Temp}/tools/Modules"}
+    		}});
```
mholo65: Why was this changed?

tests/integration/Cake.Core/Scripting/AddinDirective.cake
```diff
@@ -11,7 +11,13 @@ Task("Cake.Core.Scripting.AddinDirective.LoadNetStandardAddin")
         Information(""The answer to life: {0}"", TheAnswerToLife);
     ";
 
-    CakeExecuteExpression(script);
+    CakeExecuteExpression(script,
+    	new CakeSettings {
+    		EnvironmentVariables = new Dictionary<string, string>{
+    			{"CAKE_PATHS_ADDINS", $"{Paths.Temp}/tools/Addins"},
```
devlead: @mholo65 this was changed because saw an issue on multiple Windows systems with addin installation as Cake changes current directory to script dir, which for expression is Windows temp director, setting it to be in a known directory within integration tests sorted it.

Long term CakeExecuteExpression probably should set these by default if not already set to the paths in current cake process config as that's probably the expected behaviour. 

Unreleated to issue so could remove, just couldn't run integration tests locally without it atm.

mholo65:
@devlead, your changes have been merged. Thank you for your contribution.

PR #2124, opened 2018-04-11
GHXXXX: Add Support for IDictionary tokens

maldworth:
I would like to add the ability for adding a dictionary of tokens.

I have a JSON file of key values, and when I use newtonsoft to deserialize, it defaults into an IDictionary<string,object>, and so I'd like to simply pass that into the TextTransform Extension Method.

### DIFF
--- a/src/Cake.Common/Text/TextTransformationExtensions.cs
+++ b/src/Cake.Common/Text/TextTransformationExtensions.cs
@@ -6,6 +6,9 @@ using Cake.Core.Text;
 
 namespace Cake.Common.Text
 {
+    using System;
+    using System.Collections.Generic;
+
     /// <summary>
     /// Contains extension methods for <see cref="TextTransformation{TTemplate}"/>.
     /// </summary>
@@ -27,6 +30,32 @@ namespace Cake.Common.Text
             where TTemplate : class, ITextTransformationTemplate
         {
             transformation?.Template.Register(key, value);
+            return transformation;
+        }
+
+        /// <summary>
+        /// Registers all keys and values in the dictionary for text transformation.
+        /// </summary>
+        /// <typeparam name="TTemplate">The text transformation template.</typeparam>
+        /// <param name="transformation">The text transformation.</param>
+        /// <param name="tokens">The tokens.</param>
+        /// <returns>
+        /// The same <see cref="TextTransformation{TTemplate}" /> instance so that multiple calls can be chained.
+        /// </returns>
+        public static TextTransformation<TTemplate> WithTokens<TTemplate>(
+            this TextTransformation<TTemplate> transformation, IDictionary<string, object> tokens)
+            where TTemplate : class, ITextTransformationTemplate
+        {
+            if (tokens == null)
+            {
+                throw new ArgumentNullException(nameof(tokens));
+            }
+
+            foreach (var kvp in tokens)
+            {
+                transformation?.Template.Register(kvp.Key, kvp.Value);
+            }
+
             return transformation;
         }
     }
--- a/src/Cake.Common.Tests/Unit/Text/TextTransformationExtensionsTests.cs
+++ b/src/Cake.Common.Tests/Unit/Text/TextTransformationExtensionsTests.cs
@@ -9,6 +9,8 @@ using Xunit;
 
 namespace Cake.Common.Tests.Unit.Text
 {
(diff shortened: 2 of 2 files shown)


src/Cake.Common/Text/TextTransformationExtensions.cs
```diff
@@ -29,5 +31,22 @@ public static class TextTransformationExtensions
...
+        /// <summary>
+        /// Registers all keys and values in the dictionary for text transformation.
+        /// </summary>
+        /// <typeparam name="TTemplate">The text transformation template.</typeparam>
+        /// <param name="transformation">The text transformation.</param>
+        /// <param name="tokens">The tokens.</param>
+        /// <returns>
+        /// The same <see cref="TextTransformation{TTemplate}" /> instance so that multiple calls can be chained.
+        /// </returns>
+        public static TextTransformation<TTemplate> WithTokens<TTemplate>(
```
patriksvensson: I would prefer if we didn't add more methods to the interface since that is an breaking change for the ABI. It would probably be better if we added this as an extension method since it just delegate to the other Register method.

patriksvensson: Commented on the wrong line.

src/Cake.Core/Text/ITextTransformationTemplate.cs
```diff
@@ -16,6 +18,12 @@ public interface ITextTransformationTemplate
         /// <param name="value">The value.</param>
         void Register(string key, object value);
 
+        /// <summary>
+        /// Registers a dictionary of key values
+        /// </summary>
+        /// <param name="tokens">The dictionary of tokens</param>
+        void Register(IDictionary<string, object> tokens);
```
patriksvensson: Above comment should be here.

maldworth:
Updated as requested in feedback.

maldworth:
@gep13 No open issue. I just thought this would be helpful to have this option

@patriksvensson I believe I made the changes as requested. Can you take another look when you have a chance?

thank you

gitfool:
@gep13 I recently added a similar request in #2375 before discovering this PR just now.

@maldworth @patriksvensson can we sort out the final details and merge this in time for Cake 0.31? Before that however, I do have a couple of suggestions for my own requirements:

As per the sample code in my issue, this extension method would be much more convenient and useful if:
- do not throw if the `tokens` parameter is `null`
- do not register tokens without a value

Re the second point, from experience I've found it much cleaner and safer to leave token placeholders behind for tokens without a value. This is especially helpful if staging a template after rendering, before publishing as an artifact, as it helps to diagnose issues at build-time.

maldworth:
@gitfool if it was silent to missing tokens or null params, then perhaps it would be a .TryWithTokens(). Or perhaps add a parameter to ignore missing tokens? anyways it would be nice to see this merged in sooner rather than later. Or we can combine both into one PR. Doesn't matter to me.

gitfool:
@maldworth your suggested behavior is problematic as [TextTransformationTemplate.Replace](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Text/TextTransformationTemplate.cs#L107) already silently replaces tokens without a value, technically with a `null` value, with empty strings. These can be hard to spot, from recent experience, whereas leaving the token placeholders behind clearly stands out. Note that that would also be equivalent to not registering the token at all, which can happen, so has a nice consistency to it. (This is not a problem when calling [TextTransformationTemplate.Render](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Text/TextTransformationTemplate.cs#L57) directly as you can conditionally call it.)

I suppose the other option would be to throw an exception if a token value is `null`, and ideally this behaviour is managed and enforced in one place, by `TextTransformationTemplate` itself, so you can register tokens with `null` values, either directly or via the proposed dictionary extension, and `Render` will later enforce its policy. This would require plumbing settings and may complicate the API unnecessarily, so let's wait for feedback from the Cake maintainers.

patriksvensson:
This have been merged. Sorry for the delay...
Thank you for your contribution 👍

gep13:
As a result of this PR https://github.com/cake-build/cake/pull/2509 the overall aim of this PR has changed a little bit, so I have updated the title of this one (which will be included in the release notes) to reflect the final outcome.

Issue #2134, opened 2018-04-23
Assembly conflicts during compilation


daveaglick:
### What You Are Seeing?

@redth reports the following messages during script compilation:
```
error CS1705: Assembly 'Cake.Json' with identity 'Cake.Json, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null' uses 'Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' which has a higher version than referenced assembly 'Newtonsoft.Json' with identity 'Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'
```

And then the compilation fails with:

```
Error: Cake.Core.CakeException: Error(s) occurred when compiling build script:
  at Cake.Scripting.Roslyn.RoslynScriptSession.Execute (Cake.Core.Scripting.Script script) [0x00182] in <f2dfd1d7ff874bb5925674ed0d0fc83d>:0 
  at Cake.Core.Scripting.ScriptRunner.Run (Cake.Core.Scripting.IScriptHost host, Cake.Core.IO.FilePath scriptPath, System.Collections.Generic.IDictionary`2[TKey,TValue] arguments) [0x00358] in <bf8648e2e0a54f859ebcbb1728e8c9cf>:0 
  at Cake.Commands.BuildCommand.Execute (Cake.CakeOptions options) [0x00026] in <f2dfd1d7ff874bb5925674ed0d0fc83d>:0 
  at Cake.CakeApplication.Run (Cake.CakeOptions options) [0x00015] in <f2dfd1d7ff874bb5925674ed0d0fc83d>:0 
  at Cake.Program.Main () [0x000d1] in <f2dfd1d7ff874bb5925674ed0d0fc83d>:0 
```

### What is Expected?

Successful compilation.

### What version of Cake are you using?

0.27.1

### Output Log

https://gist.github.com/Redth/37bfa030e2890d22d93c06b154648e72

---

[...]

PR #2139, opened 2018-04-24
GH2134: References latest Newtonsoft.Json from Cake.NuGet

daveaglick:
Confirmed this results in `Newtonsoft.Json` 11.0.2 ending up in Cake bin for net46 target framework. This should resolve 90% of cases, but does not introduce a long-term fix of building a reference dependency graph and loading the newest (that'll take some time).

#2134


devlead:
@daveaglick your changes have been merged, thanks for your contribution 👍

Issue #2140, opened 2018-04-24
DotNetCorePublish does not respect SelfContained DotNetCorePublishSettings property


klabranche:
### What You Are Seeing?
Setting the SelfContained property for DotNetCorePublishSettings does not emit the respective parameter when executed by DotNetCorePublish:

     var settings = new DotNetCorePublishSettings
     {
         Framework = "netcoreapp2.0",
         Runtime = "win-x64",
         SelfContained = false, 
         Configuration = $"{configuration}",
         OutputDirectory = artifactsRootDir + "/dist"
     };

    DotNetCorePublish("./src/" + appName + "/" + appName + ".csproj",settings);

What it produces at the command line:

`Executing: "C:/Program Files/dotnet/dotnet.exe" publish "./src/ProjectName/ProjectName.csproj" --output "C:/projects/ProjectName/artifacts/debug/dist" --runtime win-x64 --framework netcoreapp2.0 --configuration debug`

The --self-contained false parameter is missing.  When --self-contained is omitted defaults to true when using --runtime.

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish?tabs=netcore2x

This results in creating a self-contained publish.

### What is Expected?

DotNetCorePublish to produce:

`Executing: "C:/Program Files/dotnet/dotnet.exe" publish "./src/ProjectName/ProjectName.csproj" --output "C:/projects/ProjectName/artifacts/debug/dist" --runtime win-x64 --framework netcoreapp2.0 --configuration debug --self-contained false`

### What version of Cake are you using?

Version 0.26.1+Branch.main.Sha.b26790017b2befa7b35799eb9a0c1cba322b0c90

and 

 Version 0.27.1+Branch.main.Sha.7e44aa0319b6ef12ecc415d79a1c75cc98e2c788

### Are you running on a 32 or 64 bit system?
x64

### What environment are you running on?  Windows? Linux? Mac?

Windows

### Are you running on a CI Server?  If so, which one?
Not yet....  just locally atm...

### How Did You Get This To Happen? (Steps to Reproduce)

have a task to package .net core 2.0 command line app like below:

```
Task("Package")
	.Description("packages app")  
	.Does(() =>
{
     var settings = new DotNetCorePublishSettings
     {
         Framework = "netcoreapp2.0",
         Runtime = "win-x64",
         SelfContained = false,
         Configuration = $"{configuration}",
         OutputDirectory = artifactsRootDir + "/dist"
     };

[...]

Issue #2136, opened 2018-04-24
DotNetCoreTest fails to use XUnit2Settings on Cake 0.27


erik-am:
### What You Are Seeing?

I'm using the `DotNetCoreTest` command using both `DotNetCoreTestSettings` as well as `Xunit2Settings`:

```
DotNetCoreTest(dotnetTestSettings, project, xunitSettings);
```

which results in the following exception:

```
Error: One or more errors occurred.
        Method not found: 'Cake.Common.Tools.XUnit.ParallelismOption Cake.Common.Tools.XUnit.XUnit2Settings.get_Parallelism()'.
```

### What is Expected?

Expected is that the tests will be executed.

### What version of Cake are you using?

The error occurs in 0.27.0 and 0.27.1.
The error does _not_ occur in 0.26.1.

I think the regression has been introduced by the change of `ParallelismOption` to a nullable.

See https://github.com/cake-build/cake/pull/2040/files#diff-133ec58ad83fd466eed3b129d32daefaL84

### Are you running on a 32 or 64 bit system?

64 bit

### What environment are you running on?  Windows? Linux? Mac?

Windows

### Are you running on a CI Server?  If so, which one?

Yes, Jenkins.

### How Did You Get This To Happen? (Steps to Reproduce)

See "What You Are Seeing?".

###

kcamp:
@erik-am It would help to have more of the stack trace and/or verbose output, but it appears that the exception is coming out of the `Cake.Incubator` assembly at the method below, right?
```
void DotNetCoreTest(this ICakeContext context, FilePath project, XUnit2Settings xunitSettings)
```

erik-am:
Thank you @kcamp, indeed it seems the issue is in `Cake.Incubator`. So, Cake 0.27.0 did include a breaking change in contrast to the release notes,

For now, I changed my usage of `DotNetCoreTest`, so that I'm using the method provided by Cake itself instead of the extension. Thanks for pointing me in the right direction! 👍

gep13:
@wwwlicious are you in a position to take a look at this?

@erik-am unfortunately, this does look like it was an unintended breaking change in Cake 0.27.0 😢 

@cake-build/cake-team FYI...

wwwlicious:
@gep13 I'll have a look..

PR #2142, opened 2018-04-25
GH2140: PR DotNetCorePublish does not respect SelfContained DotNetCorePublishSettings property #2140

klabranche:
Made SelfContainedproperty of DotNetPublishSettings a nullable boolean since no setting means it takes…the .Net default of true when runtime is included.  Added the logic to include false for self contain and test when running DotNetPublish.

https://github.com/cake-build/cake/issues/2140

### DIFF
--- a/src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublisher.cs
+++ b/src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublisher.cs
@@ -114,9 +114,17 @@ namespace Cake.Common.Tools.DotNetCore.Publish
             }
 
             // Self contained
-            if (settings.SelfContained)
+            if (settings.SelfContained.HasValue)
             {
-                builder.Append("--self-contained");
+                if (settings.SelfContained.Value)
+                {
+                    builder.Append("--self-contained");
+                }
+                else
+                {
+                    builder.Append("--self-contained");
+                    builder.Append("false");
+                }
             }
 
             // Sources
--- a/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Publish/DotNetCorePublisherTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Publish/DotNetCorePublisherTests.cs
@@ -125,6 +125,30 @@ namespace Cake.Common.Tests.Unit.Tools.DotNetCore.Publish
             }
 
             [Fact]
+            public void Should_Add_SelfContained_False_Settings()
+            {
+                // Given
+                var fixture = new DotNetCorePublisherFixture();
+                fixture.Settings.NoDependencies = true;
+                fixture.Settings.NoRestore = true;
+                fixture.Settings.Framework = "dnxcore50";
+                fixture.Settings.Configuration = "Release";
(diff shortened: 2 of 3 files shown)


src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublisher.cs
```diff
@@ -114,11 +114,19 @@ private ProcessArgumentBuilder GetArguments(string path, DotNetCorePublishSettin
...
+            if (settings.SelfContained.HasValue)
             {
-                builder.Append("--self-contained");
+                if (settings.SelfContained.Value)
+                {
+                    builder.Append("--self-contained");
+                }
+                else
+                {
+                    builder.Append("--self-contained false");
```
patriksvensson: Can we divide this into two separate appends instead?

klabranche: You mean two separate ifs, one for true and another if for false?

kcamp:
As discovered in GH-2136, this (change from `bool` to `bool?`) will be a breaking change.  See GH-2039 and GH-2040.

src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublisher.cs
patriksvensson: No, like this:

```csharp
builder.Append("--self-contained");
builder.Append("false");
```

patriksvensson:
@kcamp Correct, but I think the impact will be minimal from changing this property (especially since it hasn't worked properly before).

kcamp:
No prob.  Just trying to avoid a repeat.  :+1:

klabranche:
@kcamp is this a breaking change, it's not working as is....  This wouldn't create a regression at all?

src/Cake.Common/Tools/DotNetCore/Publish/DotNetCorePublisher.cs
klabranche: oh... sure

klabranche:
Always nice to see every build fail.... Anything I can do @patriksvensson?  Assuming not....

kcamp:
@klabranche  It looks like some extra whitespace (line break) in the style analysis.  You should be able to check build output via TC by logging in as guest.
```
[18:57:40][Step 1/1] Tools\DotNetCore\Publish\DotNetCorePublisher.cs(129,1): error SA1507: Code should not contain multiple blank lines in a row [Z:\buildAgent\work\218bb411df7337f7\src\Cake.Common\Cake.Common.csproj]
[18:57:40][Step 1/1] Tools\DotNetCore\Publish\DotNetCorePublisher.cs(129,1): error SA1507: Code should not contain multiple blank lines in a row [Z:\buildAgent\work\218bb411df7337f7\src\Cake.Common\Cake.Common.csproj]
```

klabranche:
OH.... didn't know that but also read the contribs and didn't realize I had some extra lines... my bad....

klabranche:
All good to go it looks like. :-)

devlead:
@klabranche  your changes have been merged, thanks for your contribution 👍

Issue #2141, opened 2018-04-25
Add Verbosity property to GitVersionSettings


WhiteWaterCoder:
### What You Are Seeing?

GitVersionSettings does not contain a property for `verbosity` which means that the only way to increase the default logging level is through `ArgumentCustomization`.

[Gitter link](https://gitter.im/cake-build/cake?at=5ae0915d109bb04332fd7f14)

### What is Expected?

A Verbosity property of type GitVersionVerbosity which would add the `verbosity` flag. It would also be handy to set the default value of the property to match that of Cake's verbosity.

### What version of Cake are you using?

0.25 (will upgrade to 0.27.1 soon)

### Are you running on a 32 or 64 bit system?

x64

### What environment are you running on?  Windows? Linux? Mac?

Windows

### Are you running on a CI Server?  If so, which one?

Both locally and CI (TeamCity)

### How Did You Get This To Happen? (Steps to Reproduce)

This is a new feature so not applicable.

### Output Log

n/a

PR #2145, opened 2018-04-26
GH2141: Added Verbosity property to GitVersionSettings

WhiteWaterCoder:
Pull request to address [2141](https://github.com/cake-build/cake/issues/2141)

A Verbosity property has been added to GitVersionSettings which can be explicitly set otherwise it defaults depending on the verbosity of Cake itself. The rationale behind defaulting is in CI environments (TeamCity in my case) where it is handy to increase verbosity of logging without having to modify the script.


gep13:
@WhiteWaterCoder your changes have been merged, thanks for your contribution 👍

PR #2147, opened 2018-04-30
GH2146: Enabled initializer syntax for all collection properties

### LINKED ISSUE #2146: Enabling initializer syntax for all collection properties
It's intuitive to try this super clean C#:

```cs
new FooToolSettings
{
    EnvironmentVariables = { ["FOO"] = "foo" }
}
```

`EnvironmentVariables`, like (almost?) every Cake settings collection property, is not initialized in the constructor. Thus, this code results in NullReferenceException.

The workaround doesn't leave me with the sense of happiness I expect from a mutable C# API:
```cs
new FooToolSettings
{
    EnvironmentVariables = new Dictionary<string, string> { ["FOO"] = "foo" }
}
```

### Proposal

Please initialize all collection properties (including dictionary properties) either on construction or lazily in the getter. [For example](https://github.com/cake-build/cake/blob/52bd7efc142f47887bdd8d73ef8f508e12c5d654/src/Cake.Core/Tooling/ToolSettings.cs#L69):

```cs
public IDictionary<string, string> EnvironmentVariables { get; set; } =
    new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```

### Recommending not to use the setters

Due to backwards compatibility the setters must remain, but is there room to add XML docs or maybe even an obsoletion warning recommending that you use the pre-initialized instance?
To me there seems to be an abstraction issue with allowing this property to be set: what is the difference in meaning between these two lines, and why should I be offered this control as a user of Cake?

```cs
EnvironmentVariables = new Dictionary<string, string>(StringComparer.CurrentCulture) { ["FOO"] = "foo" }
EnvironmentVariables = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase) { ["FOO"] = "foo" }
```

-----

If this sounds good to you, I'm willing to add the tests and do the PR.
I'd add cross-cutting tests that dynamically locate each class assignable to `ToolSettings`, instantiate it, get the value of each `ICollection<>` property, and assert that it is not null. There could be a whitelist of properties to ignore if the need ever arises.

jnm2:
Fixes #2146 

Only eleven properties needed to be changed! I had run into four of them in real life.

### DIFF
--- a/src/Cake.Common.Tests/CrossCutting/ToolSettingsTests.cs
+++ b/src/Cake.Common.Tests/CrossCutting/ToolSettingsTests.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Cake.Core.Tooling;
+using Cake.Testing;
+using Cake.Testing.Extensions;
+using Xunit;
+
+namespace Cake.Common.Tests.CrossCutting
+{
+    public static class ToolSettingsTests
+    {
+        // Ensures that C# initializer syntax will not throw NullReferenceException when collection properties are used.
+        [Theory]
+        [MemberData(nameof(ToolSettingsTypes))]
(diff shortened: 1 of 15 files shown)


jnm2:
Rebased, ping for review? The TeamCity build failed, but I can't see details.

devlead:

jnm2:
@devlead Thanks for helping!

Issue #2154, opened 2018-05-16
Problem with loading abolute path scripts with #load preprocessor


Kemyke:
### What You Are Seeing?
Loading a script with absolute path is only working if the full definition is applied: `#load \"local:?path=c:/utils.cake\"`. It is not working with legacy format `#load \"c:/utils.cake\"`.

### What is Expected?
It should work for both syntaxes.

### What version of Cake are you using?
0.27.1

### Are you running on a 32 or 64 bit system?
64 bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No.

The problem is with the `ParseUriFromTokens` method in `UriDirectiveProcessor` class. It first tries to create `Uri` from the token. If it is a relative path the creation will fail and the `CreateUriFromLegacyFormat` will be applied because of the `UriKind.Absolute` parameter in `TryCreate` call. But if the token is an absolute path, the Uri creation will success and the `Uri` will be `file:///c:/CakeScripts/utils.cake`. But this `Uri` won't be processed because in `FileLoadDirectiveProvider` class the `CanLoad` method will return false because the `Scheme` will be `file` and not `local` as expected.
Maybe the `Uri.TryCreate` is not the best way to decide if the `CreateUriFromLegacyFormat` should be applied or not. Maybe it would be better to have a previous step which check if the token is match with the appropriate regexp. (somtehing like `[local|nuget]:\?`)

devlead:
fixed by #2155

PR #2155, opened 2018-05-16
GH2154: Fix problem with loading abolute path scripts with #load preprocessor

Kemyke:
I did a fix for the issue https://github.com/cake-build/cake/issues/2154

I am not sure if the Regex is complete and valid. But i think this is a better approach then depending on the Uri.TryCreate output to decide if it is a legacy formatted Uri or not.

### DIFF
--- a/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
+++ b/src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
@@ -4,6 +4,7 @@
 
 using Cake.Core.Scripting.Processors.Loading;
 using Cake.Core.Tests.Fixtures;
+using Cake.Testing.Xunit;
 using Xunit;
 
 namespace Cake.Core.Tests.Unit.Scripting.Processors
@@ -136,6 +137,82 @@ namespace Cake.Core.Tests.Unit.Scripting.Processors
             Assert.Equal(result.Lines[10], "#line 2 \"/Working/a.cake\"");
             Assert.Equal(result.Lines[11], "// #l b.cake");
             Assert.Equal(result.Lines[12], "int y=2;");
+        }
+
+        [Theory]
+        [InlineData("#load \"/utils.cake\"")]
+        [InlineData("#load \"local:?path=/utils.cake\"")]
+        public void Should_Process_AbsolutePath_Script_Reference_Found_In_Source(string source)
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
+        [Theory]
+        [InlineData("#load \"test/utils.cake\"")]
+        [InlineData("#load \"local:?path=test/utils.cake\"")]
+        public void Should_Process_RelativePath_Script_Reference_Found_In_Source(string source)
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("/Working/test/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/Working/test/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
+        [WindowsTheory]
+        [InlineData("#load \"c:/utils.cake\"")]
+        [InlineData("#load \"local:?path=c:/utils.cake\"")]
+        public void Should_Process_WindowsAbsolutePath_Script_Reference_Found_In_Source(string source)
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("c:/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("c:/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
+        [WindowsTheory]
+        [InlineData("#load \"test/utils.cake\"")]
+        [InlineData("#load \"local:?path=test/utils.cake\"")]
+        public void Should_Process_WindowsRelativePath_Script_Reference_Found_In_Source(string source)
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            fixture.Providers.Add(new FileLoadDirectiveProvider());
+            fixture.GivenScriptExist("/Working/script.cake", source);
+            fixture.GivenScriptExist("/Working/test/utils.cake", "Console.WriteLine();");
+
+            // When
+            var result = fixture.Analyze("/Working/script.cake");
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/Working/test/utils.cake", result.Script.Includes[0].Path.FullPath);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
```diff
@@ -6,12 +6,15 @@
...
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using Cake.Core.Scripting.Analysis;
 
 namespace Cake.Core.Scripting.Processors
 {
     internal abstract class UriDirectiveProcessor : LineProcessor
     {
+        private Regex uriPrefixPattern = new Regex("(nuget|local|npm):");
```
patriksvensson: The `UriDirectiveProcessor` doesn't know anything about nuget, local or npm, so the regex should check for ANY scheme. Also, We generally makes members `readonly` and initialize them in the constructor.

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
```diff
@@ -57,13 +60,22 @@ public sealed override bool Process(IScriptAnalyzerContext context, string line,
             return false;
         }
 
+        private bool IsUriFromLegacyPattern(string[] tokens)
```
patriksvensson: Can we move this method below `ParseUriFromTokens`?

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
```diff
@@ -137,5 +138,41 @@ public void Should_Insert_Line_Directives_When_Processing_Load_Directives()
...
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal($"{absPath}/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
+        [Fact]
+        public void Should_Process_AbsolutePath_Script_Reference_Found_In_Source()
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            var absPath = Directory.GetCurrentDirectory().Replace('\\', '/');
```
patriksvensson: Don't use System.IO here and don't get the user's current directory.

The test fixture is using a fake filesystem, so simply declare an absolute path like this `/Absolute/utils.cake`.

If you want to test Windows paths, you will need to create a Windows specific test with `WindowsFact` or `WindowsTheory` and declare an absolute path like this `C:/utils.cake`.

Kemyke: Done

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
```diff
@@ -137,5 +138,41 @@ public void Should_Insert_Line_Directives_When_Processing_Load_Directives()
             Assert.Equal(result.Lines[11], "// #l b.cake");
             Assert.Equal(result.Lines[12], "int y=2;");
         }
+
+        [Fact]
+        public void Should_Process_AbsolutePath_Script_Reference_Found_In_Source_LegacyFormat()
+        {
+            // Given
+            var fixture = new ScriptAnalyzerFixture();
+            var absPath = Directory.GetCurrentDirectory().Replace('\\', '/');
```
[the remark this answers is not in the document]
Kemyke: Done

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
Kemyke: Done

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
Kemyke: Maybe in the `FileLoadDirectiveProvider` the `file` should be handled?

patriksvensson: Something like `^([a-zA-Z]{2,}:)` would probably solve it. Windows drives can only be one character long and schemes that we accept has to be longer than one character.

Kemyke: Thank you for the idea. I've just checked in the new pattern.

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
```diff
@@ -6,12 +6,15 @@
...
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using Cake.Core.Scripting.Analysis;
 
 namespace Cake.Core.Scripting.Processors
 {
     internal abstract class UriDirectiveProcessor : LineProcessor
     {
+        private Regex uriPrefixPattern = new Regex("^([a-zA-Z]{2,}:)");
```
patriksvensson: Could you move the assignment to the constructor of this one and rename it to `_uriPrefixPattern`. The variable should also be `readonly`. Thanks.

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
```diff
@@ -137,5 +138,43 @@ public void Should_Insert_Line_Directives_When_Processing_Load_Directives()
             Assert.Equal(result.Lines[11], "// #l b.cake");
             Assert.Equal(result.Lines[12], "int y=2;");
         }
+
+        [Theory]
+        [InlineData("#load \"/utils.cake\"")]
+        [InlineData("#load \"local:?path=/utils.cake\"")]
+        public void Should_Process_AbsolutePath_Script_Reference_Found_In_Source(string source)
```
patriksvensson: Can we also add tests for relative file paths?

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
```diff
@@ -137,5 +138,43 @@ public void Should_Insert_Line_Directives_When_Processing_Load_Directives()
...
+
+            // Then
+            Assert.Equal(1, result.Script.Includes.Count);
+            Assert.Equal("/utils.cake", result.Script.Includes[0].Path.FullPath);
+        }
+
+        [WindowsTheory]
+        [InlineData("#load \"c:/utils.cake\"")]
+        [InlineData("#load \"local:?path=c:/utils.cake\"")]
+        public void Should_Process_WindowsAbsolutePath_Script_Reference_Found_In_Source(string source)
```
patriksvensson: Can we also add tests for relative file paths?

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
Kemyke: Isn't `Should_Process_Single_Script_Reference_Found_In_Source` and `Should_Process_Single_Script_Reference_With_Spaces_In_File_Name_Found_In_Source` also test the loading of relative pathes? That's why i add tests only with absolute path. If you think that it would be useful to test relative pathes in these new tests, i can add them. But then i need to rename the methods because now they have `AbsolutePath` part in their name.

src/Cake.Core/Scripting/Processors/UriDirectiveProcessor.cs
Kemyke: Done

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
Kemyke: I am thinking how to do this correctly. All the other tests only works for one script path. Because of the second `GivenScriptExist` call eg: `fixture.GivenScriptExist("c:/utils.cake", "Console.WriteLine();");`

If i add relative path to this test, then first i should parse the directive and get the path from it, just to set up the test context. Do you have any idea how should i do that?

patriksvensson: Easiest would just be to duplicate the two tests you already written and change to relative paths.

Kemyke: Yeah, first i tried to solve it in one test. But splitting it into two is straightforward. :)

Kemyke: Done

src/Cake.Core.Tests/Unit/Scripting/Processors/LoadDirectiveProcessorTests.cs
Kemyke: Done

patriksvensson:
Looks good to me! 👍

Could you rebase the commits into a single one and give it a descriptive comment and we'll be ready to merge this.

Kemyke:
May i ask what is the state of this PR? Do you need some further modification? Or are you planning to associate a milestone to this PR?

devlead:
@Kemyke your changes have been merged, thanks for your contribution 👍

PR #2167, opened 2018-05-25
GH2196: Add repository element to nuspec file in NuGetPack tool

rh072005:


rh072005:
Having read more about [SourceLink](https://github.com/ctaggart/SourceLink) this may not actually be needed. If it's useful that's great but if not feel free to close

devlead:
Is there's an issue to go long with this PR? (as required per [Contributing guidelines](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md))

rh072005:
Apologies I'd missed that part of the contrib guide. I'll put a note in Gitter now to see if this is something people are interested in and then get an issue raised if so.

patriksvensson:
@rh072005 The reason for an issue is not (always) to get buy-in, but because we need an issue for our release process. I see no issue with adding this functionality so if you create an issue for this functionality we will merge it after a review :+1:

src/Cake.Common/Tools/NuGet/Pack/NugetRepository.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace Cake.Common.Tools.NuGet.Pack
+{
+    /// <summary>
+    /// Specifies the package's source code location, allowing IDEs to download and debug the code.
+    /// </summary>
+    public class NugetRepository
```
Meir017: Shouldn't this be `NuGetRepository` instead?

devlead:
@rh072005  your changes have been merged, thanks for your contribution 👍

PR #2182, opened 2018-06-07
try resolve msbuild.exe before guessing it

### LINKED ISSUE #2181: try resolve msbuild.exe before guessing it
```C#
#tool nuget:?package=vswhere
...
Setup(context =>
{

    var msbuildToolsPath  = VSWhereProducts("*", new VSWhereProductSettings { Requires = "Microsoft.Component.MSBuild"})
        .FirstOrDefault()
        ?.CombineWithFilePath(@".\MSBuild\15.0\Bin\MSBuild.exe")
        ;

    context.Tools.RegisterFile(msbuildToolsPath);
});
…
Task("Build-Solution")
    .Does(() =>
{
    MSBuild(solutionPath, settings =>
        {
            settings
                .SetConfiguration(configuration)
                .WithTarget("Restore")
                .WithTarget("Rebuild")
                .WithTarget("Pack")
            ;
        });
});
```

The script above doesn't work because the `MSBuildRunner` doesn't provide a list of executable names to be resolved.

https://github.com/cake-build/cake/blob/main/src/Cake.Common/Tools/MSBuild/MSBuildRunner.cs#L284-L287

paulomorgado:
Fixes #2181 

P.S.: How should I unit test this?


paulomorgado:
Added unit tests.

patriksvensson:
@paulomorgado I'm not sure that we actually want to get "the first best" msbuild that we find on PATH, since resolution of MSBuild is based on toolversion, architecture and similar. We wouldn't want another MSBuild version to be returned to the user if they ask for a specific one (which could happen).

I think the solution to the original problem stated in the issue would be to set the [ToolPath](https://cakebuild.net/api/Cake.Core.Tooling/ToolSettings/F6956DA8) of the MSBuildSettings (inherited from ToolSettings) with the path resolved by VSWhere.

paulomorgado:
Why not add a setting to the MSBuild tool to try to use the tool locator before trying to guess the path?

PR #2227, opened 2018-08-08
GH2207: Update to NuGet 4.7.0

### LINKED ISSUE #2207: NuGet hang due to bug in NuGet 4.6.0
Cake hangs when restoring certain signed NuGet packages. The longest I waited before killing was between 5 and 10 minutes. A repro is a .cake script with `#tool WiX.Toolset.UnofficialFork`. The latest version of that package, 3.11.1, is the first signed version.

I verified that the bug is present in https://dist.nuget.org/win-x86-commandline/v4.6.0/nuget.exe but not in https://dist.nuget.org/win-x86-commandline/v4.6.1/nuget.exe, running `nuget.exe verify -signatures wix.toolset.unofficialfork.3.11.1.nupkg` against each.

Cake currently bundles NuGet 4.6.0 assemblies, so it seems that updating its NuGet dependencies to 4.6.2 might resolve this problem.

/cc @gep13

mholo65:
- Fixes #2207

### DIFF
--- a/src/Cake.NuGet/Cake.NuGet.csproj
+++ b/src/Cake.NuGet/Cake.NuGet.csproj
@@ -19,17 +19,17 @@
   </ItemGroup>
   <!-- Global packages -->
   <ItemGroup>
-    <PackageReference Include="NuGet.Frameworks" Version="4.6.0" />
-    <PackageReference Include="NuGet.Versioning" Version="4.6.0" />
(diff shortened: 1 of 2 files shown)


mholo65:
@jnm2, could you verify if this works for you? Artifacts available via AppVeyor.

mholo65:
@jnm2, np! Glad that this worked!

devlead:

PR #2231, opened 2018-08-15
GH2228: Fix CakeTaskExtensions accessibility

### LINKED ISSUE #2228: CakeTaskExtensions are no longer accessible
### What You Are Seeing?

`CakeTaskExtensions` are no longer accessible.

### What is Expected?

`CakeTaskExtensions` to be accessible; as before upgrading.

0.29.0

64-bit

Windows

TeamCity

I recently updated from Cake 0.23.0 to 0.29.0 and hit the following breaking change:
[commit/105eb540d1958f11d7953000ace4377042f192a6#diff-f817cf6c892a63876e9847920aae68eaR8](https://github.com/cake-build/cake/commit/105eb540d1958f11d7953000ace4377042f192a6#diff-f817cf6c892a63876e9847920aae68eaR8)

The `CakeTaskExtensions` class accessibility was changed from `public` to `internal` and we have lots of builds scripts using it; for example using `AddCriteria` to disable a task in order to override its behaviour:

```c#
Tasks.First(x => x.Name == "Zip-DataStore")
    .AddCriteria(() => false);
```
I can hack around it but it's not pretty:
```c#
((CakeTask)Tasks.First(x => x.Name == "Zip-DataStore"))
    .Criterias.Add(new CakeTaskCriteria(_ => false, null));
```

gitfool:
Fixes #2228

### DIFF
--- a/src/Cake.Core/CakeTaskExtensions.cs
+++ b/src/Cake.Core/CakeTaskExtensions.cs
@@ -5,8 +5,18 @@ using System.Threading.Tasks;
 
 namespace Cake.Core
 {
+    /// <summary>
+    /// Contains extension methods for <see cref="CakeTask"/>.
+    /// </summary>
     public static class CakeTaskExtensions
     {
+        /// <summary>
+        /// Adds the dependency to the task's dependencies.
+        /// </summary>
+        /// <param name="task">The task.</param>
+        /// <param name="name">The name of the dependency .</param>
+        /// <param name="required">Whether or not the dependency is required.</param>
+        /// <exception cref="CakeException">The task already has the dependency.</exception>
         public static void AddDependency(this CakeTask task, string name, bool required = true)
         {
             if (task.Dependencies.Any(x => x.Name == name))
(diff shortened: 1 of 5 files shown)


gitfool:
The TeamCity build failed with an unrelated error run `git clean`.

src/Cake.Core/CakeTaskExtensions.cs
```diff
@@ -29,15 +46,24 @@ public static void AddDependee(this CakeTask task, string name, bool required =
...
+        /// <param name="predicate">The criteria's predicate.</param>
+        /// <param name="message">The criteria's message if skipped.</param>
+        public static void AddCriteria(this CakeTask task, Func<ICakeContext, bool> predicate, string message = null)
         {
-            if (criteria == null)
-            {
-                throw new ArgumentNullException(nameof(criteria));
-            }
-            task.Criterias.Add(new CakeTaskCriteria(criteria, message));
+            task.Criterias.Add(new CakeTaskCriteria(predicate, message));
```
gitfool: Note:
- I renamed the parameter from `criteria` to `predicate` to match the parameter name in the `CakeTaskCriteria` ctor
- I removed the check for null here because the `CakeTaskCriteria` ctor already does that

devlead:
@gitfool  your changes have been merged, thanks for your contribution 👍

Issue #1383, opened 2016-11-26
Add command line option to display build target graph


colin-sim:
Add a command line option, for example, ```--tree```, which displays the entire dependency graph of all tasks in an "ASCII art" format. The example below is of a build script that consists of two "root" build tasks: **Default** and **Release**.
```
Default
  |-NugetRestore
  |-Build
  |-Test
Release
  |-NugetRestore
  |-Build
  |-Test
  |-Package
  |-Publish
```
This is useful when the build script gets bigger, as in the case at our work where we have code in C#, C++, and so on. If tests, packaging, and so on, are added into the script for the entire product, it may be difficult to see all targets or see what's available. So the ```--tree``` command will show all targets available and what's done for each target when invoked. The tree will contain at least one root.

phillipsj:
I am willing to try to put together a PR if this is something that the project wants.

gep13:
@phillipsj sure thing. Consider it yours 😄

flcdrg:
If anyone is wondering, the argument is `--showtree`

Issue #1506, opened 2017-02-23
Split tasks in main and sub tasks


pascalberger:
The [gulp-task-listing](https://www.npmjs.com/package/gulp-task-listing) plugin has a concept of main and sub-tasks which IMHO improves readability of a task listing like `showdescription` a lot. It would be nice if Cake can support a similar feature.

pascalberger:
Closed in favor of #1383

PR #2248, opened 2018-08-22
Add command line option to display build target graph

andrewlock:
First off, thanks for Cake - I use it for all my personal projects and in my day job, so really appreciate the work you guys have put into it 🙂 

I've put together a PR for #1383 as it's something I'd find pretty useful. I made the command line option `--showtree` or `--tree`. Most of the implementation is pretty straight forward, but there's one point in particular I'm not sure about. 

The existing `CakeGraph` is one of the easiest ways to get the "top level" tasks. However, currently the `CakeGraph` and `CakeGraphBuilder` are `internal` classes of `Cake.Core`. In order to make it available to `Cake`, we could either 

* Make `CakeGraphBuilder`, `CakeGraph`, and `CakeGraphEdge` public
* Add `[InternalsVisibleTo("Cake")] to the `Cake.Core` project

I went for the latter, as it seemed the easiest!

The other point of interest is a `MaxDepth` constant that could be used to limit how deep the tree is. At the moment it's just "hidden" functionality (a constant set to `0`) as I wasn't sure whether it was worth exposing it.


patriksvensson:
@andrewlock Really cool and useful feature!

Would it be possible to use different foreground colors for delegating tasks (tasks without actions) and top level tasks? Perhaps gray for delegating and cyan for root tasks? I've already approved this PR, so if you feel like moving on, we can revisit this at a later point.

Great job! 👍 

@cake-build/cake-team Any additional feedback?

andrewlock:
Good idea - I've added this to the PR. IT now displays top-level tasks in Cyan, delegating in Gray, and standard in Green. 

The `Actions` and `DelayedActions` properties aren't exposed in `ICakeTaskInfo`, so I had to cast them to a `CakeTask` to find out if it's a delegating task. A bit hacky, but the other option of modifying the `ICakeTaskInfo` interface, or adding a new interface to `CakeTask` seemed t0o heavy-handed for a small feature.  

The output from the command line now looks something like this: 

image

Or with inverted colours (the console in VS Code):

image

pascalberger:
LGTM and I think fixes also the use case of #1506

PR #2299, opened 2018-09-19
(build) Unit test results artifacts

devlead:


devlead:
Many tests
https://dev.azure.com/cake-build/Cake/_build/results?buildId=1083&view=ms.vss-test-web.test-result-details

PR #2305, opened 2018-09-20
GH2303: Use .NETCoreApp instead of .NETStandard

### LINKED ISSUE #2303: Built framework for .NET Core should be NETCoreApp not NETStandard
I'd say [this](https://github.com/cake-build/cake/blob/9dd24e46aee8cf38a425f184b447ee00820bece8/src/Cake.Core/Polyfill/EnvironmentHelper.cs#L122) should be `NETCoreApp,Version=2.1`

mholo65:
- Fixes #2303

### DIFF
--- a/src/Cake.Core/Polyfill/EnvironmentHelper.cs
+++ b/src/Cake.Core/Polyfill/EnvironmentHelper.cs
@@ -119,7 +119,7 @@ namespace Cake.Core.Polyfill
         public static FrameworkName GetBuiltFramework()
         {
 #if NETCORE
-            return new FrameworkName(".NETStandard,Version=v2.0");
+            return new FrameworkName(".NETCoreApp,Version=v2.1");
 #else
             return new FrameworkName(".NETFramework,Version=v4.6.1");
 #endif
--- a/src/Cake.Core.Tests/Unit/CakeRuntimeTests.cs
+++ b/src/Cake.Core.Tests/Unit/CakeRuntimeTests.cs
@@ -26,7 +26,7 @@ namespace Cake.Core.Tests.Unit
                 var framework = runtime.BuiltFramework;
 
                 // Then
-                Assert.Equal(".NETStandard,Version=v2.0", framework.FullName);
+                Assert.Equal(".NETCoreApp,Version=v2.1", framework.FullName);
             }
         }
 
--- a/src/Cake.NuGet.Tests/Unit/NuGetContentResolverTests.cs
+++ b/src/Cake.NuGet.Tests/Unit/NuGetContentResolverTests.cs
@@ -118,7 +118,7 @@ namespace Cake.NuGet.Tests.Unit
             public void Should_Throw_If_Path_Is_Null()
             {
                 // Given
-                var fixture = new NuGetAddinContentResolverFixture(".NETStandard,Version=v1.6", Runtime.CoreClr);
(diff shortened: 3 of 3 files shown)


src/Cake.Core/Polyfill/EnvironmentHelper.cs
```diff
@@ -119,7 +119,7 @@ public static Runtime GetRuntime()
         public static FrameworkName GetBuiltFramework()
         {
 #if NETCORE
-            return new FrameworkName(".NETStandard,Version=v2.0");
+            return new FrameworkName(".NETCoreApp,Version=v2.1");
```
patriksvensson: Is Cake really `2.1`? Shouldn't this be `v2.0`?

devlead: Cake is 2.0 and Cake.Tool is 2.1.

patriksvensson: Yeah, but shouldn't this parameter reflect the framework we actually built Cake on?

devlead: Could we query runtime? As you may run on newer?

mholo65: Maybe this could help? https://weblog.west-wind.com/posts/2018/Apr/12/Getting-the-NET-Core-Runtime-Version-in-a-Running-Application I.e. doable at runtime, via reflection.

patriksvensson: No., the runtime version is not of interest here. It's the target framework of Cake. And like the property says, this should reflect what it's built target framework is.

mholo65: AFAIK, that is the built framework.
image

patriksvensson: Yes. `.NetCoreApp, Version=v2.0`

mholo65: And for `Cake.Tool` it shows `.NetCoreApp, Version=v2.1`. So, we can use this approach, right?

patriksvensson: Is Cake.dll in Cake.Tool compiled for 2.1?

patriksvensson: I mean does the Cake.dll have the 2.1 target framework or is the package just 2.1?

mholo65: Yes, it’s Cake.exe/Cake.dll I’m referring to.

devlead: So
* Cake.CoreCLR = netcoreapp2.0
* Cake.Tool = netcoreapp2.1
So why not just a have an define on Cake.Tool and if def, no need for reflection?

gep13:
@mholo65 what is the state of this one?

mholo65:
@gep13 sorry, can't merge this until `Cake.CoreCLR` targets `netcoreapp2.1`.

devlead:
@mholo65 is this now addressed by #2604?

mholo65:
Yes, fixed by #2604

PR #2309, opened 2018-09-22
GH2308: Use Mono on .NET Core for .NET tools

### LINKED ISSUE #2308: Use Mono for full framework executables if runing on Unix & .NET Core
Currently when running Cake on .NET Core on a non Windows OS and trying to execute a tool made for .NET Full Framework will fail with

```diff
- Error: One or more errors occurred. (GitVersion: Process returned an error (exit code 8).)
-       GitVersion: Process returned an error (exit code 8).
```
We should investigate if we could identify this and try to instead execute with Mono.

Example IdentityServer/IdentityServer4#2636

devlead:
* Fixes #2308 

This PR will if running Cake.CoreCLR or Cake.Tool on a non windows, if filename ends with exe and has .NET assembly headers try execute tool using Mono instead.

Example of a script that currently fails
```cake
#tool "nuget:https://api.nuget.org/v3/index.json?package=GitVersion.CommandLine&version=3.6.2"

var gitVersions = Context.GitVersion();

Information(new {
    InformationalVersion = gitVersions.InformationalVersion,
    BranchName = gitVersions.BranchName,
    PreReleaseLabel = gitVersions.PreReleaseLabel
});
```

But with this PR succeeds (if Mono is installed).
image


src/Cake.Core/IO/FileExtensions.cs
```diff
@@ -100,5 +100,49 @@ public static IEnumerable<string> ReadLines(this IFile file, System.Text.Encodin
                 return result;
             }
         }
+
+        /// <summary>
+        /// Checks if file has CLR PE Header.
+        /// </summary>
+        /// <param name="file">The file to be read from.</param>
+        /// <returns><c>true</c> if file is CLR assembly; otherwise, <c>false</c>.</returns>
+        public static bool IsCLRAssembly(this IFile file)
```
gep13: @devlead I don't doubt that this function does exactly what it says, but is it possible to provide a link to some sort of reference as to what exactly is going on here?

src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -52,6 +55,20 @@ public IProcess Start(FilePath filePath, ProcessSettings settings)
...
             var arguments = settings.Arguments ?? new ProcessArgumentBuilder();
 
+#if NETCORE
+            if (_environment.Platform.IsUnix() &&
+                fileName.EndsWith(".exe", StringComparison.OrdinalIgnoreCase) &&
+                _fileSystem.GetFile(fileName).IsCLRAssembly())
+            {
+                if (!settings.Silent)
+                {
+                    _log.Verbose(Verbosity.Diagnostic, "{0} is a .NET Framework executable, will try execute using Mono.", fileName);
```
gep13: If `mono` is not installed on the machine that this is running on, this is obviously going to fail.  Do we need to provide a sensible error message if/when this happens?  Should we first attempt to check to see if mono is installed prior to doing this?

Do we need to provide a configuration value to prevent this "magic" from happening?  i.e. is there a scenario where when running on .NET Core, on Unix with a CLRAssembly, that the end user doesn't want to run with mono?

src/Cake.Core/IO/FileExtensions.cs
devlead: We've been using this method for awhile to aid with tool resolution. I just moved it and made it a public extension method.

Essentially most windows dll:s/exe:s have pe headers, a little can be read link below or in win SDK.

https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format

gep13: @devlead I spotted that after I wrote the comment as I got to the bottom of the changes.  That URL seems very useful.  Can we get it added into the remarks section of the XML comment?

src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -17,16 +17,19 @@ public sealed class ProcessRunner : IProcessRunner
...
+        private readonly IFileSystem _fileSystem;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ProcessRunner" /> class.
         /// </summary>
         /// <param name="environment">The environment.</param>
         /// <param name="log">The log.</param>
-        public ProcessRunner(ICakeEnvironment environment, ICakeLog log)
+        /// <param name="fileSystem">The file system.</param>
+        public ProcessRunner(ICakeEnvironment environment, ICakeLog log, IFileSystem fileSystem)
```
patriksvensson: Can we move the fileSystem argument first here? We put it first everywhere else and this makes my OCD tingle 😄

src/Cake.Core/IO/FileExtensions.cs
patriksvensson: Can we rename this to IsClrAssembly now when it's public?

src/Cake.Core/IO/ProcessRunner.cs
devlead: Basically if ends with `.exe` and has all the identifiers a clr assembly has.

.NET Core has the `.dll` extension, and native shims are extension less on unix (and don't have the needed PE headers)

Could probably bring in the tool resolver to try find mono, warn and not execute it if not found.

Only thing I could think of is WSL which can execute windows binaries from the linux side. 
But that feels like a real non best practice and will require the tool to be installed in path on windows side.

src/Cake.Core/IO/ProcessRunner.cs
devlead: My OCD says put new last, but your OCD is stronger than mine, so I'll address ;)

src/Cake.Core/IO/FileExtensions.cs
devlead: Sure

devlead: sure.

devlead:
> LGTM, but could we also add a CoreCLR tests that tests the new mono logic as well?

@patriksvensson Yip, I'll add that, will create a new fixture for this, the "Given" getting a bit copy/pasty. 

@gep13 @patriksvensson I'll ping back on PR when I've addressed feedback.

src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -52,6 +55,20 @@ public IProcess Start(FilePath filePath, ProcessSettings settings)
             // Get the arguments.
             var arguments = settings.Arguments ?? new ProcessArgumentBuilder();
 
+#if NETCORE
+            if (_environment.Platform.IsUnix() &&
```
devlead: Also check running .NET Core here, as .NET standard in theory could be run under full framework too.

src/Cake.Core/IO/ProcessRunner.cs
gep13: @devlead what were your thoughts around a configuration value to control whether this "magic" happens or not?

devlead: Sure we could add an opt out for this, any suggestions around naming?

devlead: Don't have good imagination today;) So unless anyone comes up with a better name opt-out setting will be `Settings_SkipFullFrameworkMonoFallbackOnUnixCoreClr`.

gep13: A suggestion from @patriksvensson which I think makes sense is `Settings_NoMonoCoersion` and more detail about exactly what it is doing can be added in the documentation.

devlead:
@patriksvensson @gep13 think I've addressed your feedback.

Added 
* `Settings_NoMonoCoersion`
* Check for mono existence 
* Test fixture for ProcessRunner
* Tests for Coersion
* Renamed to IsClrAssembly 
* Added PE reference
* FileSystem OCD first
* Added integration test

So this should be ready for review, could probably add an integration test too. starting Cake.exe from Cake.

devlead:
Oh it's apparently not ;) I'll take a look.

devlead:
@patriksvensson Maybe there's an issue with testing a fake windows file system on a non windows system.
`PathHelper.IsPathRooted(FullPath)` obviously isn't using the FakeEnvironment, but the static EnvironmentHelper.

devlead:
did a quick workaround, so pr should compile and be reviewable.

patriksvensson:
Merged! Thank you for your contribution. Much appreciated! 👍

PR #2316, opened 2018-09-25
GH1875: Enable long path support on Windows 10

### LINKED ISSUE #1875: Folder structure of tools and addins can cause too long paths on Windows
The new folder structure for addins and tools of Cake 0.22.0 and above may result in very long path names. This avoids loading those and causes the following error:

`The specified path, file name, or both are too long. The fully qualified file na
me must be less than 260 characters, and the directory name must be less than 24
8 characters.
NuGet exited with 1
An error occured while installing package SomeCompanyName.SomeOtherName.SomeProductName.BuildScripts.Cake.`

It depends on the length of  the referenced package id which may be out of the script author's control.
The reason for this that the package id is replicated twice in the folder structure. If the package name is already long the path gets far longer now.

My suggestion is to have the package id only once and use short unique id's for the subfolder structure.
E.g.
`D:\Projects\Something\Trunk\packages\Tools\Addins\SomeCompanyName.SomeOtherName.SomeProductName.BuildScripts.Cake.0.1.0.115105\SomeCompanyName.SomeOtherName.SomeProductName.BuildScripts.Cake`

will become 
`D:\Projects\Something\Trunk\packages\Tools\Addins\SomeCompanyName.SomeOtherName.SomeProductName.BuildScripts.Cake.0.1.0.115105\uniqueid1`

Sam13:
As described in https://blogs.msdn.microsoft.com/jeremykuhne/2016/07/30/net-4-6-2-and-long-paths-on-windows-10/ 
Resolves #1875

Enables long path support on Windows 10 when:
- NuGet.exe v4.8.1 and above is used as external NuGet executable
- Long path support is enabled on the machine

Remarks:
- Windows 7 is not supported because does not support long paths
- I tried to upgrade the inprocess NuGet client to v4.8.0 but does not support long path's. So I removed the upgrade from my PR again
- I tried to add long path support to Cake's app.manifest but did not succeed (as described in the link above) - may anyone help me with that? Although I found that information that it may not work (tbc)

### DIFF
--- a/src/Cake/App.config
+++ b/src/Cake/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+  <runtime>
+    <AppContextSwitchOverrides value="Switch.System.IO.UseLegacyPathHandling=false;Switch.System.IO.BlockLongPaths=false" />
+  </runtime>
+</configuration>


devlead:
@Sam13 your changes have been merged, thanks for your contribution 👍

PR #2323, opened 2018-10-02
Main

mwolf1963:
I would like to start deleting obsolete methods and properties


patriksvensson:
@mwolf1963 Was this meant as an issue? In that case you can file one here: https://github.com/cake-build/cake/issues

I'm going to go ahead and close this one for now since it seems to be empty.

Issue #2061, opened 2018-02-27
NuGetPack overwrites `developmentDependency` and `requireLicenseAcceptance` from nuspec.


williamb1024:
### What You Are Seeing?

Builtin NuGetPack always replaces `developmentDependency` and `requireLicenseAcceptance` values when transforming a nuspec file.

Only `developmentDependency` and `requireLicenseAcceptance` operate in this manner.

### What is Expected?

NuGetPack would not replace any setting within the nuspec, unless that setting has been specifically configured within the NuGetPackSettings instance. 

### What version of Cake are you using?
0.25.0+Branch.main.Sha.05b4d3f596defbdf5baecdb3712c9bc17f849b55

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)
Occurs for any call to NuGetPack when provided with a nuspec file with the `developmentDependency` setting configured as `true`.

mattleibow:
Any progress on this? I am getting this and luckily we have a step to verify that this was set.

PR #2330, opened 2018-10-15
GH2061: NuGetPack: Fix the issue when the settings overwrite the .nuspec

mattleibow:
I noticed that there was a bug in the `NuGetPack` alias that would always overwrite the  `developmentDependency` and `requireLicenseAcceptance` values of the .nuspec. This probably has to do with the fact that the setting is `bool` and this always has a default value.

See this issue: #2061

To fix this, I changed it to `bool?` wich has a default of `null`, which is handled correctly.

BUT, this may be a BREAKING change since `bool` -> `bool?` just fine, but not the other way around. As a result, this may need more thought. But, since we are cool, and with it, I think this is a bug since I just discovered that my values were overwritten _and almost went out save for our automated checks we have at the door_.

~_Apologies for not submitting the PR according to the guidelines because I was doing this on the go via the online editor. If this turns out to be more that the two changes, I will see if I can get this according to standards when I get the opportunity - or maybe someone has a fix already._~

### DIFF
--- a/src/Cake.Common.Tests/Properties/Resources.resx
+++ b/src/Cake.Common.Tests/Properties/Resources.resx
@@ -899,8 +899,6 @@ EndGlobal</value>
     &lt;version&gt;1.0.0&lt;/version&gt;
     &lt;authors&gt;Author #1,Author #2&lt;/authors&gt;
     &lt;description&gt;The description&lt;/description&gt;
-    &lt;developmentDependency&gt;false&lt;/developmentDependency&gt;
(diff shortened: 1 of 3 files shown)


mholo65:
@mattleibow thank you for your PR.

@cake-build/cake-team I'm ok with changing `bool` to `bool?`. How about you?

devlead:
@mholo65 I'm ok with it.

patriksvensson:
Also OK with me!

devlead:
@mattleibow your changes have been merged, thanks for your contribution 👍

PR #2355, opened 2018-11-12
GH2384: Add fallback path for mono msbuild if installed under homebrew on MacOS

austinlparker:
This PR adds a fallback path to `/usr/local/bin/msbuild` in the MSBuild detector. If `mono` is installed via homebrew, `msbuild` is symlinked to this location by convention and is not located at the default `/Library/Frameworks...` path.


devlead:
@austinlparker your changes have been merged, thanks for your contribution 👍

PR #2360, opened 2018-11-18
add missing report types

### LINKED ISSUE #2327: Missing report types for ReportGenerator
### What You Are Seeing?
The enumeration [`ReportGeneratorReportType`](https://github.com/cake-build/cake/blob/main/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs#L10) doesn't contain all the report types supported by [ReportGenerator](https://danielpalme.github.io/ReportGenerator/usage.html) ([doc here also](https://github.com/danielpalme/ReportGenerator/wiki/Output-formats)).

https://github.com/cake-build/cake/blob/cab58db40b02346e562531be27fcaa5ea51a3312/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs#L10-L51

### What is Expected?
The enumeration should provide an element per report type supported by ReportGenerator. The following types are missing:

* HTMLInline
* HTMLChart
* MHTML
* Cobertura
* CsvSummary

Check code from `master` branch.

tpluscode:
I added the missing report types to ReportGenerator addin, as seen on the tool's [docs page](https://danielpalme.github.io/ReportGenerator/usage.html)

Fixes #2327


src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs
```diff
@@ -14,25 +14,60 @@ public enum ReportGeneratorReportType
...
+        /// <summary>
+        /// CSV summary report
+        /// </summary>
+        CsvSummary,
+
         /// <summary>
         /// Html report.
         /// </summary>
-        Html = 2,
+        Html,
```
tpluscode: You will notice that I removed the explicit enum numbers.

Would you like me to:
1. either, add them explicitly in alphabetical order
2. or, add the new ad the bottom and preserve the original numbering
3. or, keep as it is now?

Second option probably doesn't make sense as I don't think that any external code would explicitly rely on those specific values, but you never know...

mholo65: Since this is a public enum, anything but option 2 is a breaking change. Hence, I’d go with that option.

PR #2364, opened 2018-11-25
GH2146: Enabled initializer syntax for all collection properties (for real this time)

### LINKED ISSUE #2146: Enabling initializer syntax for all collection properties

jnm2:
Actually fixes #2146. Missed a few (27) in https://github.com/cake-build/cake/pull/2147. I thought something was missing at the time. 🤦‍♂️

### DIFF
--- a/src/Cake.Testing/Extensions/ReflectionExtensions.cs
+++ b/src/Cake.Testing/Extensions/ReflectionExtensions.cs
@@ -28,7 +28,7 @@ namespace Cake.Testing.Extensions
         /// <returns>
         /// <see langword="true"/> if  the specified type implements the specified interface definition; otherwise, <see langword="false"/>.
         /// </returns>
-        public static bool ImplementsInterfaceDefinition(this Type type, Type interfaceTypeDefinition)
+        public static bool SatisfiesInterfaceDefinition(this Type type, Type interfaceTypeDefinition)
         {
             if (type == null)
             {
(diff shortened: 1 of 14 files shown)


src/Cake.Testing/Extensions/ReflectionExtensions.cs
```diff
@@ -45,19 +45,29 @@ public static bool ImplementsInterfaceDefinition(this Type type, Type interfaceT
...
 
-            return type.GetInterfaces().Contains(interfaceTypeDefinition);
+            if (interfaceTypeDefinition.IsConstructedGenericType)
+            {
+                throw new ArgumentException("An interface type definition must be specified.", nameof(interfaceTypeDefinition));
+            }
+
+            if (type.IsGenericType && type.GetGenericTypeDefinition() == interfaceTypeDefinition)
+            {
+                return true;
```
jnm2: These are the ones we were actually missing, where the type *is* the interface rather than *implementing* the interface.

jnm2:
@patriksvensson Oops, fixed!

patriksvensson:
@jnm2 Sorry for not getting back to you on this. Could you rebase against latest develop and I'll merge asap. Thanks!

jnm2:
@patriksvensson Done.

jnm2:
@patriksvensson Since there are no merge conflicts, I believe you can merge without rebasing. That's what I usually do.

devlead:
@jnm2  your changes have been merged, thanks for your contribution 👍

jnm2:
@devlead I appreciate that, thanks for getting it in!

May I ask that future PRs not be rebased (or squashed) before merging? There's generally not a technical need to do so. Both operations obscure the context in which I wrote the code and rewrite the commit metadata.

devlead:
@jnm2 we've opted to use protected branches and not allow people to do continuous upstream merge commits. Which makes it an requirement, it also ensures the PR has built against latest bits, even if there's no merge conflicts, bootstrappers, SDKs, framework targets, tools, unit tests, etc. might have changed - resulting in failed builds we rather catch in PR before merging.

In this case I did not squash commits, just replayed them upon develop, so everything except hashes remained intact. In general we don't want too many commits per PR. The majority of PRs can be focused and limited to just one or two commits.

PR #2366, opened 2018-11-25
 Use proper casing for brand names

jnm2:
Saw some while doing the other PR, went down another rabbit hole. :D

### DIFF
--- a/src/Cake.Common/Tools/MSpec/MspecSettings.cs
+++ b/src/Cake.Common/Tools/MSpec/MspecSettings.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Cake.Core.IO;
 using Cake.Core.Tooling;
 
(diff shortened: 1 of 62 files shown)


src/Cake.Common/Tools/MSpec/MspecSettings.cs
```diff
@@ -94,12 +94,12 @@ public sealed class MSpecSettings : ToolSettings
...
         ///     Gets or sets a value indicating whether to enable reporting for AppVeyor CI integration(also auto - detected)
         /// </summary>
-        public bool Appveyor { get; set; }
+        public bool AppVeyor { get; set; }
 
         /// <summary>
         ///     Gets or sets a value indicating whether disable AppVeyor autodetection
         /// </summary>
-        public bool NoAppveyor { get; set; }
+        public bool NoAppVeyor { get; set; }
```
jnm2: Would this pair of breaking changes be accepted or should I drop them?

patriksvensson: @jnm2 Not sure. I suspect this will break a lot of things for a lot of people.

@cake-build/cake-team What do you think?

devlead: Could obsolete with warning and have it backed by correct casing property and remove it next release.

patriksvensson: @devlead Yes, that's a great solution 👍

jnm2:
Note to self: do `CoreClr` -> `CoreCLR` as well.

src/Cake.Common/Tools/MSpec/MspecSettings.cs
jnm2: Sounds good to me! Would you want to do anything like that for Convertable\*Path -> Convertible\*Path, which are entire classes? For compatibility, we'd need to have obsoleted classes with the existing names inheriting from new classes with proper spelling and all the functionality.

jnm2: I didn't hear back about renaming the classes from `Convertable` to `Convertible`, so I will leave them alone.

src/Cake.Common/Tools/MSpec/MspecSettings.cs
```diff
@@ -92,14 +93,28 @@ public sealed class MSpecSettings : ToolSettings
...
+        public bool AppVeyor { get; set; }
 
         /// <summary>
-        ///     Gets or sets a value indicating whether disable AppVeyor autodetection
+        ///     Gets or sets a value indicating whether to enable reporting for AppVeyor CI integration (also auto-detected).
         /// </summary>
-        public bool NoAppveyor { get; set; }
+        [Obsolete("Please use the AppVeyor property instead (note the capitalization). This property will be removed in a future version.")]
+        [CLSCompliant(false)] // Identifier differs only in case
+        public bool Appveyor { get => AppVeyor; set => AppVeyor = value; }
```
jnm2: @patriksvensson @devlead Just wanted to make sure you saw this. It looks like you have `[assembly: CLSCompliant(true)]`. Cake is C# Make, not VB Make, so I'm going to assume that CLS compliance isn't top priority for this particular property, but if that's not the case I can revert.

jnm2: So, is that important for this particular property? Should I test the behavior in a non-case-sensitive language, or just revert?

jnm2:
@devlead Is the stale `Waiting for issuer feedback` label causing my PR to be overlooked?

mholo65:
Let’s do `Convertable*Path` -> `Convertible*Path` in a separate PR!

jnm2:
> @jnm2 could you rebase against latest develop.

👍 done.

devlead:

PR #2380, opened 2018-12-08
Add an Encoding parameter to TextTransformation<TTemplate>.Save

### LINKED ISSUE #2379: Add an Encoding parameter to TextTransformation<TTemplate>.Save
Adding an [Encoding](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=netstandard-2.0) parameter to [TextTransformation\<TTemplate\>.Save](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Text/TextTransformation.cs#L54) would make it more flexible. It could default to `null` for backwards compatibility and match existing behavior.

Happy to submit a PR.

gitfool:
Fixes #2379. Note: I also refactored the testing in this area:

- Added FakeFileExtensions.GetBinaryContent
- Added FakeFileExtensions.HasUTF8BOM
- Added ByteArrayExtensions.StartsWith
- Changed XmlPokeFixture to use FakeFileExtensions


src/Cake.Common/Text/TextTransformation.cs
```diff
@@ -49,9 +49,10 @@ public sealed class TextTransformation<TTemplate>
         /// <summary>
         /// Saves the text transformation to a file.
         /// </summary>
-        /// <param name="path">The <see cref="FilePath"/> to save the test transformation to.</param>
+        /// <param name="path">The <see cref="FilePath"/> to save the text transformation to.</param>
+        /// <param name="encoding">The text encoding. Defaults to UTF8 without BOM.</param>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times", Justification = "Stream writer leaves stream open.")]
-        public void Save(FilePath path)
+        public void Save(FilePath path, Encoding encoding = null)
```
devlead: To not break ABI add an `public void Save(FilePath path)` overload calling the new one instead of default parameter.

src/Cake.Testing/Extensions/FakeFileExtensions.cs
```diff
@@ -44,12 +46,38 @@ public static FakeFile SetContent(this FakeFile file, string content)
...
+
+        /// <summary>
+        /// Gets the text content of the file.
+        /// </summary>
+        /// <param name="file">The file.</param>
+        /// <param name="encoding">The text encoding. Defaults to UTF8 without BOM.</param>
         /// <returns>The text content of the file.</returns>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times")]
-        public static string GetTextContent(this FakeFile file)
+        public static string GetTextContent(this FakeFile file, Encoding encoding = null)
```
devlead: To not break ABI add an `GetTextContent(this FakeFile file)` overload calling the new one instead of default parameter.

gitfool:
@devlead Fixed backwards compatibility. Rebased and squashed.

Issue #1283, opened 2016-10-13
Empty folders are not included in a ZIP archive


vladimir-kazakov:
### What You Are Seeing?
I'm trying to create a ZIP archive with pre-defined folder structure. Some folders are empty. After using `Zip("./Package", "Package.zip");`, I see that Package.zip is created, but when I look inside it, there are no empty folders. Only folders with files are included.

### What is Expected?
I expect empty folders to be present in the ZIP archive.

### What version of Cake are you using?
0.16.2.

### Are you running on a 32 or 64 bit system?
64-bit.

### What environment are you running on?  Windows? Linux? Mac?
* Windows 10 (1511).
* Ubuntu 16.04.1.

### Are you running on a CI Server?  If so, which one?
No.

### How Did You Get This To Happen? (Steps to Reproduce)
Create the following folder structure:
```
+ Package
|  + Empty
|  + NotEmpty
|  |  - File.txt
```
In the `build.cake` perform `Zip("./Package", "Package.zip");`.

### Output Log
~~~sh
Feeds used:
  /home/havok/.local/share/NuGet/Cache
  /home/havok/.nuget/packages/
  https://api.nuget.org/v3/index.json

All packages listed in /home/havok/Projects/Cake/ZipEmptyFolders/tools/packages.config are already installed.
Module directory does not exist.
Analyzing build script...
Analyzing /home/havok/Projects/Cake/ZipEmptyFolders/build.cake...
Processing build script...
Creating script session...
Adding reference to Cake.Core.dll...
Adding reference to Cake.Common.dll...
Adding reference to Cake.exe...
Adding reference to System.Xml.dll...
Adding reference to System.Xml.Linq.dll...
Adding reference to System.Data.dll...
Importing namespace Cake.Common...
Importing namespace Cake.Common.Build...
Importing namespace Cake.Common.Build.AppVeyor...
Importing namespace Cake.Common.Build.AppVeyor.Data...
Importing namespace Cake.Common.Build.BitbucketPipelines...
Importing namespace Cake.Common.Build.BitbucketPipelines.Data...
Importing namespace Cake.Common.Build.Bitrise...
Importing namespace Cake.Common.Build.Bitrise.Data...
Importing namespace Cake.Common.Build.ContinuaCI...
Importing namespace Cake.Common.Build.ContinuaCI.Data...
Importing namespace Cake.Common.Build.Jenkins...
Importing namespace Cake.Common.Build.Jenkins.Data...
[...]

PR #2382, opened 2018-12-09
GH2381: Implement standard zip behavior by default

### LINKED ISSUE #2381: Zip should behave by default like standard Zip utilities
Specifically, the following subtle but very important behaviors are expected:
- Directories should be stored as zip entries
- Empty directories should be included

For example, consider the following directory structure:
- Root/
  - Root.txt
  - Empty/
  - First/
    - First.txt
    - Nested/
      - Nested.txt

When archived by Cake Zip it looks like the following (using the flat view in 7-Zip):

image

When archived by Windows built-in zip, [7-Zip](https://www.7-zip.org/), [WinZip](https://www.winzip.com/), and [WinRAR](https://win-rar.com/), it looks like the following:

image

Only the _file_ compression method differs between the various utilities; some will `Store` them rather then `Deflate` them, depending on compression level. Note that all directories are stored, including empty directories.

As trivial as it sounds, I really need this behavior, without needing another dependency. PR incoming! 😉

gitfool:
Fixes #2381 and #1283. Note: I also found and fixed some other issues while testing.
- Added `PathCollection` (mirroring `DirectoryPathCollection` and `FilePathCollection`)
- Added `GlobbingAliases.GetPaths`
- Changed `ZipAliases.Zip` to use `GlobbingAliases.GetPaths`
  - For zipping a _directory_ only; not used by all other _file_ centric aliases
- Default `PathCollection.Comparer` to `PathComparer.Default`
- Add support for directory and file timestamps
- Refactor `ZipperTests` to use `CakeContextFixture`
  - Exercise `GlobbingAliases.GetPaths`
  - Fixed `FakeFileSystem` to work with simple root directory (`/`)
    - `GlobVisitor.VisitRecursiveWildcardSegment`
    - `Zipper.GetRelativePath`
    - `FakeFileSystemTree.ctor`
  - Verify zip archive structure, including stored directories
  - ... and stored empty directories (when zipping a _directory_ only)


gitfool:
@devlead do the integration tests run as part of the automated build? I don't see them in the Azure Pipelines logs. I ran them manually and following is the zip archive created by the new integration test:

image

PS. Should I rebase my branch with `develop`?

src/Cake.Core/IO/IFileSystemInfo.cs
```diff
@@ -30,5 +32,11 @@ public interface IFileSystemInfo
         ///   <c>true</c> if the entry is hidden; otherwise, <c>false</c>.
         /// </value>
         bool Hidden { get; }
+
+        /// <summary>
+        /// Gets the time when this <see cref="IFileSystemInfo"/> was last written to.
+        /// </summary>
+        /// <value>The last write time.</value>
+        DateTime LastWriteTime { get; }
```
devlead: @cake-build/cake-team please review breaking change.

src/Cake.Core/IO/FilePathCollection.cs
```diff
@@ -27,8 +27,8 @@ public sealed class FilePathCollection : IEnumerable<FilePath>
         /// <summary>
         /// Initializes a new instance of the <see cref="FilePathCollection"/> class.
         /// </summary>
-        /// <param name="comparer">The comparer.</param>
-        public FilePathCollection(PathComparer comparer)
+        /// <param name="comparer">The comparer. Defaults to <see cref="P:PathComparer.Default" />.</param>
+        public FilePathCollection(PathComparer comparer = null)
```
devlead: @cake-build/cake-team please review potential breaking change.

src/Cake.Core/IO/DirectoryPathCollection.cs
```diff
@@ -27,8 +27,8 @@ public sealed class DirectoryPathCollection : IEnumerable<DirectoryPath>
         /// <summary>
         /// Initializes a new instance of the <see cref="DirectoryPathCollection"/> class.
         /// </summary>
-        /// <param name="comparer">The comparer.</param>
-        public DirectoryPathCollection(PathComparer comparer)
+        /// <param name="comparer">The comparer. Defaults to <see cref="P:PathComparer.Default" />.</param>
+        public DirectoryPathCollection(PathComparer comparer = null)
```
devlead: @cake-build/cake-team please review potential breaking change.

src/Cake.Core/IO/IFileSystemInfo.cs
mholo65: Is there any way we could get LastWriteTime without breaking?

src/Cake.Core/IO/DirectoryPathCollection.cs
gitfool: When I added `GlobbingAliases.GetPaths` I saw that there was some duplication in the existing code that effectively initialized the `comparer` parameter to `PathComparer.Default` in:
- GlobbingAliases.GetPaths
- GlobbingAliases.GetDirectories
- GlobbingAliases.GetFiles
- DirectoryAliases.GetSubDirectories
- VSWhereTool.RunVSWhere

I thought it would be better to move this default initialization from the call site to the various *`PathCollection` based constructors. I did this using a default parameter, which should be backwards compatible, unless I'm missing something, but it could be also be done with another default ctor overload that calls the existing ctor (and dropping the default parameters). Technically that would be nicer as it simplifies all the call sites. Is there a preferred approach in this repository?

Happy to change it. 😃

src/Cake.Core/IO/FilePathCollection.cs
gitfool: See my response below.

src/Cake.Core/IO/IFileSystemInfo.cs
gitfool: Yeah I can't see a way to do this without a breaking change; perhaps C# 8 would help in future.

FWIW, all the existing implementations (`Directory`, `File`, `FakeDirectory` and `FakeFile`) are `sealed` and I implemented those.

Having access to directory and file timestamps is really useful so could we just push it through and manage it like previous breaking changes?

src/Cake.Core/IO/FilePathCollection.cs
devlead: This will retain API compatibility, but will break  ABI which means any addin or module utilizing this will throw an runtime error. Only non breaking way to add it would be to add a new parameterless contructor that forwards null to the existing construction.

src/Cake.Core/IO/DirectoryPathCollection.cs
devlead: This will retain API compatibility, but will break  ABI which means any addin or module utilizing this will throw an runtime error. Only non breaking way to add it would be to add a new parameterless contructor that forwards null to the existing construction.

gitfool: @devlead so would you accept the change if I refactor to add a default ctor and drop the default parameter?

src/Cake.Core/IO/IFileSystemInfo.cs
devlead: We're in general restrictive with breaking changes, and always try to weigh benefit vs. risk, and if possible see if we can avoid. Breaking changes means all existing addins/modules/assemblies will need to be updated, recompiled and published. Sometimes reward justifies it, but it also means we'll review and discuss it more before merging. 

We've added he notion of possible breaking change, when we think it's a low risk, we will warn but still execute, if we know it's a high impact breaking change we'll not even load modules/addins targeting previous version.

On the module side personally I only know of one that's implementing IFileSystemInfo / IFile / IDirectorym so might be a low risk but we would need to test and inventory that to deem low risk.

src/Cake.Core/IO/DirectoryPathCollection.cs
devlead: Probably, would need to review and test more to be certain.

gitfool:
Looks like `Build Cake Ubuntu` failed with an unrelated error?

src/Cake.Core/IO/IFileSystemInfo.cs
gitfool: Okay, it sounds like it might be low risk as it will likely only affect modules. Is there a list of publicly available modules somewhere? How can I help?

gitfool: Meanwhile, I've just realised there's a more elegant solution: remove the breaking change to `IFileSystemInfo`, keep my changes to the existing implementations (`Directory`, `File`, `FakeDirectory` and `FakeFile`), and downcast types for timestamps in `Zipper`.

gitfool: See commit [3039e28c](https://github.com/cake-build/cake/pull/2382/commits/3039e28c6c5f8b9eb6e923c6d543bbaec1e041e4).

The Azure Pipelines build for `Build Cake Windows` had an unrelated failure. Is there a way to retry those builds without another commit?

devlead: I've queued a new build.

devlead: GH status not reflecting it but build succeded
https://dev.azure.com/cake-build/Cake/_build/results?buildId=2690

gitfool:
I've got one more tweak I want to do: I was being conservative when I initially added `GlobbingAliases.GetPaths` with `internal` access only, but since then I added a lot of unit tests, so I think it would be nice to make `public`.

Besides that, @devlead @mholo65 do you want me to do anything else to keep moving this PR forward?

devlead:
@gitfool Rebase against latest deveop and squash commits, and I'll review again.

gitfool:
@devlead rebased and squashed; ready for review.

devlead:

PR #2383, opened 2018-12-09
Add possibility to filter results of the nugetlist command

bqstony:
When i list the available packages form a Azure DevOps Nuget feed, is actualy the problem, the packageid in the command to filter for, is not enouth. 

The example shows the Problem:
```
   var packageList = NuGetList("mypackage", new NuGetListSettings()
   {
      Prerelease = true,
      AllVersions = true,
      Source = new []  { "MyDevOpsPackageFeed" },
   });
```

This returns a List like:
- MyPackage 1.0.0
- MyPackage 1.0.1
- MyPackage 1.1.0-alpha
- MyPackage.Utils 2.0.0
- MyPackage.Utils 1.0.0
- MyPackage.Analyzer 1.0.0

What i relay wont was 
- MyPackage 1.0.0
- MyPackage 1.0.1
- MyPackage 1.1.0-alpha

In this PR i have add the possibility:
- to filter of a equals match of packageId
- to filter for a specific version
- to filter of own creaded predicates

I think this is a way to fill the hole of the missing cli tool functionality.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/NuGet/List/NuGetListFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/NuGet/List/NuGetListFixture.cs
@@ -24,6 +24,20 @@ namespace Cake.Common.Tests.Fixtures.Tools.NuGet.List
             });
         }
 
+        public void GivenMultibleVersionsInPackageResult()
+        {
+            ProcessRunner.Process.SetStandardOutput(new string[]
+            {
+                "Cake 0.31.0-alpha0075",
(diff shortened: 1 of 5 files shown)


mholo65:
@bqstony thank you for the PR. We'd rather see that you just filter on the [`IEnumerable<NuGetListItem>`](https://cakebuild.net/api/Cake.Common.Tools.NuGet.List/NuGetListItem/) that the  `NuGetList()`  alias returns instead of implementing the filter logic in Cake itself. Thank you!

Issue #2393, opened 2018-12-13
MethodAliasGenerator doesn't generate parameter attributes


devlead:
Currently code gen omits parameter attributes, this means i.e. the `GetCallerInfo` generates wrong results when called without a context.

This source code

https://github.com/cake-build/cake/blob/730bf9b83e794583fdd3b93677c9dc0ae9fcbee8/src/Cake.Common/Diagnostics/ScriptCallerAliases.cs#L23-L28

becomes
```csharp
ScriptCallerInfo GetCallerInfo(
            string memberName = "",
            string sourceFilePath = "",
            int sourceLineNumber = 0)
```

Reference code:
 * https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Scripting/CodeGen/MethodAliasGenerator.cs

devlead:
@kcamp pefect, assigned issue to you.

devlead:
Fixed by #2402

PR #2402, opened 2018-12-19
 GH2393: Support attribute-decorated parameters in cake alias methods for codegen

kcamp:
This addresses the shortcomings in `ParameterEmitter.cs` that were identified in #2393.
I have left this branch as two separate commits for review.

The first commit (bf50ce3) provides an exact fix to the issue surfaced by `ScriptCallerAliases` and provides matching test coverage.  It's very possible that this is the extent of the fix required or desired, and in that case, we can modify the PR to reflect that.

The second commit took the starting logic from the first and attempted to craft a more complete solution, supporting a wider range of possible parameter decorations to include both normal constructor mechanics as well as named arguments.  Based on the test cases that were implemented, I think this provides a reasonable implementation (779acfa), but obviously there are many, many test cases that could be authored to tease that out with a bit more certainty.  If that is the desire, I would be happy to continue on with that work, but it seems that fixing an already-released feature to be semantically correct makes sense as a stopgap.

Let me know how you'd like to proceed.

### DIFF
--- a/src/Cake.Core.Tests/Data/MethodAliasGeneratorData.cs
+++ b/src/Cake.Core.Tests/Data/MethodAliasGeneratorData.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq.Expressions;
+using System.Runtime.CompilerServices;
 using Cake.Core.Annotations;
 
 namespace Cake.Core.Tests.Data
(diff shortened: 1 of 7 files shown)


kcamp:
@devlead If the team is in a position to put eyes on this for review, I should be able to address any feedback this weekend/early next week.  Thanks.

kcamp:
just a nudge; if there's any adjustments to be made, let me know.

devlead:
@kcamp haven't forgotten you just need to verify and take this for a spin. Would want a integration test i.e. using [GetCallerInfo()](https://cakebuild.net/api/Cake.Common.Diagnostics/ScriptCallerAliases/D464584F) alias.

devlead:

PR #2408, opened 2018-12-21
(build) Update addins&tools,refactor bootstrappers

devlead:
This is still work in progress, mostly to see how far we can go with the Global tool now that it has Mono fall back.
Using same runtime to execute script everywhere simplifies things.

- [x] Switch to use Cake .NET Core Global tool runner
- [x] Rewritten bootstrappers 
    - [x] Reduce Cake & .NET Core SDK verson from 3 to 2 places (global.json, build.ps1, build.sh -> global.json, build.config)
    - [x] Use Cake standard command line parameters (build.ps1 / build.sh called the same way)
    - [x] Don't install Cake if same version is installed globally
    - [x] Don't install .NET Core if same version installed globally
    - [x] Update to .NET Core SDK 2.1.502
- [x] Update tools
- [x] Update addins
- [x] Switch to .NET Core Global tools where available (Botstrap with Cake.DotNetTool.Module)
- [x] [Bitbucket Pipelines](https://bitbucket.org/azurevoodoo/cake/addon/pipelines/home#!/results/5)
- [x] [Travis CI Linux](https://travis-ci.org/devlead/cake/jobs/470754427)
- [x] [Travis CI Mac](https://travis-ci.org/devlead/cake/jobs/470754426)
- [x] [GitLab Linux](https://gitlab.com/cake-build/cake/-/jobs/137508772)
- [x] [AppVeyor Windows](https://ci.appveyor.com/project/devlead/cake/builds/21161812)
- [x] Azure Pipelines
    - ~~Switch to yaml based build~~
    - [x] Utilize bootstrappers
- [x] Bitrise
    - [x] Utilize bootstrappers
- Move to Cake official containers
    - [x] Bitpucket Pipelines
    - [x] GitLab
    - [x] TravisCI
    - ~~Azure Pipelines~~
        - [x] Debian
    - [x] Bitrise
- Create official containers for
    - [x] Debian
    - ~~Fedora~~ - cake-build/docker#4
    - ~~Centos~~ - cake-build/docker#5
    - ~~Ubuntu~~ - cake-build/docker#6
- [x] Refactor and simplify integration tests bootstrappers

Azure Pipelines and Bitrise will likely fail as they currently don't use the bootstrappers.


gitfool:
@devlead If I may add a suggestion, I would add to your list to ensure that `build.ps1` works with [PowerShell Core](https://github.com/PowerShell/PowerShell) and so runs everywhere - Windows, Mac, Linux - without changes.

build.ps1
```diff
@@ -93,54 +84,40 @@ if($FoundDotNetCliVersion -ne $DotNetVersion) {
...
-}.GetEnumerator() | %{"--{0}=`"{1}`"" -f $_.key, $_.value };
-
-# Start Cake
-Write-Host "Running build script..."
-Invoke-Expression "& `"$CakePath`" `"build.cake`" $Arguments $ScriptArgs"
-exit $LASTEXITCODE
+& "$CakeExePath" ./build.cake --bootstrap
+if ($LASTEXITCODE -eq 0)
+{
+    & "$CakeExePath" ./build.cake $args
```
kll: I think for this to work as intended you need to [splat](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-5.1&viewFallbackFrom=powershell-Microsoft.PowerShell.Core) the args variable.

```suggestion
    & "$CakeExePath" ./build.cake @args
```

devlead: Not at computer now, so can't test, but pretty sure `$args` works with native executables as all are passed unparsed.

Won't splat try match to match arguments against named PowerShell cmd:let arguments?

kll: I absolutely do not understand the finicky inner workings of PowerShell, but at least in my extremely similar bootstrapping script I'm working on passing it as `$args` did not work and it was like no arguments were passed at all. For example doing something like `./build.ps1 --verbosity=verbose` did not result in verbose output but did when I switched to it use `@args` instead.

devlead: Ok, I'll give it a test run when back at computer

Looking at CI builds

`.\build.ps1 --target="AppVeyor"` seems to have worked at least.

kll: It was actually a bad example that I gave. It seems to work with exactly one parameter. I was using multiple and the first one was `-verbose` in an attempt to make the bootstrapping script also run verbosely. That resulted in no errors but no parameters effecting the cake script. When I take that out it works with a single `--parameter=value` type parameter but fails when using multiple. So a better example to try is this something like this:

```shell
./build.ps1 --verbosity=verbose --target=clean
```

That will result in the error:

> The value 'verbose --target=clean' is not a valid verbosity.

So it seems like it might be doing some kind of automatic quoting and passing them all as a single parameter (which is probably why it works with only one parameter). Maybe it's a subtle difference in behavior from removing the use of `Invoke-Expression`?

devlead: Interesting both $args and @args work for me with multiple args, when I tried

```
cmd /c echo @args
cmd /c echo $args
```

Called `.\build.ps1 --target=Clean --verbosity=Diagnostic --test="jjkkjh    ddd"`

Output was very similar
```
--target=Clean --verbosity=Diagnostic "--test=jjkkjh    ddd"
--target=Clean --verbosity=Diagnostic "--test=jjkkjh    ddd"
```

Interesting though that both move the quotes, but it still work in the cake command line parser.

Maybe you had extra quotes?

kll: I looked at it again and figured out what the real difference was between our scripts. I was calling cake from within a PowerShell function and passing the script's Args into the function and then on to the `&` call operator. Something in that messed up the arguments and splatting just happened to fix them.

Sorry for the distraction. I was writing a bootstrapper and trying to base it on the latest and greatest version from this PR and hit an issue with the parameters that took me a while to get working and was just trying to share what I learned... I just didn't actually learn what I thought I learned and wasted your time instead. 🤦‍♂️

devlead: No worries 😎

gitfool:
@devlead Cool. Which file is the "upstream obtain script"? My ideal would be _everything_ works from PowerShell Core without needing any Bash scripts at all - one less file to check-in and can be used consistently everywhere.

devlead:
@gitfool there will only need to be one build.ps1 file committed, but to download .NET Core if missing we download the official obtain PowerShell script and that currently isn't PowerShell core compatible, but there's a bash script we can fallback to on Unix systems.

Example of a boostrapper I've done this previously
https://github.com/cake-contrib/Cake.Kudu.Client/blob/8f45387cc4208cfbf8d2367c04cf6210159b2801/tests/integration/build.ps1#L45-L82

Issue #2420, opened 2018-12-30
Add new label alias for GitReleaseManager


gep13:
With the release of GitReleaseManager 0.8.0, a new label verb has been created which allows the creation of standard labels for GitHub issues.

A new alias needs to be created within Cake.Common to exercise this verb.

PR #2422, opened 2018-12-30
GH2419: Add token parameter to GitReleaseManager

### LINKED ISSUE #2419: Extend GitReleaseManager aliases to use token parameter
With the release of GitReleaseManager 0.8.0 a new token parameter has been added.  This can be used instead of the username/password combination.

Additional aliases will have to be created for each function of GitReleaseManager to allow this to work.

gep13:
This builds on #2420 and will need to rebased once that is merged.

Fixes #2419


devlead:

Issue #2354, opened 2018-11-09
NuGetPush alias should capture NuGet.exe output


hienng:
When pushing with NuGetPush and expecting something like a failure because of a disabled package override on the NuGet Feed.

The expected NuGet.exe output is: 
`Response status code does not indicate success: 403 (Package Already Exists)`

However it is not captured and instead the actual return in OnError(exception => { ... }); is:
`NuGet: Process returned an error (exit code 1)`

Relevant Gitters:
[November 8, 2018 8:10 AM](https://gitter.im/cake-build/cake?at=5be3e1532138723405d2cbf5)
[November 8, 2018 9:53 AM](https://gitter.im/cake-build/cake?at=5be3f98092b7d1172156dfb8)
[November 8, 2018 11:20 AM](https://gitter.im/cake-build/cake?at=5be40e07d001b917202969c9)

fseegraeber:
After looking at different tool aliases, this seems to be the current behavior of most (all?) tools.
https://github.com/cake-build/cake/blob/c39e42433275f6cbc39d565e6a3bc341fc07aec4/src/Cake.Core/Tooling/Tool.cs#L133-L141
If the exit code is not zero you'll get this generic error exception.

To get a hold of the standard output or error you'd need to run NuGet from a `ProcessRunner` and set `ProcessSettings.RedirectStandardOutput`/`RedirectStandardError` afaik. But that defeats the whole purpose of aliases, I guess.

It would probably be nice to get the standard error output in the exception message of a tool process that exits with != 0. I'm not too sure about a clean way to do that though. We could default to redirecting the the standard error and pass that to the `ProcessExitCode` method.

devlead:
There's a few issues with always redirecting output, you would lose console log unless you proxy it, if you proxy it you would lose colors in output. Some aliases already redirect and uses output so you would need to take that into consideration too.

One could potentially do a proxy IProcess as a module to test that outside Cake.Core.

Perhaps it could be optin via toolsettings and perhaps it's enough to proxy stderror and use that as message if avail.

tapika:
Please note that `class NuGetPusher` is sealed, so end-user cannot even override it.
Also not so clear on where anykind of error handling should be located (`OnError` does not contains sufficient information), suspect it's located inside `ProcessRunner > ProcessWrapper`, which is intermediate class contructed on the way.

It would be nice to report - if you have authentication failure, please contact this and that guy or configure nuget like written in wiki pages.

Issue #1485, opened 2017-01-30
Enable correlation of stdout and stderr in ProcessRunner


jnm2:
The way `ProcessRunner` is implemented right now, redirected stdout and stderror cannot be correlated. The output is [buffered into two different queues](https://github.com/cake-build/cake/blob/v0.17.0/src/Cake.Core/IO/ProcessRunner.cs#L113-L140).

The API is designed such that you are forced to consume each collection of lines separately. [`IProcess.GetStandardOutput()` and `GetStandardError()`](https://github.com/cake-build/cake/blob/v0.17.0/src/Cake.Core/IO/IProcess.cs#L33-L43)

This means that if any script or module needs to redirect them, there is no way to piece back together the order of events:
```
[stdout] Executing test A
[stderr] Error message
[stdout] Executing test B
[stderr] Error message
[stdout] Executing test C
```
You will have to choose between showing all error output first or all last:
```
[stdout] Executing test A
[stdout] Executing test B
[stdout] Executing test C
[stderr] Error message
[stderr] Error message
```

This is a particular blocker for #156, allowing parallel execution of tasks. My [output buffering prototype](https://github.com/cake-build/cake/issues/156#issuecomment-272663967) works well except that it's impossible to buffer the correct order of error and output lines without me duplicating and fixing the implementation of `ProcessRunner`.

**The important part** though is that this is an issue you'll want to deal with at some point for sure. You'll have to use something like my buffering prototype to make #156 useful. Also, this affects ordinary modules, some of which need to redirect standard output and errors. The sooner this can be resolved the better.

---

I propose changing [`ProcessRunner`](https://github.com/cake-build/cake/blob/v0.17.0/src/Cake.Core/IO/ProcessRunner.cs#L113-L140) to put both stdout and stderr lines in the same queue with an extra bit to indicate which are stdout and which are stderr.

I propose exposing an `IProcess.GetRedirectedLines()` which returns an `IEnumerable<RedirectedLine>` with `struct RedirectedLine` containing `string Line` and `StreamType Type` with `enum StreamType { StdOut, StdErr }`.

[...]

PR #2447, opened 2019-01-12
GH2270: Added RedirectedStandardOutputHandler and RedirectedStandar…

### LINKED ISSUE #2270: Allow to listen and modify redirected standard output of a process
Currently `ProcessAliases` do not allow to receive and modify (or filter) other process's outputs in real time while the process is running.

I am implementing wrappers for command line tools like SVN, xcodebuild and it would be great to have an opportunity to pipe outputs of tools or modify it for better build logs.

After a short Gitter discussion there is a suggestion to add following output handlers to `ProcessSettings`:

```
public Func<string, string> RedirectedStandardOutputHandler;
public Func<string, string> RedirectedStandardErrorHandler;
```

By @devlead:
> if one would want to filter out sensitive information i.e. one could retract that from the buffer and null if it should be totally omitted.

fseegraeber:
Closes #2270 .

Added `RedirectedStandardOutputHandler` and `RedirectedStandardErrorHandler` to `ProcessSettings` and the corresponding extension methods.

Moved the redirection handling into `ProcessWrapper`, by adding `StandardOutputReceived` and `StandardErrorReceived` methods. This keeps the `ConcurrentQueue` completely inside `ProcessWrapper` and facilitates the integration of features that need to modify the current queueing behaviour of the output. (E.g. #2354 and #1485)

### DIFF
--- a/src/Cake.Core.Tests/Fixtures/ProcessWrapperFixture.cs
+++ b/src/Cake.Core.Tests/Fixtures/ProcessWrapperFixture.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Diagnostics;
+using Cake.Core.Diagnostics;
+using Cake.Core.IO;
+using NSubstitute;
+
+namespace Cake.Core.Tests.Fixtures
+{
+    internal sealed class ProcessWrapperFixture
+    {
+        public Process Process { get; set; }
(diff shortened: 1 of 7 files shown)


devlead:
@fseegraeber your changes have been merged, thanks for your contribution 👍

PR #2450, opened 2019-01-14
Fix the numeric values of the enumeration

### LINKED ISSUE #2439: HtmlInline_AzurePipelines and MHtml shares the same numeric value
As raised by @dolifer in https://github.com/cake-build/cake/commit/73e1400383d5624e234b22ccede6d9f6115ab490#r31873452

> **HtmlInline_AzurePipelines** and **MHtml** shares the same numeric value, which causes incorrect output, when inline HTML is expected :(

This is a bug and needs to be corrected, affected lines

https://github.com/cake-build/cake/blob/73e1400383d5624e234b22ccede6d9f6115ab490/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs#L75

https://github.com/cake-build/cake/blob/73e1400383d5624e234b22ccede6d9f6115ab490/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs#L80

Values need to be unique, propose  `MHtml` gets put at end after
https://github.com/cake-build/cake/blob/73e1400383d5624e234b22ccede6d9f6115ab490/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs#L85
with value `15`.

Marking this as up for grabs / help wanted.

yansklyarenko:
Fixes #2439.

I've fixed it as suggested by @devlead. The appropriate fact was modified to a theory to cover all possible values of the enumeration. 

I've also dared to rename `HtmlInline_AzurePipelines` to `HtmlInlineAzurePipelines`. I understand that strictly speaking this violates your suggestion not to re-format the code, but as long as the change is tiny and quite visible, I hope it is ok.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/ReportGenerator/ReportGeneratorRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/ReportGenerator/ReportGeneratorRunnerTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Cake.Common.Tests.Fixtures.Tools;
 using Cake.Common.Tools.ReportGenerator;
 using Cake.Core;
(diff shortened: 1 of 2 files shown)


devlead:
@yansklyarenko your changes have been merged, thanks for your contribution 👍

Issue #2345, opened 2018-10-30
Allow NuGetRestoreSettings to opt out of setting -NonInteractive


flcdrg:
### What You Are Seeing?
`NuGetRestoreSettings` always sets `-NonInteractive`

### What is Expected?
Under certain circumstances, I'd like nuget to not run run with the `-NonInteractive` parameter - for example when NuGet is run in combination with the Azure Artifacts CredentialProvider.VSS.exe. In that case, when used at an interactive command prompt, I'd want the credential provider to prompt the user for credentials. Currently NuGet passes through the `-NonInteractive` parameter to CredentialProvider.VSS.exe and so it just fails (rather than prompting the user).

(In my case, I would check `System.Environment.UserInteractive` to determine whether `-NonInteractive` should be set or not)

### What version of Cake are you using?
0.30

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows
### Are you running on a CI Server?  If so, which one?
TeamCity and Azure Pipelines

### How Did You Get This To Happen? (Steps to Reproduce)

1. Use Azure Artifacts NuGet feed
2. Copy CredentialProvider.VSS.exe and put it in tools directory (alongside Cake's NuGet.exe)
3. Use NuGetRestore("myproject.sln") in Cake script
4. Note that script fails to create authentication token and no user prompt is shown.

devlead:
Fixed by #2451

PR #2451, opened 2019-01-16
GH2345: Add NonInteractive property to NuGetRestoreSettings

flcdrg:
- Defaults to true (preserving current behaviour as the default)
- Setting to false means NuGet.exe is free to prompt for user input/confirmation (useful when run interactively and when combined with authorisation providers)

Implements #2345


mholo65:
@flcdrg thank you for your contribution! Your changes have been merged! :+1:

PR #2455, opened 2019-01-19
GH2440:  Add EnvironmentVariable<T> alias

### LINKED ISSUE #2440: Add EnvironmentVariable<T> alias
Should be as convenient as [Argument\<T\>](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/ArgumentAliases.cs#L109) alias with default and support type converters.

gitfool:
Fixes #2440

### DIFF
--- a/src/Cake.Common.Tests/Unit/EnvironmentAliasesTests.cs
+++ b/src/Cake.Common.Tests/Unit/EnvironmentAliasesTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Cake.Core;
 using Cake.Testing;
 using NSubstitute;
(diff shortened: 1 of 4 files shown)


devlead:

gitfool:
@devlead I hit an issue after trying to use Cake after this merge - I've just discovered there's another implementation of this method in [Cake.Incubator/EnvironmentExtensions.cs#L64](https://github.com/cake-contrib/Cake.Incubator/blob/develop/src/Cake.Incubator/EnvironmentExtensions.cs#L64) that causes an error when compiling the build script:
```
Compiling build script...
(3312,10): error CS0111: Type 'Submission#0' already defines a member called 'EnvironmentVariable' with the same parameter types
Error: Cake.Core.CakeException: Error(s) occurred when compiling build script:
(3312,10): error CS0111: Type 'Submission#0' already defines a member called 'EnvironmentVariable' with the same parameter types
   at Cake.Scripting.Roslyn.RoslynScriptSession.Execute(Script script)
   at Cake.Core.Scripting.ScriptRunner.Run(IScriptHost host, FilePath scriptPath, IDictionary`2 arguments)
   at Cake.Commands.BuildCommand.Execute(CakeOptions options)
   at Cake.CakeApplication.Run(CakeOptions options)
   at Cake.Program.Main()
```
I don't know how I missed this since I always use Cake.Incubator, but I guess great minds think alike? 😊  

Anyway, the question is how to proceed from here? We can either revert, or move the other method over from Cake.Incubator and then remove them from there. Personally, I think it would be nice to keep this in Cake itself. What do you reckon?

devlead:
@gitfool this was bound to happen sooner or later.

Raised issue #2487 to track this.

Issue #1384, opened 2016-11-27
Enhancement: Add support for filtering files in Globbing alias


kcamp:
`GlobbingAliases.cs` has support for filtering the globbing traversal at the directory level, but no support for filtering files.  I have found that support for logic that is more complex than the globbing expression allows has been useful, and have seen gitter traffic for the same issue.

This would be a trivial addition to make at the method alias level.  I would propose that two overloads be created --

```cs
GetFiles(this ICakeContext context, string pattern, Func<IFile, bool> filePredicate)
 ...

GetFiles(this ICakeContext context, string pattern, Func<IDirectory, bool> predicate, Func<IFile, bool> filePredicate)
...
```

The former would satisfy the direct need of the feature request, the latter would provide a graceful pairing of current + extended logic.

Assuming that this is something that seems useful and aligned with the intent of the project, the question I would raise is whether it's more appropriate to implement the logic directly in the alias method, similar to .. 

```
// glob the files, and then filter down based on the filePredicate
var files = context.Globber.Match(pattern).OfType<FilePath>()
                            .Where(file => filePredicate(context.FileSystem.GetFile(file)));
```

or whether it would be more appropriately implemented in the globbing internals -- `GlobVisitor`, `GlobVisitorContext`, and so on. 

Constraining it to the alias would certainly be less involved, but the latter seems like it would probably be a cleaner (and faster) implementation, particularly in eliminating the extra calls to `FileSystem.GetFile(...)` -  it seems quite apt that it would amount to refactoring the `GlobVisitor` and `GlobVisitorContext` so that the logic in `FindCandidates` would be extended to encompass something like .. 

```
if (includeFiles)
{
    foreach (var file in current.GetFiles("*", option))
    {
        var lastPath = file.Path.Segments.Last();
        if (node.IsMatch(lastPath) && context.ShouldInclude(file))
        ....
```

Thoughts on whether or not this would be worth pursuing?  Any other feedback?

kcamp:
For what it's worth & work doesn't get duplicated, I've made good progress in extending the globbing to support this and expect to have a PR for it in the next couple of days.

devlead:
Fixed by #2458

PR #2458, opened 2019-01-20
GH1384: Add support to pass Func<IFile,bool> to globbing aliases

kcamp:
This satisfies the feature/enhancement described in GH-1384 by adding the member to the `GlobberSettings` type and wires in
all the necessary logic, providing a clean implementation
that relies on default values (null) and a single point of entry
for invocation to provide a flexible solution without the syntactic
overhead.

This refactors/revisits a [previous attempt](https://github.com/kcamp/cake/tree/GH-1384-old) to achieve the same.

Earlier work was based on methods/implementations that would have
come to rely on method overloads (before the `GlobberSettings`
implementation was added).  Now that the settings type exists,
this represents a much stronger feature/change without any potential
risk for backward compatibility.

i.e., an implicitly typed predicate specified as `IFileSystemInfo`
would have created ambiguity between the desired pairing of overloads -
`GetFiles(string pattern, Func<IFile, bool> predicate)`
and
`GetFiles(string pattern, Func<IDirectory,bool> predicate)`
which would have resulted in unpredictable behavior.

### DIFF
--- a/src/Cake.Core.Tests/Fixtures/GlobberFixture.cs
+++ b/src/Cake.Core.Tests/Fixtures/GlobberFixture.cs
@@ -125,10 +125,15 @@ namespace Cake.Core.Tests.Fixtures
             return Match(pattern, null);
         }
 
-        public Path[] Match(string pattern, Func<IFileSystemInfo, bool> predicate)
+        public Path[] Match(string pattern, Func<IFileSystemInfo, bool> directoryPredicate)
+        {
+            return Match(pattern, directoryPredicate, null);
+        }
+
+        public Path[] Match(string pattern, Func<IFileSystemInfo, bool> directoryPredicate, Func<IFile, bool> filePredicate)
         {
             return new Globber(FileSystem, Environment)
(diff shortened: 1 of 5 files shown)


src/Cake.Core.Tests/Fixtures/GlobberFixture.cs
```diff
@@ -125,10 +125,10 @@ public Path[] Match(string pattern)
             return Match(pattern, null);
         }
 
-        public Path[] Match(string pattern, Func<IFileSystemInfo, bool> predicate)
+        public Path[] Match(string pattern, Func<IFileSystemInfo, bool> directoryPredicate, Func<IFile, bool> filePredicate = null)
```
devlead: Keep `Match(string pattern, Func<IFileSystemInfo, bool> predicate)` as an overload calling this method so we don't break binary compatibility.

kcamp:
Rebased on `develop` and made requested change. Let me know if there's anything else required.

devlead:
@kcamp  your changes have been merged, thanks for your contribution 👍

PR #2469, opened 2019-02-06
(GH-2468) Convert the DotNetCoreVerbosity to lowercase

### LINKED ISSUE #2468: DotNet commands do not respect the verbosity
As the title says, the DotNetCommands do not respect the verbosity setting. The problem seems to be that the verbosity expects a lowercase text (see https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet?tabs=netcore21) but cake just uses the `Enum.ToString()`.
I will create a PR which adds a `ToLower()`.

Roemer:
Fixes #2468


devlead:

Issue #2474, opened 2019-02-14
TeamCityProvider.BuildProblem method should conform to TeamCity API 


kcamp:
### What You Are Seeing?
Cake is failing to successfully write TeamCity `buildProblem` service messages

### What is Expected?
Cake should validate arguments to methods to conform with documented TeamCity API and either omit them or provide sane defaults as necessary.  In my opinion, for this case, it should be an omission rather than an `string.Empty`.

Per the [docs](https://confluence.jetbrains.com/display/TCD9/Build+Script+Interaction+with+TeamCity)

```
##teamcity[buildProblem description='<description>' identity='<identity>']
where
description - (mandatory) a human-readable plain text describing the build problem. 
By default, the description appears in the build status text and in the list of build's problems. 
The text is limited to 4000 symbols, and will be truncated if the limit is exceeded.

identity - (optional) a unique problem id. 
Different problems must have different identity, same problems - same identity, which should not change throughout builds if the same problem occurs, e.g. the same compilation error. 
It should be a valid Java id up to 60 characters. If omitted, the identity is calculated based on the description text.
```

If the caller has omitted the `identity` parameter, it should't be passed as a `KeyValuePair<string, string>` with a `null` value to the `WriteServiceMessage` method.

### What version of Cake are you using?
0.29.0 (but have validated that the problematic code `main`/`0.32.1`)

### Are you running on a 32 or 64 bit system?
x64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
TeamCity

### How Did You Get This To Happen? (Steps to Reproduce)
I have this in my error deferral mechanism, being invoked in a custom teardown method.
```cs
// in this case, problem.Identity is null, which creates the issue.
if(_context.TeamCity().IsRunningOnTeamCity)
{
    _context.TeamCity().BuildProblem(problem.Message, problem.Identity);
}
```

(I have since amended the script for `problem.Identity ?? ""` to successfully work around this issue.

### Output Log
```
[08:47:57]W:	 [Step 1/1] An error occurred in a custom teardown action.
[08:47:57]W:	 [Step 1/1] Error: System.AggregateException: One or more errors occurred. ---> System.NullReferenceException: Object reference not set to an instance of an object.
[08:47:57]W:	 [Step 1/1]    at Cake.Common.Build.TeamCity.TeamCityProvider.Sanitize(String source)
[...]

PR #2475, opened 2019-02-15
GH2474 Refactor TeamCityProvider.BuildProblem method

kcamp:
This addresses GH-2474 and brings the `BuildProblem` method to be more consistent with the expected usage within TeamCity.

This change will suppress the "identity" token if the value provided is null or empty.  This directory resolves the issue of the provider attempting to sanitize a null reference on a value that, per Jetbrains documentation, is not required for the `buildProblem` service message


devlead:

PR #2482, opened 2019-02-19
Implicit operators should not throw

### LINKED ISSUE #2481: FilePath and DirectoryPath implicit conversions throw for null
Cake 0.30.0

This is the code I would like to write. It uses a conditional expression to keep it short and sweet:

```cs
        VSTest(testAssembly, new VSTestSettings
        {
            SettingsFile = keepWorkspaces ? "KeepWorkspaces.runsettings" : null
        });
```

If `keepWorkspaces` is false, that causes this error:

> Value cannot be null.
> Parameter name: path

Presumably `SettingsFile` is already null to start with, representing the option to not use a settings file, so null should be a valid option when setting the property.

As a workaround, I have to introduce two new statements and break the order that I'd prefer:

```cs
        var settings = new VSTestSettings();
        if (keepWorkspaces) settings.SettingsFile = "KeepWorkspaces.runsettings";
        VSTest(testAssembly, settings);
```

jnm2:
Fixes #2481.

### DIFF
--- a/src/Cake.Core/IO/DirectoryPath.cs
+++ b/src/Cake.Core/IO/DirectoryPath.cs
@@ -143,7 +143,7 @@ namespace Cake.Core.IO
         /// <returns>A <see cref="DirectoryPath"/>.</returns>
         public static implicit operator DirectoryPath(string path)
         {
-            return FromString(path);
+            return path is null ? null : FromString(path);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


devlead:

jnm2:
@devlead Thanks for the quick merge! I have no idea what I was thinking, but I didn't add tests. PR soon.

jnm2:
See #2483.

PR #2485, opened 2019-02-21
GH2484: Add Spaces support to Octopus Deploy settings and octo.exe aliases

kcamp:
This adds the `--space YourSpaceName` parameter to the `OctopusDeployCommonToolSettings` and updates the implementation in `OctopusDeployArgumentBuilder`.    This covers all aliases that are currently implemented except `pack`, which doesn't accept the optional space qualifier.

As noted in the issue, this will require users to use at least v5.2.0 of `octo.exe`.


devlead:

PR #2457, opened 2019-01-19
GH2456: Add logging aliases to override the log verbosity

gitfool:
Fixes #2456:
- Add `Disposable` helper for creating anonymous disposables
- Refactor `TeamCityDisposableExtensions` to use `Disposable` helper
- Add logging aliases and extension methods:
  - QuietVerbosity, MinimalVerbosity, NormalVerbosity, VerboseVerbosity, DiagnosticVerbosity

Example usage:
```c#
Task("Test")
    .Does(() =>
{
    Debug("Hide me");
    using (DiagnosticVerbosity())
    {
        Error("Show me.");
        Warning("Show me.");
        Information("Show me.");
        Verbose("Show me.");
        Debug("Show me.");
    }
    Debug("Hide me again");
});
```

Issue #820, opened 2016-04-13
Log tools command-line at higher log level (preferably default)


gitfool:
I'm used to and prefer being able to see the command-line of all tools being run by default. This is also the behavior in TeamCity. Obviously subjective though.

References:
- [ProcessRunner.cs#L63](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessRunner.cs#L63)
- [ProcessWrapper.cs#L61](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessWrapper.cs#L61)

petriashev:
see my answer: https://stackoverflow.com/questions/50826394/how-to-print-tool-command-line-in-cake/50826495#50826495

devlead:
Fixed by #2486

PR #2486, opened 2019-02-22
GH820: Add global override to show process command-line

gitfool:
Following up from #2457, I can cleanly address #820 with a global setting that temporarily overrides the process runner log level.

Example usage:
```
$env:Cake_Settings_ShowProcessCommandLine='true';  .\build.ps1
```

### DIFF
--- a/src/Cake.Core/IO/ProcessRunner.cs
+++ b/src/Cake.Core/IO/ProcessRunner.cs
@@ -23,6 +23,7 @@ namespace Cake.Core.IO
         private readonly IToolLocator _tools;
         private readonly ICakeConfiguration _configuration;
         private readonly bool _noMonoCoersion;
+        private readonly bool _showCommandLine;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ProcessRunner" /> class.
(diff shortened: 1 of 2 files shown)


src/Cake.Core/IO/ProcessRunner.cs
```diff
@@ -123,9 +126,12 @@ internal ProcessStartInfo GetProcessStartInfo(FilePath filePath, ProcessSettings
...
             if (!settings.Silent)
             {
-                // Log the filename and arguments.
-                var message = string.Concat(fileName, " ", arguments.RenderSafe().TrimEnd());
-                _log.Verbose(Verbosity.Diagnostic, "Executing: {0}", message);
+                using (_log.WithVerbosity(_showCommandLine ? Verbosity.Diagnostic : _log.Verbosity))
+                {
+                    // Log the filename and arguments.
+                    var message = string.Concat(fileName, " ", arguments.RenderSafe().TrimEnd());
+                    _log.Verbose(Verbosity.Diagnostic, "Executing: {0}", message);
```
devlead: Is using WithVerbosity needed?

Wouldn't if be enough to 

```csharp
_log.Verbose(_showCommandLine  ? _log.Verbosity : Verbosity.Diagnostic,
    "Executing: {0}", message);
```

gitfool: Oh yeah. Doh!

gitfool:
@devlead your suggestion worked, of course, so changed as requested.

devlead:

Issue #2087, opened 2018-03-16
Include more detailed exception information when Exception is AggregateException


richardgavel:
I have Tasks that will execute a Parallel.Invoke of an array of Actions. However, when an error happens, all I'll see is the "One or more errors occured". Maybe AggregateExceptions should output the message of each child exception.

devlead:
Which version on Cake are you using, have you tried executing Cake with [diagnostic verbosity](https://stackoverflow.com/questions/38658660/how-to-enable-diagnostic-verbosity-for-cake)?

jnm2:
If changing the verbosity affects it, I would wonder is it ever helpful to see only the message in the special case of an AggregateException?

devlead:
@jnm2 Issue could be that it's already within an aggregate exception because the async support. So could be an extra layer of abstraction. Whereas diagnostic gives the whole stack trace.

jnm2:
@devlead It is an extra layer of abstraction because he is using `Parallel.Invoke`. I ran into the same thing with `Task.WhenAll`. It's sort of like if I wrote `throw new AggregateException(new InvalidOperationException("foo"))`. I could be wrong, but wouldn't it be nice if Cake showed the messages of the InnerExceptions collection even when I'm on the default verbosity and not looking for the stack trace?

devlead:
Well anything's possible would need to be recursive then I guess. to fetch inner inner exception messages and so on too.

jnm2:
Yes, that's a good idea! it would be like Visual Studio's exception helper window!

```cs
throw new AggregateException(
    new AggregateException(
        new InvalidOperationException("hello")));
```

Makes it show:

![image](https://user-images.githubusercontent.com/8040367/37523783-599c2c86-28fe-11e8-8b84-0d6fa10b8d0a.png)

And here's the middle one:

![image](https://user-images.githubusercontent.com/8040367/37523826-79877262-28fe-11e8-9096-399664a4a434.png)

patriksvensson:
I wonder if Ben Adams magic exception demystifier exists as a source package... 😄

Duracell1989:
Some issue here;
I do multiple validations and combine them in 1 AggregateException.
I don't want the users to: fix 1 validation; re-run; new error; fix; re-run; etc.
That's why I want to combine them in 1 AggregateException and then I just need to throw 1 exception with all the relevant information.

tlecomte:
[...]

PR #2500, opened 2019-03-18
GH2087: log nested aggregate exceptions

tlecomte:
https://github.com/cake-build/cake/issues/2087

When the exception thrown in a task is actually a nested `AggregateException`, the output log is:

```
An error occurred when executing task 'taskName'.
Error: One or more errors occurred.
	One or more errors occurred.
```

The first occurence of `One or more errors occurred.` is the message of the parent `AggregateException`. The second occurence is the message if the inner exception, which is itself another `AggregateException`.

Here we add a call to `.Flatten()`. This creates a new single instance of `AggregateException` without nesting, so that the last log line will now show the actual inner exception message.

Reference: https://docs.microsoft.com/en-us/dotnet/api/system.aggregateexception.flatten?view=netframework-4.7.2

### DIFF
--- a/src/Cake/Program.cs
+++ b/src/Cake/Program.cs
@@ -100,7 +100,7 @@ namespace Cake
                 var aex = ex as AggregateException;
                 if (aex != null)
                 {
-                    foreach (var exception in aex.InnerExceptions)
+                    foreach (var exception in aex.Flatten().InnerExceptions)
                     {
                         log.Error("\t{0}", exception.Message);
                     }


tlecomte:
cc @jlbrown

tlecomte:
An integration test is failing on 3 Azure pipelines (out of 6): `HWApp.Tests.GreeterTests+TheGreaterMethod.Should_Not_Fail_Test [FAIL]`
Looking at this test, I don't any clear link with the change in this PR.

Please let me know if I can do something about that! Thanks!

devlead:
That test is supposed to fail :)
Issue seems to be with the T4 tool template execution, I'll try a take look that and get back to you.

devlead:
@tlecomte your changes have been merged, thanks for your contribution 👍

Issue #2461, opened 2019-01-25
RedirectedStandardOutputHandler &  RedirectedStandardErrorHandler on ToolSettings


devlead:
As discussed in #2460, we should expose `RedirectedStandardOutputHandler` and  `RedirectedStandardErrorHandler` on `ToolSettings`, which was just recently added to `ProcessSettings` in pr #2447.

fseegraeber:
I'd suggest that, instead of exposing the `RedirectedStandardOutputHandler` and `RedirectedStandardErrorHandler` from `ProcessSettings`, we extend `ToolSettings` by an handler that gets passed the stdout and stderror, but doesnt intercept. We could implement that using a intercepting `ProcessSettings.RedirectedStandardOutputHandler` from inside `Tool`. Possibly a `ToolSettings.StandardOutputHandler` and `ToolSettings.StandardErrorHandler` ?

That way we can grant access to stdout and stderror of a Tool without breaking the `GetStandardOutput()` some Tools use.

A downside to this would be that other `Tool` implementations could not use the `ProcessSettings.RedirectedStandardOutputHandler`.

augustoproiete:
Related discussion: https://github.com/cake-build/cake/discussions/3202

pascalberger:
This came up a few time lately. Also related discussion: https://github.com/cake-build/cake/discussions/3130

devlead:
Closing this as completed via `ToolSettings.SetupProcessSettings`.

That delegate is invoked on the `ProcessSettings` used to start the tool, so the `ProcessSettings` redirect flags and handlers from #2447 are already available on every tool alias without dedicated properties on `ToolSettings`.

```csharp
new DotNetTestSettings
{
    SetupProcessSettings = p =>
    {
        p.RedirectStandardOutput = true;
        p.RedirectedStandardOutputHandler = line =>
        {
            // inspect / log; return the line to keep pass-through
            return line;
        };
    }
}

PR #2508, opened 2019-03-24
Fix typo

gitfool:
I noticed what looks like a copy-paste typo in [Cake.Core/IO/ProcessWrapper.cs#L71](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessWrapper.cs#L71) when looking around the code for #2461.


devlead:

Issue #2375, opened 2018-12-04
Add convenience extensions to TextTransformationExtensions


gitfool:
I would like to add convenience extensions to TextTransformationExtension:

Support for dictionaries:

```c#
public static TextTransformation<TTemplate> WithTokens<TTemplate>(
    this TextTransformation<TTemplate> transformation, IDictionary<string, object> tokens)
    where TTemplate : class, ITextTransformationTemplate
{
    if (transformation != null && tokens != null)
    {
        foreach (var token in tokens)
        {
            if (token.Value != null) // leave placeholder for tokens without a value
            {
                transformation.Template.Register(token.Key, token.Value);
            }
        }
    }
    return transformation;
}
```
Support for shallow object properties, with optional key prefix, that feeds the first extension method:
```c#
public static Dictionary<string, object> ToTokens(this object value, string prefix = null, BindingFlags? bindingAttr = null)
{
    return value?.GetType()
        .GetProperties(bindingAttr ?? BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public)
        .ToDictionary(property => string.Concat(prefix, property.Name), property => property.GetValue(value));
}
```

Thoughts? Happy to submit a PR!

gitfool:
Here's an example of how to use the `ToTokens` convenience extension:
```c#
Task("StageManifest")
    .Does(() =>
{
    TransformTextFile(Build.Files.Manifest, "{{", "}}")
        .WithTokens(Build.Version.ToTokens("Version."))
        .Save(Build.Files.ArtifactsManifest);
});
```
where the manifest contains placeholders for `Version` properties (of type `GitVersion`):
```yaml
---
name: Test
version:
    semantic_version: {{ Version.FullSemVer }}
    assembly_version: {{ Version.AssemblySemVer }}
    assembly_file_version: {{ Version.AssemblySemFileVer }}
    informational_version: {{ Version.InformationalVersion }}
...
```

gitfool:
I use a much more elaborate `ToTokens` extension now... [Cake.Dungeon/scripts/extensions.cake#L9](https://github.com/gitfool/Cake.Dungeon/blob/master/scripts/extensions.cake#L9).

PR #2509, opened 2019-03-24
Use IEnumerable<KeyValuePair<K, V>> in TextTransformation.WithTokens

gitfool:
Following up from #2375 and the recently merged #2124, I can't easily use this extension without being forced to materialize a dictionary, which is inconvenient, especially if I need to filter the keys first (using LINQ), hence this tweak to use the `IEnumerable<KeyValuePair<K, V>>` base type instead.


devlead:

PR #2515, opened 2019-03-26
GH2514: Add additional report types to ReportGenerator

### LINKED ISSUE #2514: Add additional report types for ReportGenerator
Add new report types, and correct `HtmlInline_AzurePipelines` that was renamed in https://github.com/cake-build/cake/commit/d73c086976d92ab1e7ceffd61a35c1026d18d2b4

- HtmlInline_AzurePipelines_Dark (since 4.0.10),
- SonarQube (since 4.0.6)

flcdrg:
- Also correct `HtmlInline_AzurePipelines`

Fixes #2514

### DIFF
--- a/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs
+++ b/src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs
@@ -72,7 +72,7 @@ namespace Cake.Common.Tools.ReportGenerator
         /// <summary>
         /// Same as HTMLInline but with modified CSS that matches the look and feel of Azure Pipelines.
         /// </summary>
-        HtmlInlineAzurePipelines = 13,
+        HtmlInline_AzurePipelines = 13,
 
         /// <summary>
         /// A single PNG file containing a chart with historic coverage information.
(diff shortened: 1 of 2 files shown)


src/Cake.Common/Tools/ReportGenerator/ReportGeneratorReportType.cs
```diff
@@ -72,7 +72,7 @@ public enum ReportGeneratorReportType
         /// <summary>
         /// Same as HTMLInline but with modified CSS that matches the look and feel of Azure Pipelines.
         /// </summary>
-        HtmlInlineAzurePipelines = 13,
+        HtmlInline_AzurePipelines = 13,
```
devlead: This value has already shipped with Cake so would be a breaking change.

flcdrg: Yeah.. this is tricky. I don't know why it was changed in https://github.com/cake-build/cake/commit/d73c086976d92ab1e7ceffd61a35c1026d18d2b4, but the correct parameter to pass to ReportGenerator.exe has the underscore in it.

Has this actually shipped though? The last release v0.32.1 was on Jan 4, and that breaking commit was made on Jan 14, so if this is merged before the next release then everything would be ok?

gep13: Agreed, this hasn't shipped yet.

devlead: Flag has shipped but had wrong number. This is in common and if tool expects with underscore this is correct. Agree this should break and mark this as resolved.

devlead:
@flcdrg your changes have been merged, thanks for your contribution 👍

Issue #1618, opened 2017-05-29
Support different Git servers in TeamCityPullRequestInfo


pascalberger:
### What You Are Seeing?
[TeamCityPullRequestInfo](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs) currently works only with certain Git servers like Stash, since it expects pull requests in `refs/pull-requests/`. This doesn't work for other Git servers like eg. TFS.

### What is Expected?
It should be possible to use the class with any Git server

### What version of Cake are you using?
0.19.5

### Are you running on a 32 or 64 bit system?
64 bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
TeamCity

### How Did You Get This To Happen? (Steps to Reproduce)
* Create a pull request on TFS and trigger the build for the pull request
* `TeamCityPullRequestInfo.IsPullRequest` returns `false`

mholo65:
:+1: e.g. Gerrit uses `refs/changes`, and Github `refs/pull`.

This should also be taken into account in other build systems.

Should the ref be user configurable or should Cake check some known possibilities? I bet Gitlab, Github, Bitbucket, Gerrit, TFS, etc all use different refs for PRs..

mvput:
Sounds best to make this configurable for the user. Setting someting like `TeamCity.VcsProvider = VcsProviders.GitHub`. With that you can put more configuration settings for that particular git host.

Kralizek:
Any information about this?

devlead:
@Kralizek I believe this issue is till up for grabs, I'll tag it so.

Kralizek:
I could try to give a swing at it. Any specific design direction I should consider?

devlead:
@Kralizek cool, then I'll remove the up-for-grabs/help wanted labels.
Read through the [Contribution Guidelines](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md), feel free to ping if you have any questions. 
Running release build will catch many code style issues

These are the points that would need to be smarter
https://github.com/cake-build/cake/blob/a0f142af2dfbf18eecc240789529da428cb0d82a/src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs#L23

[...]

PR #2555, opened 2019-05-15
GH1618: Enhance the Teamcity Provider's PullRequest detection

nmbro:
Implements a lot of #1618

### DIFF
--- a/src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs
+++ b/src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs
@@ -11,6 +11,39 @@ namespace Cake.Common.Build.TeamCity.Data
     /// </summary>
     public class TeamCityPullRequestInfo : TeamCityInfo
     {
+        private static bool InferIsPullRequest(string gitReferenceName)
+        {
+            var branchSlices = gitReferenceName.Split('/');
+
+            if (branchSlices.Length >= 3)
+            {
+                switch (branchSlices[1].ToUpper())
+                {
+                    case "CHANGES":
+                    case "MERGE-REQUESTS":
+                    case "PULL":
+                    case "PULL-REQUESTS":
+                        return true;
+                    default:
+                        return false;
+                }
+            }
+
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


nmbro:
Tested working in my Teamcity environment using Github

src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs
```diff
@@ -11,6 +11,39 @@ namespace Cake.Common.Build.TeamCity.Data
     /// </summary>
     public class TeamCityPullRequestInfo : TeamCityInfo
     {
+        private static bool InferIsPullRequest(string gitReferenceName)
+        {
+            var branchSlices = gitReferenceName.Split('/');
```
devlead: Should check & return if `gitReferenceName` is null or empty.

src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs
```diff
@@ -11,6 +11,39 @@ namespace Cake.Common.Build.TeamCity.Data
...
+                    default:
+                        return false;
+                }
+            }
+
+            return false;
+        }
+
+        private static int? GetPullRequestNumber(string gitReferenceName)
+        {
```
devlead: Should check & return if `gitReferenceName` is null or empty.

src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs
nmbro: done

src/Cake.Common/Build/TeamCity/Data/TeamCityPullRequestInfo.cs
nmbro: done

nmbro:
@devlead done! ;)

martin-pritchard:
Hi @nmbro / @devlead. Just ran into this bug, and found this PR. What's outstanding? Could see the requested changes, but they appear to have been addressed.

Issue #2551, opened 2019-05-11
Call MSBuild without specifying a target does not use DefaultTarget


Diogomrol:
### What You Are Seeing?
When I call MSBuild without specifying WithTarget(), it looks like MSBuild tries to build my project using "Build" target instead of using the file's DefaultTarget.

I'm calling a MSBuild-compatible XML file as the solution input (namely Babel Obfuscator's configuration file, which I can execute through the command line using MSBuild).

### What is Expected?
I expected that DefaultTarget would be used if I do not specify any other target.

### What version of Cake are you using?
0.33

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No, local machine.

### How Did You Get This To Happen? (Steps to Reproduce)
Create a target like this:

```
var obfuscate = Task("Obfuscate")
    .Does(() => {
        MSBuild("obfuscatorConfigurationFile.babel");
    });
```

### Output Log

> Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
> Copyright (C) Microsoft Corporation. All rights reserved.
> 
> Build started 11/05/2019 21:51:34.
> Project "C:\svn\branch\ObfuscationConfiguration.babel" on node 1 (Build target(s)).
> "C:\svn\branch\ObfuscationConfiguration.babel" (Build target) (1) ->
>  C:\svn\branch\ObfuscationConfiguration.babel : error MSB4057: The target "Build" does not exist in the project.

Diogomrol:
Well I have been looking at the code and it defaults to target:Build if no Target is passed, as I observed. 

Why not calling msbuild without the Target parameter if no Target is provided? Any reason for it to be implemented that way?

z93blom:
We are running into the same problem. 

I understand the fact that it might not be possible to change how the default is implemented (since it might potentially break a lot of existing scripts), but there needs to at least be some way of overriding the way that Cake automatically defaults to target:Build if there is no Target passed, so that we can call MSBuild with an empty target.

This is currently blocking us from moving to Cake.

devlead:
Could have a `NoTarget` that defaults to false on MSBuild settings class. That way there's a flag to opt-out, with retained backwards compat.

Diogomrol:
@devlead yup, just thought about doing something in that way. May I make a PR for this?

devlead:
Sure go ahead.

[...]

PR #2557, opened 2019-05-23
GH2551: Implement the NoImplicitTarget property and extension method in MSBuild

Diogomrol:
Tackling issue #2551.

Brief summary of changes:
- Added `NoImplicitTarget `property to MSBuild;
- Added `SetNoImplicitTarget()` extension method to get/set the property mentioned above;
- Added some tests to validate the relevant scenarios for this change.

Some behavior details:
- If the developer provides no targets (meaning doesn't use `WithTarget()`) and doesn't use `NoImplicitTarget`, MSBuild will be called with `/target:Build`, keeping the current behavior;
- If the developer provides no targets (meaning doesn't use `WithTarget()`) and sets `NoImplicitTarget` to true, MSBuild will be called with no targets at all;
- If the developer provides targets (using `WithTarget()`) the provided targets will be used regardless of the value of `NoImplicitTarget`.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildRunnerTests.cs
@@ -610,6 +610,37 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
             }
 
             [Fact]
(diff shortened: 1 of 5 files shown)


devlead:
@Diogomrol your changes have been merged, thanks for your contribution 👍

PR #2565, opened 2019-06-10
GH2663: Add support for Inno Setup 6

Pireax:
This PR adds support for the new Inno Setup 6. By default the newest available version is used, the version to use can also be set by using the Version property on the settings class.


devlead:
@Pireax your changes have been merged, thanks for your contribution 👍

Issue #2571, opened 2019-07-04
OctopusDeploy DeployTo property to take collection of string to specify multiple environments


nckl:
As currently implemented the DeployTo property on both CreateReleaseSettings and OctoDeployRelease method only allows the specification of one environement, however the octo.exe documentation here:  
https://octopus.com/docs/octopus-rest-api/octo.exe-command-line/create-release 
https://octopus.com/docs/octopus-rest-api/octo.exe-command-line/deploy-release

states that deployTo can be specified many times for multiple environments.

> --deployto=VALUE       [Optional] Environment to automatically deploy
                             to, e.g., Production; specify this argument
                             multiple times to deploy to multiple environments

nckl:
Modifying the existing DeployTo property would be a breaking change for anyone who is currently using that property. What are the guidelines when it comes to breaking changes? Should there be a new property instead perhaps called DeployToMany that is the collection?

devlead:
We try to avoid breaking changes generally, unless they make sense then we usually do it by obsoleting property with warning first then removing it in a future release.

PR #2574, opened 2019-07-11
Convert octopus deploy to property to collection

nckl:
In this pull request I have added support to enable multiple environments to be deployed to using the deployTo parameter multiple times as well as added the ExcludeMachines parameter which are both supported by the octopus cli.

In order to support multiple deployTo parameters without interfering with existing implementation a new string array has been added and the original deployTo has been left alone within the CreateReleaseSettings class. Additionally a new overload method has been created in the OctopusDeployAliases class in which the original will call the new overload once the single deployTo parameter has been made an array. This is to prevent code duplication.

At the moment I split if it makes sense to mark the original deployTo as obsolete because those who only deploy to a single environment shouldn't have to create a collection of one to do so. Open to thoughts/suggestions.

Conversation about this issue has been brief but is contained here: https://github.com/cake-build/cake/issues/2571


nckl:
@devlead, Unless I have done something wrong, which is entirely possible as this is my first PR. The issue that I had created earlier was https://github.com/cake-build/cake/issues/2571. If I have done something incorrectly  please let me know. Thanks.

augustoproiete:
LGTM. @nckl Would you be able to rebase  against latest `develop` & resolve conflicts?

PR #2576, opened 2019-07-17
GH-2578: Support version ranges for referenced nuget packages

msioen:
Extended the version pinning support of the nuget package installer to also support version ranges.

Would support formats like the following to indicate you want to stay on major version 1 but ok to get minor/fix updates.
#addin nuget:?package=Cake.Foo&version=[1.0.0,2.0.0)

Our usecase is for a set of shared buildscripts which will be used across several projects. We'd like to version pin on major changes but still allow bugfixes to go through without having to change all the projects. This isn't currently possible.

Note: this code isn't really covered by any unit test - I'm also not really sure on how to add one to specifically test this.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -664,7 +664,7 @@ Task("Run-Integration-Tests")
             },
             Arguments = {
                 ["target"] = Argument("integration-tests-target", "Run-All-Tests"),
-                ["verbosity"] = "quiet",
+                ["verbosity"] = "diagnostic",
                 ["platform"] = parameters.IsRunningOnWindows ? "windows" : "posix",
                 ["customarg"] = "hello"
             }
--- a/src/Cake.NuGet/Install/NuGetPackageInstaller.cs
+++ b/src/Cake.NuGet/Install/NuGetPackageInstaller.cs
@@ -278,10 +278,14 @@ namespace Cake.NuGet.Install
                     {
                         foundVersion = foundVersions.FirstOrDefault();
                     }
+                    if (foundVersion == null)
+                    {
+                        continue;
+                    }
 
                     // Find the highest possible version
                     version = version ?? foundVersion;
-                    if (foundVersion != null && foundVersion > version)
+                    if (foundVersion > version)
                     {
                         version = foundVersion;
                     }
--- a/src/Cake.NuGet/Install/NuGetPackageInstaller.cs
+++ b/src/Cake.NuGet/Install/NuGetPackageInstaller.cs
@@ -241,9 +241,18 @@ namespace Cake.NuGet.Install
 
         private static NuGetVersion GetNuGetVersion(PackageReference package, IEnumerable<SourceRepository> repositories, NuGetFramework targetFramework, SourceCacheContext sourceCacheContext, ILogger logger)
         {
+            VersionRange versionRange = null;
             if (package.Parameters.ContainsKey("version"))
             {
-                return new NuGetVersion(package.Parameters["version"].First());
+                var requestedVersion = package.Parameters["version"].First();
+                if (NuGetVersion.TryParse(requestedVersion, out NuGetVersion parsedVersion))
+                {
+                    return parsedVersion;
+                }
+                else if (VersionRange.TryParse(requestedVersion, out VersionRange parsedVersionRange))
+                {
+                    versionRange = parsedVersionRange;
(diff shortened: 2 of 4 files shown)


msioen:
Must have glossed over the required issue while reading the guidelines, sorry. Just created one with some additional background: https://github.com/cake-build/cake/issues/2578

I'll have a look at tests tomorrow. Can you give me a hint how to start on this? I might be missing something but there doesn't seem to be a test I can really start from.
- the 'uri parsing' logic could be made public and unit tested separately: clear input uri and output nuget version with expected data => this does change public interface
- a separate testfixture could be made for the Install/NuGetPackageInstaller - currently only the mocked nuget installer is used in tests => not sure if this will work properly
- maybe something else?

I'd like to avoid writing the tests in a way that aren't desired

devlead:
Regarding tests we could start with integration tests ( https://github.com/cake-build/cake/blob/develop/tests/integration/Cake.Core/Scripting/AddinDirective.cake )

I.e. load an nuget package an with an in-between version number and check assembly version on a type in that assembly.

msioen:
Pushed some tests. Note that my initial example wasn't correct according to expected nuget versioning behaviour.
'normal' version ranges will favor the lowest version possible within the range ie [1.0.0,2.0.0) will return 1.0.0 if both 1.0.0 and 1.1.0 exist.
wildcard version ranges do offer the functionality I proposed which is part of the same nuget versioning class so also implemented. ie [1.*,2.0.0) will favor the highest version possible within 1->2 major range

exec-starlizard:
This would be a great feature that we're excited to use.  

Is there anything we can help with here?  Is the remaining items on someone's list or can I get involved?

megakid:
Is there anything I can help out here?  This is a great feature that'd we're looking forward to putting to good use.

msioen:
Kind of lost track of this due to holidays. The code and unit tests work on mac and linux, no idea why the tests fail on Windows.

@megakid (or someone else)
I currently don't have a dev machine setup with windows so if anyone can have a look on Windows what's going wrong there that'd be great.

megakid:
Ok so locally on my windows box, the only test that fails is `Tool_settings_collection_properties_must_be_initialized`.  However, when I check out the `develop` branch, the same test fails so unsure if that's the test failing above.

Unfortunately the build history has already been cleared - can we trigger another build somehow? a new commit to this branch perhaps?

megakid:
Actually that test only failed within Visual Studio.

Using `build.ps1`, everything worked.

msioen:
I had a chance to test on Windows today as well. The integration tests seem to work ok locally. Rebased develop and pushed to see if the automated builds go through or not.

megakid:
I can see the failure now but haven't looked at what's causing it:

```Starting test execution, please wait...
[xUnit.net 00:00:00.8333158]     HWApp.Tests.GreeterTests+TheGreaterMethod.Should_Not_Fail_Test [FAIL]
  X HWApp.Tests.GreeterTests+TheGreaterMethod.Should_Not_Fail_Test [12ms]
  Error Message:
   Assert.NotEqual() Failure
Expected: Not "true"
Actual:   "true"
  Stack Trace:
     at HWApp.Tests.GreeterTests.TheGreaterMethod.Should_Not_Fail_Test() in D:\a\1\s\tests\integration\temp\Cake.Common\Tools\DotNetCore\hwapp.tests\Tests.cs:line 14

Test Run Failed.
Total tests: 3
     Passed: 2
     Failed: 1
 Total time: 2.5399 Seconds
Shutting down MSBuild server...
Shutting down VB/C# compiler server...
MSBuild server shut down successfully.
VB/C# compiler server shut down successfully.
Package Source with Name: TestSource added successfully.
Package source with Name: TestSource removed successfully.
Error: Failed to install addin 'Serilog'.
An error occurred when executing task 'Cake.Nuget.Install.NugetPackageInstaller.VersionPinFull'.
Error: One or more errors occurred. (.NET Core CLI: Process returned an error (exit code 1).)
	.NET Core CLI: Process returned an error (exit code 1).
An error occurred when executing task 'Run-Integration-Tests'.```

devlead:
@msioen huge thanks for your contribution 👍 ❤️ 
This PR has been superseded by #3240, so I'm going to go ahead a close this PR.

Issue #2527, opened 2019-04-11
Dont limit the #load to only .cake files


llaaccssaapp:
### What You Are Seeing?
Since 0.33 there is no longer support for loading .cs files with the #load directive

### What is Expected?
#l "./file.cs" will include the file and parse it

### What version of Cake are you using?
0.33

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Code location
 src/Cake.Core/Scripting/Processors/Loading/FileLoadDirectiveProvider.cs
public bool CanLoad(IScriptAnalyzerContext context, LoadReference reference)
return extension != null && extension.Equals(".cake", StringComparison.OrdinalIgnoreCase);

please allow .cake and .cs file extensions

slicht:
We need this as well. We can't use the current implementation since we need .cs files for code completion in VS.

mythgarr:
This is also blocking me from updating. Using an older release will complicate things - I'd like to use some features introduced in January.
I'm not sure why the Where is necessary here at all - wouldn't you just add *.cake to the end of the glob? Maybe I'll just put together a pull request and see if it gets accepted.

mythgarr:
As part of supporting globbing within the #load directive the file filter was made consistent with nuget packages. I'm looking for ways to make that filter more selective so that explicit file paths are always included, globs are always filtered.
Even this seems a little excessive. If I write `#load "plugins/*.cs"` I've explicitly stated that I want to load all cs files at the specific location.

PR #2588, opened 2019-08-02
GH2527: #load directives that explicitly look for a specific extension will include all matching files

mythgarr:
Fixes https://github.com/cake-build/cake/issues/2527

### DIFF
--- a/src/Cake.Core/Scripting/Processors/Loading/FileLoadDirectiveProvider.cs
+++ b/src/Cake.Core/Scripting/Processors/Loading/FileLoadDirectiveProvider.cs
@@ -54,15 +54,17 @@ namespace Cake.Core.Scripting.Processors.Loading
             var current = context.Current.Path.GetDirectory();
             path = path.MakeAbsolute(current);
 
+            var expectedExtension = path.HasExtension ? path.GetExtension() : ".cake";
             var files = _globber
                 .GetFiles(path.FullPath)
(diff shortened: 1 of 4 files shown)


src/Cake.Core/Scripting/Processors/Loading/FileLoadDirectiveProvider.cs
```diff
@@ -54,14 +54,16 @@ public void Load(IScriptAnalyzerContext context, LoadReference reference)
             var current = context.Current.Path.GetDirectory();
             path = path.MakeAbsolute(current);
 
+            var expectedExtension = path.HasExtension ? path.GetExtension() : ".cake";
```
mythgarr: This check handles local file specifications authored prior to https://github.com/cake-build/cake/commit/0488e2eff24f86240684216fd2e34a470745b721 and simple glob patterns (eg: *.cs, scripts/*.cs). It doesn't handle more complex globs such as *.{cs,kake}.

My first thought was to use a FileExists check but this class doesn't have access to IEnvironment or any other mechanism for performing that check.

mythgarr:
@patriksvensson What is the correct process for getting this merged? It's been several months without a response, do I need to coordinate directly with a member of the cake team?

devlead:
@mythgarr your changes have been merged, thanks for your contribution 👍

Issue #2116, opened 2018-04-05
Loading Newtonsoft.Json in Cake.CoreCLR throws during assembly loading


daveaglick:
### What You Are Seeing?

Loading Newtonsoft.Json in Cake.CoreCLR throws during assembly loading.

### What is Expected?

Rainbows and sunshine.

### What version of Cake are you using?

Cake.CoreCLR 0.26.1

### Are you running on a 32 or 64 bit system?

64

### What environment are you running on?  Windows? Linux? Mac?

Windows

### How Did You Get This To Happen? (Steps to Reproduce)

Run the following `build.cake`:

```
#addin nuget:?package=Newtonsoft.Json&version=11.0.2

Task("Default")
    .Does(() =>
    {
    });

RunTarget("Default");
```

Raw assembly loading also fails:

```
#r "E:\Code\discoverdotnet\tools\Addins\Newtonsoft.Json.11.0.2\lib\netstandard2.0\Newtonsoft.Json.dll"

Task("Default")
    .Does(() =>
    {
    });

RunTarget("Default");
```

### Output Log

```
E:\Code\discoverdotnet>build -target preview -verbosity diagnostic
  Writing C:\Users\dglick\AppData\Local\Temp\tmp390D.tmp
info : Adding PackageReference for package 'cake.coreclr' into project 'E:\Code\discoverdotnet\tools\build.csproj'.
log  : Restoring packages for E:\Code\discoverdotnet\tools\build.csproj...
info : Package 'cake.coreclr' is compatible with all the specified frameworks in project 'E:\Code\discoverdotnet\tools\build.csproj'.
info : PackageReference for package 'cake.coreclr' version '0.26.1' updated in file 'E:\Code\discoverdotnet\tools\build.csproj'.
Module directory does not exist.
NuGet.config not found.
Analyzing build script...
Analyzing E:/Code/discoverdotnet/build.cake...
Processing build script...
Installing addins...
Found package 'Newtonsoft.Json 11.0.2' in 'E:/Code/discoverdotnet/tools/Addins'.
Package Newtonsoft.Json.11.0.2 has already been installed.
Successfully installed 'Newtonsoft.Json 11.0.2' to E:/Code/discoverdotnet/tools/Addins
Executing nuget actions took 35.23 ms
The addin Newtonsoft.Json will reference Newtonsoft.Json.dll.
[...]

PR #2597, opened 2019-08-15
Updating Newtonsoft.Json nuget package version to latest stable

pritchums:
Using the * notation for version spec (latest stable version).
This will bring in the latest stable version at compile time, and reference that.
Related to #2116, which is not resolved on dotnet core, and causes exceptions when referencing Newtonsoft.Json.

### DIFF
--- a/src/Cake.Core/Cake.Core.csproj
+++ b/src/Cake.Core/Cake.Core.csproj
@@ -16,7 +16,7 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
     <PackageReference Include="Microsoft.Win32.Registry" Version="4.4.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.4" />
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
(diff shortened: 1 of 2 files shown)


gep13:
@pritchums thank you for taking the time to raise this PR, however, at this time, I don't think we will be able to accept this contribution.  

One of the things that we strongly recommend when using Cake itself is to always pin to a specific version of Cake and any addins/tools/modules that you are using.  Doing this, means that you have the most repeatable/reliable/resilient build possible, as upstream changes to anything doesn't have an impact on your own build.  

We follow this same advice in our own build, and as such, using the wildcard to always pull in the latest stable version of this PackageReference violates that philosophy since one build one day might succeed, and the same build the next might fail, due to upstream change.

My suggestion would be to raise an issue discussing the specific problem that you are facing, so that we can agree a path forward.

pritchums:
Thanks @gep13 - will definitely make sure I raise an issue first next time!

Accept your point on the use of "*" to specify the version - I did erm and err about that, but should have discussed first.

I've raised #2606 to cover the underlying issue. I believe the fix is pretty much what is in this PR, albeit with specifying the version explicitly to 12.0.2. I can confirm this resolves the underlying issue on dotnet core.

PR #2598, opened 2019-08-15
Updating netcoreapp to 2.1

pritchums:
Resolves an issue I was observing when trying to reference System.ComponentModel.Annotations 4.5.0 under dotnet core (could not load assembly).
This will match the netcoreapp version used for Cake.Tool.csproj (the cake dotnet tool).

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -179,7 +179,7 @@ Task("Run-Unit-Tests")
         // .NET Core
         DotNetCoreTest(project.FullPath, new DotNetCoreTestSettings
         {
-            Framework = "netcoreapp2.0",
+            Framework = "netcoreapp2.1",
             NoBuild = true,
             NoRestore = true,
             Configuration = parameters.Configuration,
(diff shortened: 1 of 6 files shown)


gep13:
@pritchums thank you for taking the time to make this contribution, we really appreciate it, however, at this time, I don't think we will be considering making this change.

The reason for this is that this would be a breaking change for anyone who is currently using Cake.CoreClr, and we (the Cake Team) don't feel that there is a compelling reason to do this.

Instead, our recommendation would be for you to use Cake.Tool, which already targets 2.1, for your builds.  Have you tried using this?

For now, I am going to close this PR, and I would suggest that if you want to continue down this path, that you create an issue discussing the exact problem that you are facing, and we can discuss a path forward.

pritchums:
Thanks @gep13. Using the dotnet core tool does seem to work, and doesn't cause the System.ComponentModel.Annotations issue I highlighted. Will make sure I raise an issue first next time!

PR #2564, opened 2019-06-10
GH2275: Correctly reference runtime libraries from addins

mholo65:
- Fixes #2275

~~Don't know if this is the right direction to go. Basically we *should* load the runtime libs based on the current RID if running .NET Core. Problem is there's a whole lot of them (see [list or runtimes](https://github.com/dotnet/corefx/blob/7e56d8bd8b224d834b782a98bfd7035f71c3f547/pkg/Microsoft.NETCore.Platforms/runtime.json)  and [docs](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog)).~~

~~@daveaglick do you have anything to add? Maybe we should consult our friends at NuGet?~~

Went with the route of correctly findint runtime assemblies and loading them, including native libraries.

PR #2604, opened 2019-08-17
GH2603: Added .NET Core 3 moniker to Cake.Tool

### LINKED ISSUE #2603: Add .NET Core 3 to Cake.Tool update to .NET Core 3 SDK
* Add netcoreapp3.0 moniker to Cake.Tool
* Execute unit tests under netcoreapp3.0
* Update SDK to .NET Core 3

devlead:
(GH-2603) Added .NET Core 3 moniker to Cake.Tool
* Update SDK to .NET Core 3 (3.0.100)
* Add netcoreapp3.0 moniker to Cake.Tool
* Add netcoreapp3.0 moniker to Unit tests
* Execute unit tests under netcoreapp3.0
* Consolidate Cake.Tool.csproj into Cake.csproj (sorts VS2019 rebuild issues)
* Refactor Runtime Framework detection
* Add runtime compiler constants
* Microsoft.CodeAnalysis.CSharp.Scripting 3.3.1
* Remove mono workaround
* Add .NET Core rollForward policy
* Build script cleanup
* Update NuGet command line to 5.2.0 (5.3.0 not blessed yet)
* Update NuGet client libs to 5.3.0
* Temp fix for rollforward test
* fixes #2603


src/Shared.msbuild
```diff
@@ -38,6 +39,10 @@
     <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
     <DebugType>portable</DebugType>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.0' ">
```
mholo65: Can we combine all these above?

devlead:
> I think this PR should include the following, if it’s merged before my PR https://github.com/cake-build/cake/pull/2564/files#r314080758

Will take a look at that, code seems to work but, fails for unit tests as xunit isn't compiled (which is entry assembly) doesn't have to be compiled for same version as Cake.dll, so might need to look at executing runtime another way.

devlead:
@mholo65 did a quick PoC detecting running and not compiled runtime.  And that works...a little too well, because unit tests fail if not all runtimes installed, as 2.0 will run on 2.1,  2.1 will run on 3.0, etc.

Taking entry attribute won't work in unit tests as entry assembly is xunit which is netcoreapp 1.0

So while it probably works correctly at runtime, testing "expects" are calculated at compile time, so this is interesting.

At same time for integration tests it's not wrong to test on all supported runtimes.

devlead:
@cake-build/cake-team please review, there's a unit test that smells, but other than that most kinks have been ironed out to run and compile on .NET Core 3.

build.ps1
```diff
@@ -50,6 +50,11 @@ if ($PSVersionTable.PSEdition -ne 'Core') {
 # INSTALL .NET CORE CLI
 ###########################################################################
 
+$env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
+$env:DOTNET_CLI_TELEMETRY_OPTOUT=1
+$env:DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX=1
```
gep13:
```suggestion
$env:DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX=2
```

devlead: Fixed.

src/Shared.msbuild
```diff
@@ -6,38 +6,32 @@
     <Authors>Patrik Svensson, Mattias Karlsson, Gary Ewan Park, Alistair Chapman, Martin Björkström, Dave Glick, Pascal Berger and contributors</Authors>
     <Company>Patrik Svensson, Mattias Karlsson, Gary Ewan Park, Alistair Chapman, Martin Björkström, Dave Glick, Pascal Berger and contributors</Company>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <PackageIconUrl>https://raw.githubusercontent.com/cake-build/graphics/master/png/cake-medium.png</PackageIconUrl>
+    <PackageIcon>cake-medium.png</PackageIcon>
```
gep13: From the experiments that @Jericho has done, I believe that NuGet.org will strip the icon out, as it isn't currently supported there.  Can you confirm @Jericho?

devlead: .NET Core 3 requires it though or you get an compiler error.
```diff
- error NU5048: The 'PackageIconUrl'/'iconUrl' element is deprecated.
- Learn more at https://aka.ms/deprecateIconUrl
```

gep13: Hmm, interesting.  So I wonder if there is going to be an update to NuGet.org ahead of the RTM of .Net Core 3.

gep13: Also, was this previously a warning?  Or has it jumped straight to error?

devlead: Well currently we treat warnings as errors.

gep13: Oh, I see.  Ok, so that changes things :smile:

Jericho: I confirm that NuGet.org strops our embedded icons. I uploaded a package with such icon and you can see here that the icon is now missing: https://www.nuget.org/packages/Cake.Email.Common/

devlead: Looking at NuGet/NuGetGallery/issues/6907 it looks like it's in milestone [S159 - 2019.09.16](https://github.com/NuGet/NuGetGallery/milestone/130) which is due by October 04, 2019 which feels late as .NET Core 3 is due by next week.

devlead: @Jericho @gep13 Just pushed https://www.nuget.org/packages/LitJson/0.14.1 with PackageIcon and it worked just fine.

Exchanged emails with nuget team and there is no `<icon>` element in nuspec in Cake.Email.Common 0.4.0.
Issue with Cake.Email.Common 0.4.0 is likely not building using .NET Core 3.0 SDK

Can't find any global.json, .NET Core SDK install in build.ps1/build.sh nor appveyor.yml (which seems to use image: Visual Studio 2017), so can't find anything indicating it using newer required SDK.

Jericho: @devlead You're probably right: when I built on my machine (where .net 3.0 is available), the generated package included the icon but when built on [AppVeyor](https://ci.appveyor.com/project/cakecontrib/cake-email-common/builds/27254454/artifacts), the icon was stripped and therefore the package published to NuGet was missing an icon. I incorrectly assumed that NuGet had stripped the icon.

devlead: That's one use case for having a `global.json` file, so same SDK is used locally and CI.

devlead:
@cake-build/cake-team PR rebased, updated to RTM .NET Core SDK 3.0.100 and Microsoft.CodeAnalysis.CSharp.Scripting 3.3.1

src/Cake.Core.Tests/Unit/CakeRuntimeTests.cs
```diff
@@ -26,7 +26,35 @@ public void Should_Return_Correct_Result_For_CoreClr()
...
+                    Assert.Equal(".NETCoreApp,Version=v" +
+
+#if NETCOREAPP2_0
+                    "2.0",
+#elif NETCOREAPP2_1
+                    "2.1",
+#elif NETCOREAPP2_2
+                    "2.2",
+#elif NETCOREAPP3_0
+                                 "3.0",
```
patriksvensson: Indentation

src/Cake.Core.Tests/Unit/CakeRuntimeTests.cs
```diff
@@ -26,7 +26,35 @@ public void Should_Return_Correct_Result_For_CoreClr()
...
+#endif
+                        framework.FullName);
+                }
+                catch
+                {
+                    // Temp fix for if only netcore3 installed
+                    if (framework.FullName != ".NETCoreApp,Version=v3.0")
```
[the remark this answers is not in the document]
devlead: Fixed, but would like to refactor that test, perhaps make runtime detection more "mockable" but maybe that could be future improvement.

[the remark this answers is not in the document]
patriksvensson: Yes, that would be good. Agree to push it for now.

src/Cake.Core/Polyfill/EnvironmentHelper.cs
```diff
@@ -119,7 +123,28 @@ public static Runtime GetRuntime()
...
+
+            var assemblyPath = typeof(System.Runtime.GCSettings)?.GetTypeInfo()?.Assembly?.CodeBase;
+            if (string.IsNullOrEmpty(assemblyPath))
+            {
+                return NetStandardFramework;
+            }
+
+            const string microsoftNetCoreApp = "Microsoft.NETCore.App";
+            var runtimeBasePathLength = assemblyPath.IndexOf(microsoftNetCoreApp) + microsoftNetCoreApp.Length + 1;
+            var netCoreAppVersion = string.Concat(assemblyPath.Skip(runtimeBasePathLength).Take(3));
```
devlead: > Could this be explained in a comment? Where is the magic number 22 coming from?

@mholo65 better?

patriksvensson:
Merged! Thank you for your contribution. Much appreciated! 👍

Issue #2585, opened 2019-07-26
Cake.Tool - How in the world do I run a specific task?


VictorioBerra:
```
PS C:\Source\Templates> dotnet cake --help

Usage: Cake.exe [script] [--verbosity=value]
                [--showdescription] [--dryrun] [..]

Example: Cake.exe
Example: Cake.exe build.cake --verbosity=quiet
Example: Cake.exe build.cake --showdescription

Options:
    --verbosity=value    Specifies the amount of information to be displayed.
                         (Quiet, Minimal, Normal, Verbose, Diagnostic)
    --debug              Performs a debug.
    --showdescription    Shows description about tasks.
    --showtree           Shows the task dependency tree.
    --dryrun             Performs a dry run.
    --exclusive          Execute a single task without any dependencies.
    --bootstrap          Download/install modules defined by #module directives
    --version            Displays version information.
    --info               Displays additional information about Cake execution.
    --help               Displays usage information.
```

I see no way here on how I can specify the task to run.

Not mentioned here either https://cakebuild.net/blog/2018/08/cake-v0.30.0-released

VictorioBerra:
I just answered my question, I had to open the build.ps1 file and find the `-target="whatever"` which worked. Should this be documented somewhere at least?

Insire:
it is [here](https://cakebuild.net/docs/fundamentals/running-targets)

VictorioBerra:
@Insire Yes but the dotnet CLI global tool Cake.Tool bootstrap wrapper for cake does not have any documentation on those.

Roadrunner67:
I've just made  a PR #2605

Roadrunner67:
@VictorioBerra The answer you should have had to your issue post is something like this:

Target is a term unknown to Cake itself.

It is something we usually build into our Cake scripts, beginning with:
`var target = Argument("target", "Default");
`
..and ending with:

`RunTarget(target);
`
So you can choose the target from outside by adding -target=value to your command line.
If you don't, the target 'Default' will be executed (where you can place your own instruction to the user to select a specific task, if there is no logical default).

build.ps1/build.sh should only in very rare cases need modification.

PR #2605, opened 2019-08-18
GH2585: Helpcommand includes help about Target

Roadrunner67:
Usage: Cake.exe [script] **[--target=value]** [--verbosity=value]
                [--showdescription] [--dryrun] [..]

Example: Cake.exe
Example: Cake.exe build.cake --verbosity=quiet
Example: Cake.exe build.cake --showdescription

Options:
    **--target=value       Specifies the target to build. Surround with quotes if target contains special chars.**
    --verbosity=value    Specifies the amount of information to be displayed.
                         (Quiet, Minimal, Normal, Verbose, Diagnostic)
    --debug              Performs a debug.
    --showdescription    Shows description about tasks.
    --showtree           Shows the task dependency tree.
    --dryrun             Performs a dry run.
    --exclusive          Execute a single task without any dependencies.
    --bootstrap          Download/install modules defined by #module directives
    --version            Displays version information.
    --info               Displays additional information about Cake execution.
    --help               Displays usage information.

### DIFF
--- a/src/Cake/Commands/HelpCommand.cs
+++ b/src/Cake/Commands/HelpCommand.cs
@@ -32,7 +32,7 @@ namespace Cake.Commands
             else
             {
                 _console.WriteLine();
-                _console.WriteLine("Usage: Cake.exe [script] [--verbosity=value]");
+                _console.WriteLine("Usage: Cake.exe [script] [--target=value] [--verbosity=value]");
                 _console.WriteLine("                [--showdescription] [--dryrun] [..]");
                 _console.WriteLine();
                 _console.WriteLine("Example: Cake.exe");
@@ -40,6 +40,7 @@ namespace Cake.Commands
                 _console.WriteLine("Example: Cake.exe build.cake --showdescription");
                 _console.WriteLine();
                 _console.WriteLine("Options:");
+                _console.WriteLine("    --target <TARGET>    Target task to invoke. Script must support this explicitly.");
                 _console.WriteLine("    --verbosity=value    Specifies the amount of information to be displayed.");
                 _console.WriteLine("                         ({0})",
                     string.Join(", ", Enum.GetNames(typeof(Verbosity))));


gep13:
@Roadrunner67 thank you for submitting this PR, we really appreciate your contribution.

I have one problem with this addition, and I would like to discuss it further...

Cake.exe, Cake.CoreClr.dll, and Cake.Tool don't have a `--target` argument.  By this, I mean that nothing in the internals of the Cake Engine will do anything with that argument.  Unlike the other arguments, which the Cake Engine will do something with directly, like show the help menu, or include descriptions for tasks, etc.  The `--target` argument is a convention, which needs to be implemented within your Cake script, which is ultimately either passed to the `RunTarget` method, or not.

For example, it is perfectly valid to have a Cake script that *doesn't* have a target argument, in which case, this addition to the help menu wouldn't do anything.

As a result, I have a problem with adding this directly into the help menu for Cake.

I agree that we might need to better document this convention "somewhere" but I don't think this is the right place.

@cake-build/cake-team what are your thoughts?

Roadrunner67:
@gep13 You are right, it is completely a conventional thing, that we divide our Cake scripts into tasks.

I think on the other hand, that the answer given to VictorioBerra in issue #2585 was less than optimal, as he also seems to be unaware about this.

devlead:
@Roadrunner67 @gep13 IMHO I think target is such a common parameter so it would be good if it was documented. That said it must be clear that it's up to the script to support this. i.e something like.
`--target <TARGET>  Target task to invoke. Script must support this explicitly`

Also we don't need to have quotes mentioned explicitly as this is the same for all shell commands.

Roadrunner67:
> @Roadrunner67 @gep13 IMHO I think target is such a common parameter so it would be good if it was documented. That said it must be clear that it's up to the script to support this. i.e something like.
> `--target <TARGET> Target task to invoke. Script must support this explicitly`
> 
> Also we don't need to have quotes mentioned explicitly as this is the same for all shell commands.

@devlead @gep13 Tell me if you want me to change this in my PR, or should I just close it.

patriksvensson:
@devlead @gep13 Perhaps I should add this to my CLI PR?

gep13:
>@devlead said...
IMHO I think target is such a common parameter so it would be good if it was documented. 

The ideal would be if it was _only_ shown if the Cake script expected a target argument.  However, the same could be said about any "expected" arguments from the Cake script.

@patriksvensson would your CLI PR be able to generate automatically some additional documentation based on first inspecting the cake script for any Argument aliases?

patriksvensson:
@gep13 That would be difficult since the help arguments are displayed before any cake script or any other variables have been specified. Let me think about it a little.

devlead:
> > @Roadrunner67 @gep13 IMHO I think target is such a common parameter so it would be good if it was documented. That said it must be clear that it's up to the script to support this. i.e something like.
> > `--target <TARGET> Target task to invoke. Script must support this explicitly`
> > Also we don't need to have quotes mentioned explicitly as this is the same for all shell commands.
> 
> @devlead @gep13 Tell me if you want me to change this in my PR, or if I should just close it.

@Roadrunner67 yes please, make that change, and we can do a more refined improvement later.

devlead:
@Roadrunner67 your changes have been merged, thanks for your contribution 👍

Issue #2623, opened 2019-09-20
DotNetCoreTestSettings Can Have Multiple Logger's


RehanSaeed:
With .NET Core 3.0 and with the latest `Microsoft.NET.Test.Sdk` `16.3.0` NuGet package installed in your test project, you can now create HTML reports as well as TRX XML files like so:

```bash
dotnet test --logger trx;LogFileName=foo.trx --logger html;LogFileName=foo.html
```

The [DotnetCoreTestSettings.Logger](https://github.com/cake-build/cake/blob/e3550d249da09388c573e5af6206bb98ebf79b8f/src/Cake.Common/Tools/DotNetCore/Test/DotNetCoreTestSettings.cs#L35) property is currently a `string` and needs to be an `IEnumerable<string>`.

devlead:
Fixed by #2867

PR #2624, opened 2019-09-25
GH2623: Make Logger a collection

RehanSaeed:
Fixes https://github.com/cake-build/cake/issues/2623

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Test/DotNetCoreTesterTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Test/DotNetCoreTesterTests.cs
@@ -115,7 +115,7 @@ namespace Cake.Common.Tests.Unit.Tools.DotNetCore.Test
                 fixture.Settings.Settings = "./demo.runsettings";
                 fixture.Settings.Filter = "Priority = 1";
                 fixture.Settings.TestAdapterPath = @"/Working/custom-test-adapter";
-                fixture.Settings.Logger = new string[] { @"trx;LogFileName=/Working/logfile.trx" };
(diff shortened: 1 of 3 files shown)


RehanSaeed:
@devlead What's the process for breaking changes?

devlead:
@RehanSaeed haven't reviewed this PR in depth yet, but in general we try to obsolete stuff with warning one version,  with error next version and remove after that.

One way to not break could be to add a new property called Loggers concat it with Logger, obsolete Logger with warning.


But only took a quick look at the code, could be Loggers should be more typed i.e. FilePath and a trx/html/etc be a report type enum of some kind.

gitfool:
Turns out I need this for multiple `Loggers` too, and now I think I should make the recently added `Collector` plural as well, before the next release though to avoid the breaking change dance.

Is `ICollection<string>` the preferred property type - which would play nicely with C# collection initializers?

Finally, there's a recent addition to add [`-e|--environment`](https://github.com/dotnet/sdk/blob/7d267abebfb4ff8b970158e38cd30045da85bd05/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs#L37), which will also be accepted multiple times by `dotnet test`, so I'd add support for that the same way.

azchohfi:
@RehanSaeed what is the status on this PR? The Logger**s** approach seems the best so far.
I'm up for helping get this to the finishing line. Let me know if you are doing the work or if you want me to help here.

RehanSaeed:
I got snowed under by other work. Please feel free to take over this PR or raise a new one.

gitfool:
@RehanSaeed @azchohfi I've submitted a new PR that covers this (multiple loggers) and multiple collectors.

devlead:
Fixed by #2867

PR #2639, opened 2019-10-15
GH2637 GH2638: Update CreateAssemblyInfo alias

### LINKED ISSUE #2637: CreateAssemblyInfo alias creates invalid file when using .vb
When running the following cake script:

```
Task("Default")
  .Does(() =>
{
  var file = "./SolutionInfo.vb";
var version = "0.0.1";
var buildNo = "123";
var semVersion = string.Concat(version + "-" + buildNo);
CreateAssemblyInfo(file, new AssemblyInfoSettings {
    Product = "SampleProject",
    Version = version,
    FileVersion = version,
    InformationalVersion = semVersion,
    Copyright = string.Format("Copyright (c) Contoso 2014 - {0}", DateTime.Now.Year),
    ComVisible = true
});
});

RunTarget("Default")
```

The following file is generated:

```
'------------------------------------------------------------------------------
' <auto-generated>
'     This code was generated by Cake.
' </auto-generated>
'------------------------------------------------------------------------------
Imports System.Reflection
Imports System.Runtime.InteropServices

<Assembly: AssemblyProduct("SampleProject")>
<Assembly: AssemblyVersion("0.0.1")>
<Assembly: AssemblyFileVersion("0.0.1")>
<Assembly: AssemblyInformationalVersion("0.0.1-123")>
<Assembly: AssemblyCopyright("Copyright (c) Contoso 2014 - 2019")>
<Assembly: ComVisible(true)>
```

This is invalid, as this should be `True` not `true` for the ComVisible attribute.

### LINKED ISSUE #2638: CreateAssemblyInfo alias should allow creation of custom boolean attributes and also empty attributes
Currently, it is possible to use the CustomAttributes collection of the AssemblyInfoSettings class to create custom attributes that are not covered in the base AssemblyInfoSettings class.  However, it is not currently possible to create a custom attribute that looks like the following:

```
[assembly: CustomAttributeA(false)]
```

or like this:

```
[assembly: CustomAttributeB()]
```

This is due in part to the fact that the CreateAssemblyInfo alias was never intended to cover all bases, however, I believe that this scenario is common enough that it should be possible.

This is NOT related to this issue: https://github.com/cake-build/cake/issues/1027 which is actually looking for a new alias `ModifyAssemblyInfo`, which knows and understands all the attributes in the assemblyinfo file.  This issue is constrained only to the CreateAssemblyInfo alias.

gep13:
@cake-build/cake-team would be good to get some thoughts on this...

Fixes #2637 and fixes #2638

### DIFF
--- a/src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
+++ b/src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -16,6 +16,8 @@ namespace Cake.Common.Solution.Project.Properties
         private readonly Dictionary<string, string> _metadatattributes;
         private readonly HashSet<string> _namespaces;
         private readonly HashSet<string> _internalVisibleTo;
+        private readonly string _trueStringValue;
+        private readonly string _falseStringValue;
 
         public IDictionary<string, string> Attributes => _dictionary;
 
@@ -27,13 +29,16 @@ namespace Cake.Common.Solution.Project.Properties
 
         public ISet<string> InternalVisibleTo => _internalVisibleTo;
 
-        public AssemblyInfoCreatorData(AssemblyInfoSettings settings)
+        public AssemblyInfoCreatorData(AssemblyInfoSettings settings, bool isVisualBasicAssemblyInfoFile)
         {
             _dictionary = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             _customAttributes = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             _metadatattributes = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             _namespaces = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             _internalVisibleTo = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            _falseStringValue = isVisualBasicAssemblyInfoFile ? "False" : "false";
+            _trueStringValue = isVisualBasicAssemblyInfoFile ? "True" : "true";
 
             // Add attributes.
             AddAttribute("AssemblyTitle", "System.Reflection", settings.Title);
@@ -82,7 +87,7 @@ namespace Cake.Common.Solution.Project.Properties
         {
             if (value != null)
             {
-                AddAttributeCore(Attributes, name, @namespace, value.Value ? "true" : "false");
+                AddAttributeCore(Attributes, name, @namespace, value.Value ? _trueStringValue : _falseStringValue);
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
```diff
@@ -94,9 +99,17 @@ private void AddAttribute(string name, string @namespace, string value)
...
 
-        private void AddCustomAttribute(string name, string @namespace, string value)
+        private void AddCustomAttribute(string name, string @namespace, object value)
         {
-            if (value != null)
+            if (value is bool)
+            {
+                AddAttributeCore(CustomAttributes, name, @namespace, (bool)value ? _trueStringValue : _falseStringValue);
+            }
+            else if (value == null || string.IsNullOrEmpty(value as string))
```
patriksvensson: This should be `else if (value == null || (value is string stringValue && string.IsNullOrEmpty(stringValue))` otherwise we will never reach the `else` clause.

devlead: IMHO a pattern switch would be cleaner something like
```csharp
        private void AddCustomAttribute(string name, string @namespace, object value)
        {
            var attributeValue = AttributeValueToString(value);

            AddAttributeCore(CustomAttributes, name, @namespace, attributeValue);
        }

        private string AttributeValueToString(object value)
        {
            switch (value)
            {
                case null:
                {
                    return string.Empty;
                }
                case bool boolValue:
                {
                    return boolValue ? _trueStringValue : _falseStringValue;
                }
                case string stringValue:
                {
                    return stringValue == string.Empty
                        ? string.Empty
                        : string.Concat("\"", value, "\"");
                }
                default:
                {
                    return Convert.ToString(value, CultureInfo.InvariantCulture);
                }
            }
        }
```

also should non strings be quoted? I.e. int/long/etc? In example above I to string object invariant.  To be more robust should we escape `"` like `\"`?

devlead: Also unrelated `name` in `AddMetadataAttribute(string name, string @namespace, string key, string value)` doesn't seem to be use, it's private so wouldn't break anything to remove it it it's indeed not used.

gep13: I agree, a switch statement here makes a lot of sense.  I will change it up to use this.  Honestly, I am not sure regarding handling for int/longs, etc.  As I mentioned in the commits/comments, this change wasn't an attempt to handle all conditions, but rather the ones that I know currently are prevented.  I would suggest that we handle any edge cases on a case by case basis.  Thoughts?

gep13: We can create another issue for the name property, as this is not related to this PR.

devlead: >  As I mentioned in the commits/comments, this change wasn't an attempt to handle all conditions, but rather the ones that I know currently are prevented. I would suggest that we handle any edge cases on a case by case basis. Thoughts?

Then I think it makes more sense in default switch case to either not quote and use invariant as my example above, or throw a not supported exception.

gep13:
@cake-build/cake-team this is ready for another review.

devlead:

Issue #2533, opened 2019-04-23
Setting SecurityRules with CreateAssemblyInfo


dnperfors:
### What You Are Seeing?
When I want to use the CreateAssemblyInfo to set the `[assembly: SecurityRules(SecurityRuleSet.Level2)]` attribute (which is [recommended](docs.microsoft.com/en-us/dotnet/framework/misc/security-transparent-code-level-2?view=netframework-4.7.2) to do by Microsoft) I get a compile error that I can't convert a string to a SecurityRuleSet.

The code I use is:
```
    CreateAssemblyInfo("./code/SharedAssemblyInfo.cs", new AssemblyInfoSettings
    {
        CustomAttributes = new List<AssemblyInfoCustomAttribute>
        {
            new AssemblyInfoCustomAttribute { Name = "SecurityRules", NameSpace = "System.Security", Value ="SecurityRuleSet.Level2" }
        },
    });
```

And generates:
```
[assembly: SecurityRules("SecurityRuleSet.Level2")]
```

### What is Expected?
I would expect to be able to give any type of value to AssemblyInfoCustomAttribute, for example a generic:
```
    CreateAssemblyInfo("./code/SharedAssemblyInfo.cs", new AssemblyInfoSettings
    {
        CustomAttributes = new List<AssemblyInfoCustomAttribute>
        {
            new AssemblyInfoCustomAttribute<SecurityRuleSet> { Name = "SecurityRules", NameSpace = "System.Security", Value =SecurityRuleSet.Level2 }
        },
    });
```

And generates:
```
[assembly: SecurityRules(SecurityRuleSet.Level2)]
```

### What version of Cake are you using?
0.33.0

### Are you running on a 32 or 64 bit system?
64 bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

devlead:
Fixed by  #2648

PR #2648, opened 2019-10-26
GH2533: Add new RawValue parameter to skip attribute value quoting

agc93:
This PR should solve #2533 in a more generic way. Rather than hardcoding for `SecurityRules` this adds opt-in behaviour to skip attribute quoting.

I'm not 100% sure on the whole "raw" naming, but I couldn't think of anything better. Other option would be to more explicit and use something like `NoQuote`.

### DIFF
--- a/src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
+++ b/src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
@@ -72,7 +72,7 @@ namespace Cake.Common.Solution.Project.Properties
             {
                 foreach (var item in settings.CustomAttributes.Where(item => item != null))
                 {
-                    AddCustomAttribute(item.Name, item.NameSpace, item.Value);
+                    AddCustomAttribute(item.Name, item.NameSpace, item.Value, item.UseRawValue);
                 }
             }
             if (settings.MetaDataAttributes != null)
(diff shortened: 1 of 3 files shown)


src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
```diff
@@ -123,7 +123,9 @@ private string AttributeValueToString(object value)
                 {
                     return stringValue == string.Empty
                         ? string.Empty
-                        : string.Concat("\"", value, "\"");
+                        : isRawValue
+                            ? value.ToString()
+                            : string.Concat("\"", value, "\"");
```
devlead: Currently quotes with in string would break output, wondering if we should do something like this?
```suggestion
                            : string.Concat("\"", stringValue.Replace("\"", "\\\""), "\""));
```

devlead:
Personally I think `RawValue` is fine, maybe it would be more clear with `UseRawValue`.

src/Cake.Common/Solution/Project/Properties/AssemblyInfoCreatorData.cs
agc93: That part was existing behaviour, but I think you're right that that's a better way of doing it. I'll add that push again.

devlead:

Issue #2498, opened 2019-03-13
ParseProject fails on an aboslute HintPath


veselfil:
### What You Are Seeing?

When attempting to parse a .csproj file which contains an absolute HintPath, Cake exits with the following exception: `Cannot combine a directory path with an absolute file path.`

### What is Expected?

The project should be parsed successfully, never had any issues with it in Visual Studio.

### What version of Cake are you using?

0.32.1, also tried with 0.30.0.

### Are you running on a 32 or 64 bit system?

64-bit.

### What environment are you running on?  Windows? Linux? Mac?

Windows 10 Pro, all updates installed.

### Are you running on a CI Server?  If so, which one?

No.

### How Did You Get This To Happen? (Steps to Reproduce)

I'm not sure what you have to do to add a HintPath, but the library that did it for me is DotNetBar. I'm not the original author of the project, and he's not around anymore, so I cannot really give you more details as how it was added.

Here's a snippet from one of the failing projects:

```
<Reference Include="DevComponents.DotNetBar2 ...">
    <SpecificVersion>False</SpecificVersion>
    <HintPath>C:\Program Files (x86)\DotNetBar for Windows Forms\Signed\DevComponents.DotNetBar2.dll</HintPath>
</Reference>
```

Commenting out the HintPath line fixes the issue.

### Output Log

Here's the stack trace:

~~~sh
System.InvalidOperationException: Cannot combine a directory path with an absolute file path.
   v Cake.Core.IO.DirectoryPath.CombineWithFilePath(FilePath path)
   v Cake.Common.Solution.Project.ProjectParser.<>c__DisplayClass3_0.<Parse>b__32(<>f__AnonymousType16`2 <>h__TransparentIdentifier9)
   v System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   v System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   v System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   v Cake.Common.Solution.Project.ProjectParser.Parse(FilePath projectPath)
   v Submission#0.ParseProject(FilePath projectPath)
   v Submission#0.<<Initialize>>b__0_5()
~~~

gjonespf:
Had same on a project that added a reference to System.Management.Automation (needed for PowerShell automation).  Item group in the project looks like following:

```
  <ItemGroup>
    <Reference Include="System" />
    <Reference Include="System.Configuration.Install" />
[...]

PR #2649, opened 2019-10-26
GH2498: Preserves rooted path values in HintPath

agc93:
When parsing HintPath values, "rooted" path values will be preserved as-is.
Relative path values will still be combined with current dir.

Tests have also been updated but to be honest, those tests could probably be cleaned up a bit.


agc93:
Thought that seemed too simple. I'll update that test to use a Unix-y path.

devlead:
Relates to: #2498

devlead:

PR #2681, opened 2019-12-03
GH2682: Update to .NET Core SDK 3.1.100

devlead:
LTS Version fo .NET Core 3
https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1.0.md


devlead:
Exception message has changes if running under 3.1, will investigate failing integration test
```
Expected: ···alue cannot be null.\nParameter name: name
Actual:   ···alue cannot be null. (Parameter 'name')
```
Will also need to adjust GitHub action setup .net will only point to one version, see work arounds to get multiple versions side by side.

devlead:
This was a little more than an SDK upgrade so I raised #2682 which this PR fixes.

devlead:
@cake-build/cake-team this should be ready for review now.

PR #2689, opened 2019-12-16
(build) Update GitVersion to 5.1.2

devlead:


gep13:
@devlead does this bump address the issue that you mentioned previously regarding the asserted version numbers?

devlead:
@gep13 doesn't seem so, I'll investigate some more.

devlead:
Maybe related to this GitTools/GitVersion#1927

devlead:
Ideally PR branches shouldn't fail the build, would be nice if it generated same as on AppVeyor `0.36.0-PullRequest2689...` instead of `0.36.0-gitversion-5-1-2...`

gep13:
Agreed.

PR #2730, opened 2020-03-09
Better handling for GitVersion failure cases

michaelnoonan:
GitVersion was failing for our builds this week, but the reason was a mystery until we were able to dig into the GitVersion log file on a failed build. What made this really difficult is the actual error was masked by the GitVersionRunner.

The error according to Cake was `Expecting state 'Element'.. Encountered 'Text'  with name '', namespace ''.` This is accurate because the `DataContactJsonSerializer` is failing to deserialise what is actually an error message from `gitversion.exe`. However, it is ultimately misleading since the actual error message is masked by this exception.

We changed our process to call GitVersion twice:

1. Run GitVersion with `OutputType = GitVersionOutput.BuildServer` so it logs to stdout, including any error messages
2. Run GitVersion with `OutputType = GitVersionOutput.Json` to get the deserialised version information

This PR proposes a similar approach, but within the GitVersionRunner itself so others fall into the pit of success. It appears the intent of the code was to run GitVersion twice, but the deserialization exception prevents the intended behaviour.

Hope this helps!

### DIFF
--- a/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
+++ b/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
@@ -51,6 +51,8 @@ namespace Cake.Common.Tools.GitVersion
                 throw new ArgumentNullException(nameof(settings));
             }
 
+            Run(settings, GetArguments(settings));
+
             if (settings.OutputType != GitVersionOutput.BuildServer)
             {
                 var jsonString = string.Empty;
@@ -65,8 +67,6 @@ namespace Cake.Common.Tools.GitVersion
                     return (jsonSerializer.ReadObject(jsonStream) as GitVersionInternal)?.GitVersion;
                 }
             }
-
-            Run(settings, GetArguments(settings));
 
             return new GitVersion();
         }


gitfool:
@michaelnoonan this should improve slightly now that PR https://github.com/GitTools/GitVersion/pull/2161 has merged and released in GitVersion 5.2.1.

I certainly wouldn't want to call GitVersion twice, especially with noisy build server log output.

I was going to submit a followup PR to Cake that scrapes the `ERROR` line from the GitVersion when the process exits with non-zero exit code. This would show the relevant error without needing to increase Cake build verbosity or use any other tricks.

michaelnoonan:
Cool. I'm happy to leave this with you since it's a concrete example of where we ran into trouble. Calling gitversion twice is not terribly appealing, even though it will likely hit the gitversion caches and be reasonably fast.

gep13:
@michaelnoonan apologies for not getting back to you sooner about this PR.  Was there an issue opened for the work in this PR?  Our contributing guidelines mention that an issue should be created first.

devlead:
@michaelnoonan thanks for your contribution 👍 
I'm going a head and going to close this PR as it should be fixed by #2741, which should be released in Cake [0.38.0](https://github.com/cake-build/cake/milestone/65).

PR #2738, opened 2020-03-20
GH2733: Remove redundant process output checks

### LINKED ISSUE #2733: NuGetSetApiKey with Verbosity set to Quiet causes an exception.
### What You Are Seeing?
When calling NuGetSetApiKey with NuGetSetApiKeySettings where Verbosity is set to NuGetVerbosity.Quiet, I get an error message for the task and the script exits.
The error message say:
"SetApiKey returned unexpected response"

### What is Expected?
That the call should not throw an exception.

Tested with multiple versions of Cake (0.30.0, 0.34.x, 0.36.0 and 0.37.0) and Cake.Tool 0.36.0 and 0.37.0
Also tested with older versions of NuGet.CommandLine as tool in the script.

64 bit

Windows 10

No, local build, but I'm pretty sure I can reproduce on the CI server also

[build.cake.txt](https://github.com/cake-build/cake/files/4349904/build.cake.txt)
To reproduce with the attached build.cake script:
dotnet cake -Target="Quiet"
To test with Verbosity.Normal (the working one):
dotnet cake -Target="Normal"

### Output Log
[error.log](https://github.com/cake-build/cake/files/4349905/error.log)

gitfool:
Fixes #2733.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/NuGet/SetApiKey/NuGetSetApiKeyFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/NuGet/SetApiKey/NuGetSetApiKeyFixture.cs
@@ -15,17 +15,6 @@ namespace Cake.Common.Tests.Fixtures.Tools.NuGet.SetApiKey
         {
             ApiKey = "SECRET";
             Source = "http://a.com";
-
-            // Set the standard output.
-            ProcessRunner.Process.SetStandardOutput(new[]
-            {
-                string.Concat("The API Key '", ApiKey, "' was saved for '", Source, "'.")
-            });
-        }
-
-        public void GivenUnexpectedOutput()
-        {
-            ProcessRunner.Process.SetStandardOutput(new string[] { });
         }
 
         protected override void RunTool()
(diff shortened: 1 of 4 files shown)


devlead:
This is a change in behavior, could you add an integration test to ensure it still works as expected.
https://github.com/cake-build/cake/blob/develop/tests/integration/Cake.Common/Tools/NuGet/NuGetAliases.cake


Should do something like
1. have NuGet.config without key
1. Set key unencrypted
1. Read back xml
1. Assert value compare.

gitfool:
@devlead I added an integration test but note that I couldn't check the actual value since it's encrypted, so I just check that it exists.

Also, while testing I saw that the api key was indeed in the standard output(!), so I put back the code to redirect the standard output; the net difference then is that the output text is not checked for any particular text.

gitfool:
@devlead I had another look at how output redaction currently works. My newfound understanding is that it does actually work, I just missed the subtlety that it's done by [ProcessArgumentBuilder](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessArgumentBuilder.cs#L16) [FilterUnsafe](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessArgumentBuilder.cs#L84-L102) in conjunction with  [ProcessWrapper](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessWrapper.cs#L13) [GetStandardOutput](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/ProcessWrapper.cs#L95-L107). i.e. you still need to redirect standard output and the `ProcessWrapper` will then redact sensitive arguments in the process output as read and logged at diagnostic verbosity.

devlead:

Issue #2745, opened 2020-03-25
Extend XmlPeekSettings with FileShare.* option


Asemirski:
### What You Are Seeing?
Using xmlpeek from different builds at the same time causes "The process cannot access the file ecause it is being used by another process.

### What is Expected?
It is possible to work with file while cake xmlpeeks in there

### What version of Cake are you using?
0.37.0

### Are you running on a 32 or 64 bit system?
x64
### Are you running on a CI Server?  If so, which one?
Jenkins

### How Did You Get This To Happen? (Steps to Reproduce)
Run a large number of builds simultaneously.

devlead:
Fixed by #2747

PR #2747, opened 2020-03-29
GH2745: Change file share access to Read for XmlPeekAliases

Marusyk:
A small change in file-sharing access to avoid the error: "The process cannot access the file because it is being used by another process."

Related to #2745

Please review,
Thank you in advance


devlead:
@marusyk your changes have been merged, thanks for your contribution 👍

PR #2750, opened 2020-04-03
GH2749: Add ANSI terminal support to log

patriksvensson:
This commit adds support for ANSI escape codes in Cake's log on platforms that support it.


src/Cake.Core/CakeEnvironment.cs
```diff
@@ -104,23 +100,7 @@ public string GetEnvironmentVariable(string variable)
...
-                                _log.Warning("GetEnvironmentVariables() encountered duplicate for key: {0}, value: {1} (existing value: {2})",
-                                    key,
-                                    value,
-                                    existingValue);
-                            }
-                        }
-                        else
-                        {
-                            dictionary.Add(key, value);
-                        }
```
patriksvensson, line 122: Not possible to have duplicate environment variables, so removed this code. Unnecessary to require a log instance for the environment as well since it's one of those classes that shouldn't require anything that in theory might require itself.

devlead: It is possible as environment variables is just a string in some OS:s i.e. we had issue cake wouldn't start on systems with "corrupt" variables, caused by some APIs had different case sensitivity. So you could get duplicate keys between system/user/process env variables.

patriksvensson: Ok, I need to think about this a bit. Can't require environment for the logger when we require the logger for the environment.

devlead: I think this is a bit of an edge case perhaps just `return dictionary;` if found? And skip the logging.

patriksvensson: @devlead Can you elaborate what you mean with ”return the dictionary if found”?

src/Cake.Core/CakeEnvironment.cs
```diff
@@ -104,23 +100,7 @@ public string GetEnvironmentVariable(string variable)
...
-
+                        dictionary[(string)entry.Key] = entry.Value as string;
```
devlead, line 103:
```suggestion
                        var key = (string)entry.Key;
                        if (!dictionary.TryGetValue(key, out _))
                        {
                            dictionary.Add(key, entry.Value as string);
                        }

                        return dictionary;
```

src/Cake.Core/Diagnostics/Console/AnsiDetector.cs
```diff
@@ -0,0 +1,124 @@
...
+                new Regex("scoansi"), // SCO ANSI
+                new Regex("cygwin"), // Cygwin, MinGW
+                new Regex("linux"), // Linux console
+                new Regex("konsole"), // Konsole
+                new Regex("bvterm"), // Bitvise SSH Client
+            };
+        }
+
+        public static bool SupportsAnsi(ICakeEnvironment environment)
+        {
```
devlead, line 45: Github action doesn't setup a correct PTY but does support ansi, so should addd something like.
```suggestion
        {
            if (!string.IsNullOrEmpty(environment.GetEnvironmentVariable("GITHUB_ACTION")))
            {
                return true;
            }
```

patriksvensson: @devlead Is that the environment variable name or pseudo code?

patriksvensson: @devlead Nevermind, found it here: https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

patriksvensson:
@devlead Ready for review

devlead:
@patriksvensson interestingly there's something in this PR that makes integrations tests setup/teardown fail, that doesn't fail with other PRs.

patriksvensson:
@devlead Strange. It passes all integration tests locally. Will take a look.

devlead:
@patriksvensson yip not been able to repro myself yet.

devlead:
This is really weird,  it only fails on the Ubuntu GH agent, but it's only this PR that fails so it's something in this PR and that environment.

`ubuntu-latest` is supposed to be `Ubuntu 18.04`, but I can't repro in a standard Ubuntu 18.04 container image.

patriksvensson:
Will take a closer look at it

patriksvensson:
@devlead I've fixed the bug and should be ready for review now

devlead:
> 
> 
> @devlead I've fixed the bug and should be ready for review now

I'll merge once CI passes.

devlead:

PR #2753, opened 2020-04-04
GH2752: Fix precedence of VS2019 MSTest.exe to be higher than VS2017.

kcamp:
This re-orders the "Year and Edition" folder path segments of VS2019 to appear
first in the list as the newest VS version, which allows any alternative tool path
resolution to resolve latest first.  This follows the patterns established where the
resolution has historically resolved VS2015 over VS2013, and so on.

This also adds tests to explicitly exercise this evaluation.  If additional coverage is desired, I can add that, but I think that's probably superfluous at this point.  As always, any feedback or change requests are welcome 👍


devlead:

Issue #2152, opened 2018-05-13
try resolve vstest.console.exe before guessing it


paulomorgado:
Visual Studio Test Runner is not part of the [Visual Studio Build Tools](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools "Visual Studio Build Tools 2017 component directory") but is available on NuGet as [Microsoft.TestPlatform](https://www.nuget.org/packages/Microsoft.TestPlatform).

If the user installs the `Microsoft.TestPlatform`, that should be used instead of trying to guess the path.

mholo65:
Closed by #2151

Issue #2077, opened 2018-03-13
Locating the correct vstest.console.exe (VS2017)


Kemyke:
### What You Are Seeing?
there is a closed issue https://github.com/cake-build/cake/issues/1522 for locating the correct VS2017 `mstest.exe`. But that should be applied when locating vstest.console.exe.

### What is Expected?
If i am right the same modification should be applied to find the correct `vstest.console.exe` [here](https://github.com/cake-build/cake/blob/ad97ab26842fb417189ac36f248f99b14539520c/src/Cake.Common/Tools/VSTest/VSTestRunner.cs).

RaringCoder:
Microsoft have a package:

https://www.nuget.org/packages/Microsoft.TestPlatform/

For deploying the console for VSTest. Should be not be using this as the `#tool` directive for it? I'd rather download a pinned, isolated tool that hope for some console runner on the PC somewhere.

jnm2:
Yes!

jnm2:
Until Cake's VSTest converts to using the new, [officially-recommended](https://developercommunity.visualstudio.com/content/problem/183405/there-is-no-installer-package-id-which-signals-the.html#reply-185103) package https://www.nuget.org/packages/Microsoft.TestPlatform/, this is what I'm doing:

```cake
VSTest(asssemblyPath, new VSTestSettings
{
    // https://github.com/cake-build/cake/issues/2077
    #tool Microsoft.TestPlatform
    ToolPath = Context.Tools.Resolve("vstest.console.exe")
});
```

jnm2:
Ah, see @paulomorgado's https://github.com/cake-build/cake/issues/2152!

mholo65:
#2152 is now merged and functionality for resolving `VSTest` through `Microsoft.TestPlatform` will be available in 0.29.0.

jnm2:
Was this issue meant to be closed?

devlead:
Fixed by #2754

PR #2754, opened 2020-04-05
GH2077: Adds alternate tool path support to vstest.console.exe  for VS2017 + 2019

kcamp:
This addresses the outstanding issue identified in GH-2077.
While acknowledging that the better-supported solution was introduced with GH-2152, this gives the Cake implementation a more complete solution, especially for users that are constrained to the Visual Studio provided executable (i.e., build server, non-homogenized environments, etc.)

This introduces the "year and edition" path querying that is implemented in the `MSTestRunner`.

The one thing worth mentioning is that I am pretty sure the inclusion of `"15.0"` in the block below isn't quite right - the `2017/` paths added in this PR most likely covers the VS2017 ("15.0") case, but given the logic employed, the implementation is most likely fully backward-compatible, and I saw no pragmatic reason to alter the existing implementation.
```csharp
foreach (var version in new[] { "15.0",
```

Let me know if there's anything you'd like added or changed 👍 Thanks!


devlead:

PR #2758, opened 2020-05-04
GH2378: Enable SourceLink with embedded PDB

### LINKED ISSUE #2378: Publish Cake packages with symbols and EmbedAllSources or SourceLink
Publishing Cake packages with symbols and `EmbedAllSources` or [SourceLink](https://github.com/dotnet/sourcelink) would make step-debugging Cake scripts much more pleasant. 😉 

As an example, to build using `EmbedAllSources` I use the following:
```c#
Task("Build")
    .Does(() =>
{
    var msbuildSettings = new DotNetCoreMSBuildSettings
    {
        MaxCpuCount = Build.ToolSettings.BuildMaxCpuCount,
        TreatAllWarningsAs = Build.ToolSettings.BuildTreatWarningsAsErrors ? MSBuildTreatAllWarningsAs.Error : MSBuildTreatAllWarningsAs.Default
    }
        .WithProperty("EmbedAllSources", Build.ToolSettings.BuildEmbedAllSources.ToString().ToLower())
        .WithProperty("Version", Build.Version.AssemblySemVer)
        .WithProperty("FileVersion", Build.Version.AssemblySemFileVer)
        .WithProperty("InformationalVersion", Build.Version.InformationalVersion);
    var buildSettings = new DotNetCoreBuildSettings
    {
        Configuration = Build.Parameters.Configuration,
        MSBuildSettings = msbuildSettings
    };
    DotNetCoreBuild(Build.Files.Solution.FullPath, buildSettings);
}
```
Also, SourceLink is easily enabled for all modern projects in one place using [Directory.Build.props](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2017#directorybuildprops-and-directorybuildtargets):
```xml
<Project>
  <PropertyGroup>
    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
    <EmbedUntrackedSources>true</EmbedUntrackedSources>
    <PublishRepositoryUrl>true</PublishRepositoryUrl>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All"/>
  </ItemGroup>
</Project>
```
Note: this should be easy for new compact `csproj` files targeting .NET Core. I don't think this is possible for other targets, but that shouldn't affect making the debugging experience as good as possible for builds using Cake on .NET Core.

gitfool:
Fixes #2378.


gitfool:
@devlead "Azure Pipelines - Build Cake Ubuntu" failure is unrelated. Please retry.

devlead:
@gitfool kicked off a rerun of pipeline builds.

gitfool:
@devlead how do I access the pr build artifacts again?

gitfool:
Meanwhile, I downloaded and installed [Cake.Tool.0.38.0-gh2378-0001.nupkg](https://ci.appveyor.com/api/buildjobs/a306rt0gtkrpya8d/artifacts/Cake.Tool.0.38.0-gh2378-0001.nupkg) into a local repo, then I managed to successfully step debug into Cake using `dotnet cake -debug` and Visual Studio 2019! 😁

devlead:
@gitfool looks ok to me, but will have to take it for a spin, see Cake & CakeCoreClr packages changed too, have you tested `cake.exe --debug` and `cake.dll --debug` still works too without symbols packages?

gitfool:
@devlead yes, I just tested `cake.exe --debug` with [Cake.0.38.0-gh2378-0001.nupkg](https://ci.appveyor.com/api/buildjobs/p8h0aclrhm76ocwf/artifacts/Cake.0.38.0-gh2378-0001.nupkg) and it also **worked**:
```
nuget add "D:\Downloads\Cake.0.38.0-gh2378-0001.nupkg" -source D:\NuGet
nuget sources add -name local -source D:\NuGet -configfile NuGet.config
```
`NuGet.config`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <disabledPackageSources>
    <clear />
  </disabledPackageSources>
  <packageSources>
    <clear />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
    <add key="local" value="D:\NuGet" />
  </packageSources>
</configuration>
```
`packages.config`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Cake" version="0.38.0-gh2378-0001" />
</packages>
```
Not sure how to test with [Cake.CoreCLR.0.38.0-gh2378-0001.nupkg](https://ci.appveyor.com/api/buildjobs/p8h0aclrhm76ocwf/artifacts/Cake.CoreCLR.0.38.0-gh2378-0001.nupkg) as I've never used `Cake.CoreCLR` directly, but I'd expect it to work based on what I see in its NuGet package, which mirrors what you see with the `Cake` package:

image
i.e. the correct source links are in the embedded PDB. 😉

gitfool:
PS. don't worry about the Source Link warning re "contains untracked sources":

image

This [should be fixed](https://github.com/dotnet/sourcelink/issues/572#issuecomment-583548551) when dotnet sdk `3.1.3xx` drops; hopefully next week! 🤞

devlead:
Rebased against develop which should use 3.1.202 released yesterday.

gitfool:
Yeah I was hoping for 3.1.3xx. Nevertheless, the above warning is harmless, so can be ignored.

devlead:

gep13:
@gitfool follow up question about this...

Does it make sense to both embed the pdb and also to generate the snupkg file? If so, is this even possible? We were discussing this topic on my stream tonight, and thought I would ask since you did the work in this PR.  Thanks!

gitfool:
@gep13 my understanding is you'd choose one or the other, not both, but I'm not sure as I've never used snupkg files, preferring the embedded pdb approach.

Perhaps @clairernovotny could provide a more authoritative answer?

clairernovotny:
You don't need a snupkg if you use embedded pdb's

gep13:
@gitfool @clairenovotny thank you for the replies. We are just trying to decide what is the best plan for cake addins going forward. We push to other sources other than NuGet.org, and we thought embedding would make debugging possible in all scenarios, but then we also like the idea of the snupkg being available.

Is there any harm in doing both, and can they both be generated at once?

gitfool:
@gep13 I just watched your recent Twitch stream related to this. 😉

I have to admit I don't understand why you would want to create both an embedded `pdb` _and_ a `snupkg`. You should just recommend one approach for Cake and contribs. IMHO an embedded `pdb` is simpler for both producer and consumer, assuming you don't mind the extra size from the portable `pdb` (bearing in mind they're tiny compared to a Windows `pdb`, so this shouldn't really be an issue). Using an embedded `pdb` is also more flexible as it would support debugging prerelease packages on other servers, like Cake alpha releases. In these cases a `snupkg` is completely redundant as the debugger wouldn't even use it!?

The Cake VS Code extension config etc should support both of course, since the above would only be a recommendation.

gep13:
@gitfool thanks for checking out the stream!  Part of the reason is the visibility on NuGet.org.  If you look at Cake.FileHelpers:

https://www.nuget.org/packages/Cake.FileHelpers

You can see that there are symbols, and you can download them, if required.

If you embed the pdb into the nupkg, that visibility is lost.

I totally get that things are easier putting the pdb directly into the nupkg, and I suspect that this is the route that I will go, and recommend, however, if it doesn't harm anything, I think I would like to do both.

clairernovotny:
There’s really no point to the snupkg as @gitfool says. Nothing will use them.

PR #2769, opened 2020-05-17
Added options to skip the analysis of the output file

### LINKED ISSUE #2768: Add an option to skip the default warning/error output when running JetBrains command line tools.
In our current enviroment, we run the Jetbrains commandline tools to check our CI/CD builds for static code issues. After the analysis is finished, the current implementation of DupFinder and InspectCode in Cake always starts analysing the file and will output some of the issues to the console. There is no option disable this step.

In our case we use Cake.Issues and other cake addins to analyse the output file after the analysis is finished. We would like to skip the initial analysis of the output file by Cake, because we will take care of it on our own.

A new option for both commandline tools to skip the described behavior would be great.

janniksam:
Resolves #2768 

I've added an option to skip the built-in analysis part that is performed after the JetBrains command line tool is done. In our case we use another tool to do the analysis using a custom logic. That's why we do not want to output the "issues" directly.

### DIFF
--- a/src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
+++ b/src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
@@ -106,6 +106,6 @@ namespace Cake.Common.Tools.DupFinder
         /// Gets or sets a value indicating whether to skip analysis of the file
         /// that was output by the command line tool or not.
         /// </summary>
-        public bool DoNotAnalyseOutput { get; set; }
+        public bool SkipOutputAnalysis { get; set; }
     }
 }
--- a/src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
+++ b/src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
@@ -101,5 +101,11 @@ namespace Cake.Common.Tools.DupFinder
         /// Gets or sets a value indicating whether to throw an exception on finding duplicates
         /// </summary>
         public bool ThrowExceptionOnFindingDuplicates { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to skip analysis of the file
+        /// that was output by the command line tool or not.
+        /// </summary>
+        public bool DoNotAnalyseOutput { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Cake.Common/Tools/DupFinder/DupFinderSettings.cs
```diff
@@ -101,5 +101,11 @@ public sealed class DupFinderSettings : ToolSettings
         /// Gets or sets a value indicating whether to throw an exception on finding duplicates
         /// </summary>
         public bool ThrowExceptionOnFindingDuplicates { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to skip analysis of the file
+        /// that was output by the command line tool or not.
+        /// </summary>
+        public bool DoNotAnalyseOutput { get; set; }
```
devlead, line 109:
```suggestion
        public bool SkipOutputAnalysis { get; set; }
```

src/Cake.Common/Tools/InspectCode/InspectCodeSettings.cs
```diff
@@ -114,5 +114,11 @@ public sealed class InspectCodeSettings : ToolSettings
         /// Gets or sets the minimal severity of issues to report.
         /// </summary>
         public InspectCodeSeverity? Severity { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to skip analysis of the file
+        /// that was output by the command line tool or not.
+        /// </summary>
+        public bool DoNotAnalyseOutput { get; set; }
```
devlead, line 122:
```suggestion
        public bool SkipOutputAnalysis { get; set; }
```

janniksam: Naming things is hard... but that is way better. :)
I will rename it, when I am done working for today. ;)

devlead: Excellent 👍, please rebase against the latest develop too.

janniksam:
Should be good now.

devlead:
@janniksam your changes have been merged, thanks for your contribution 👍

Issue #2788, opened 2020-06-02
Tool:OpenCover - the register-setting should be an option, rather than a string


nils-a:
### What You Are Seeing?
The current way to use OpenCover is like this:
```
    OpenCover(cc => {
        cc.NUnit3(tests); // or whatever...
    },
        coverFile,
        new OpenCoverSettings{
            Register = "some-path-to-dll" 
        }
    );
```
`register` is represented as a string. Actually `register` could be empty (admin-mode) the word "user" (user-mode) or the path to a dll.

Additionally it seems impossible not to use the register-setting

### What is Expected?
 I propose something like this:

```
    OpenCover(cc => {
        cc.NUnit3(tests); // or whatever...
    },
        coverFile,
        new OpenCoverSettings()
            .WithoutRegister() // to omit the register-settings
            .WithRegisterAdmin() // -register
            .WithRegisterUser() // -register:user <-- this is the default
            .WithRegisterDll(someFilePath) // -register:path-to-dll
    );
```

Kamesh-devops:
I am getting below error while cake.tool 2.2.0
Error: Error downloading 'OpenCover.4.7.1221' from '[https://api.nuget.org/v3-flatcontainer/opencover/4.7.1221/opencover.4.7.1221.nupkg'.](https://api.nuget.org/v3-flatcontainer/opencover/4.7.1221/opencover.4.7.1221.nupkg%27.)

PR #2789, opened 2020-06-02
changed the register-property in OpenCover/OpenCoverSettings.cs from string

nils-a:
…to a custom OpenCoverRegisterOption to make it clearer what the actual options are. 

As suggested in #2788 this commit implements an `OpenCoverRegisterOption` with predifined static accessors:
* `OpenCoverRegisterOption.User` 
* `OpenCoverRegisterOption.Admin` 
* `OpenCoverRegisterOption.Dll(FilePath)` 

additionally I introduced extensions to `OpenCoverSettings`:

* `WithoutRegister()` this will set the register property to `null` an thereby omitting register alltogether
* `WithRegisterAdmin()` this will set the register property to `OpenCoverRegisterOption.Admin`
* `WithRegisterUser()` this will set the register property to `OpenCoverRegisterOption.User`
* `WithRegisterDll(FilePath)` this will set the register property to `OpenCoverRegisterOption.Dll(FilePath)`

To avoid breaking changes I have implemented an implicit cast from string to the new OpenCoverRegisterOption. So all cake file using the "old" syntax will continue to work without problems.

### DIFF
--- a/src/Cake.Common/Tools/OpenCover/OpenCoverRegisterOption.cs
+++ b/src/Cake.Common/Tools/OpenCover/OpenCoverRegisterOption.cs
@@ -1,0 +1,82 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Core.IO;
+
+using System;
+
+namespace Cake.Common.Tools.OpenCover
+{
+    /// <summary>
+    /// Represents the register-options:
+    ///  - empty for admin-registry-access
+    ///  - "user" for user-registry-access
+    ///  - path for non-registry-dll dll
+    /// </summary>
+    public class OpenCoverRegisterOption
+    {
+        private string value;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OpenCoverRegisterOption"/> class.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        public OpenCoverRegisterOption(string value)
+        {
+            this.value = value;
+        }
+
+        /// <summary>
+        /// Converts to string.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="System.String" /> that represents this instance.
+        /// </returns>
+        public override string ToString()
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return string.Empty;
+            }
+
+            return $":{value}";
+        }
+
+        /// <summary>
+        /// Gets the register-option representing the "user"-mode.
+        /// (this will translate to -register:user)
+        /// this
+        /// </summary>
+        public static OpenCoverRegisterOption User => new OpenCoverRegisterOption("user");
+
+        /// <summary>
+        /// Gets the register-option representing the "admin"-mode.
+        /// (this will translate to -register)
+        /// this
+        /// </summary>
+        public static OpenCoverRegisterOption Admin => new OpenCoverRegisterOption(string.Empty);
+
+        /// <summary>
+        /// Gets a register-option pointing to a dll.
+        /// (this will translate to -register:[path-to-dll])
+        /// </summary>
+        /// <param name="path">The path.</param>
+        /// <returns>The <see cref="OpenCoverRegisterOption"/></returns>
+        public static OpenCoverRegisterOption Dll(FilePath path)
+        {
+            return new OpenCoverRegisterOption(path.FullPath);
+        }
+
+        /// <summary>
+        /// Performs an implicit conversion from <see cref="System.String"/> to <see cref="OpenCoverRegisterOption"/>.
+        /// (Since the switch from pure string to <see cref="OpenCoverRegisterOption"/> is a breaking change)
+        /// </summary>
+        /// <param name="option">The option.</param>
+        /// <returns>
+        /// The result of the conversion.
+        /// </returns>
+        [Obsolete("use new OpenCoverRegisterOption() instead.")]
+        public static implicit operator OpenCoverRegisterOption(string option) => new OpenCoverRegisterOption(option);
+    }
+}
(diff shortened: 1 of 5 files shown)


src/Cake.Common/Tools/OpenCover/OpenCoverRegisterOption.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            return $":{value}";
+        }
+
+        /// <summary>
+        /// Gets the register-option representing the "user"-mode.
+        /// (this will translate to -register:user)
+        /// this
+        /// </summary>
+        public static OpenCoverRegisterOption User => new OpenCoverRegisterOption("user");
```
Roemer, line 50: This will create a new object each time which is unnecessary. I think you can use
`public static OpenCoverRegisterOption User { get; } = new OpenCoverRegisterOption("user");`

src/Cake.Common/Tools/OpenCover/OpenCoverRegisterOption.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        /// <summary>
+        /// Gets the register-option representing the "admin"-mode.
+        /// (this will translate to -register)
+        /// this
+        /// </summary>
+        public static OpenCoverRegisterOption Admin => new OpenCoverRegisterOption(string.Empty);
```
Roemer, line 57: Same as with `User`.

src/Cake.Common/Tools/OpenCover/OpenCoverRegisterOption.cs
```diff
@@ -0,0 +1,81 @@
...
+                return string.Empty;
+            }
+
+            return $":{value}";
+        }
+
+        /// <summary>
+        /// Gets the register-option representing the "user"-mode.
+        /// (this will translate to -register:user)
+        /// this
```
Roemer, line 48: Either this `this` is too much or something is missing.

src/Cake.Common/Tools/OpenCover/OpenCoverRegisterOption.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// </summary>
+        public static OpenCoverRegisterOption User => new OpenCoverRegisterOption("user");
+
+        /// <summary>
+        /// Gets the register-option representing the "admin"-mode.
+        /// (this will translate to -register)
+        /// this
```
Roemer, line 55: Either this `this` is too much or something is missing.

src/Cake.Common/Tools/OpenCover/OpenCoverRunner.cs
```diff
@@ -141,7 +141,10 @@ public sealed class OpenCoverRunner : Tool<OpenCoverSettings>
                 builder.Append("-mergeoutput");
             }
 
-            builder.AppendSwitch("-register", ":", settings.Register);
+            if (settings.Register != null)
+            {
+                builder.AppendSwitch("-register", string.Empty, settings.Register.ToString());
```
Roemer, line 146: I personally would prefer to have the quote in here as second argument instead of the `ToString` in `Register`

nils-a: @Roemer Having the colon here leads to `-register:` for `OpenCoverRegisterOption.Admin` which in turn is not valid. So the alternative would be something like 
```
var registerValue = settings.Register.toString();
if(string.isNullOrEmtpy(registerValue){
   builder.AppendSwitch("-register");
} else {
  builder.AppendSwitch("-register", ":", registerValue);
}
```

which imho is much less readble. What do you think?

(or is it possible to have `AppendSwitch` omit the separator if the last parameter is empty?)

Roemer: Ah I see. Maybe add this information as a comment above this line so nobody would change and break that in the future.

nils-a:
@devlead I dont't think it's a breaking change, is it? I added https://github.com/cake-build/cake/pull/2789/files#diff-6cf2812a0f7922beab63c57ec49cc496R89 to be fully backwards-compatible with existsing cake-scripts. 

(Old scripts using the `string`-syntax option will get a deprecation-note, though.)

devlead:
> 
> 
> @devlead I dont't think it's a breaking change, is it? I added https://github.com/cake-build/cake/pull/2789/files#diff-6cf2812a0f7922beab63c57ec49cc496R89 to be fully backwards-compatible with existsing cake-scripts.
> 
> (Old scripts using the `string`-syntax option will get a deprecation-note, though.)

So I haven't taken this for a spin yet or reviewed it in depth yet.

But in general, any kind of type change will break ABI compatibility and needs further investigation.
As this will target 1.0 or later it's probably fine, but any addin or software referencing Cake common might fail.

Also becuse 
```diff
-                fixture.Settings.Register = "Path32";
+                fixture.Settings.Register = new OpenCoverRegisterOption("Path32");
```
we don't have any unit tests nor integration tests ensuring it still works, would be nice to cover both new and old way in tests.

Is null a valid value? Otherwise, we should guard against that in the constructor.

I'm wondering if value & ToString should be cached, so things are validated early.

I'm also wondering if OpenCoverRegisterOption maybe should be an abstract class, and admin, user and FilePath should be separate implementations.

nils-a:
> But in general, any kind of type change will break ABI compatibility and needs further investigation.
> As this will target 1.0 or later it's probably fine, but any addin or software referencing Cake common might fail.

While I *think* nothing will break, you are right: Something could very well break.


> we don't have any unit tests nor integration tests ensuring it still works, would be nice to cover both new and old way in tests.

Thats why I added https://github.com/cake-build/cake/pull/2789/files#diff-5994199ce162e54f3717f057e65ecb5bR276 to test the implicit string conversion and to ensure compatibility.

> Is null a valid value? Otherwise, we should guard against that in the constructor.
> 
> I'm wondering if value & ToString should be cached, so things are validated early.
> 
> I'm also wondering if OpenCoverRegisterOption maybe should be an abstract class, and admin, user and FilePath should be separate implementations.

I will re-work this.

pascalberger:
Failing AppVeyor build will be fixed with https://github.com/cake-build/cake/pull/3574

nils-a:
Superseded by https://github.com/cake-build/cake/pull/3608

PR #2823, opened 2020-07-04
(GH-2822) Add support for .Net Global Tool

### LINKED ISSUE #2822: Add support of ReportGenerator global tool
At the minute, Cake support for ReportGenerator only includes the full .net framework version.  Similar to what has been done for GitVersion and GitReleaseManager, we should add support for running the global tool version as well.

gep13:
Fixes #2822


src/Cake.Common/Tools/ReportGenerator/ReportGeneratorRunner.cs
```diff
@@ -126,7 +126,7 @@ protected override string GetToolName()
         /// <returns>The tool executable name.</returns>
         protected override IEnumerable<string> GetToolExecutableNames()
         {
-            return new[] { "ReportGenerator.exe" };
+            return new[] { "ReportGenerator.exe", "reportgenerator" };
```
pascalberger, line 129: Is reportgenerator corret? Or should it be `dotnet-reportgenerator`?

PR #2826, opened 2020-07-08
GH2825: Add HandleExitCode to ToolSettings

flcdrg:
- When true, then ProcessExitCode method is not called

### DIFF
--- a/src/Cake.Core/Tooling/ToolSettings.cs
+++ b/src/Cake.Core/Tooling/ToolSettings.cs
@@ -73,5 +73,41 @@ namespace Cake.Core.Tooling
         /// </code>
         /// </example>
         public IDictionary<string, string> EnvironmentVariables { get; set; } = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Gets or sets whether the exit code from the tool process causes an exception to be thrown.
+        /// <para>
+        /// If the delegate is null (the default) or returns false, then an exception is thrown upon a non-zero exit code.
+        /// </para>
+        /// <para>
+        /// If the delegate returns true then no exception is thrown.
+        /// </para>
+        /// <para>
+        /// This can be useful when the exit code should be ignored, or if there is a desire to apply logic that is conditional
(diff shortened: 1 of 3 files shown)


src/Cake.Core/Tooling/ToolSettings.cs
```diff
@@ -73,5 +73,17 @@ public class ToolSettings
...
+        /// Gets or sets a value indicating whether the exit code from the tool process causes an exception to be thrown.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// DotNetCoreTest("MyProject.csproj", new DotNetCoreTestSettings {
+        ///     IgnoreExitCode = true
+        /// });
+        /// </code>
+        /// </example>
+        public bool IgnoreExitCode { get; set; }
```
devlead, line 87: I'm wondering if a more flexible way would be
```suggestion
        public Func<int, bool> HandleExitCode { get; set; }
```
Then ignoring it would be as simple as
```csharp
HandleExitCode = _=> true
```
but it would also allow more granual control too i.e.
```csharp
HandleExitCode = exitCode => exitCode switch {
                                1 => true, // threat 1 and 2 as handled "ok".
                                2 => true,
                                _ => false
                             }
```

Tool implementation could then be something like
```csharp
var exitCode = process.GetExitCode();
if (!settings.HandleExitCode?.Invoke(exitCode) ?? true)
{
    ProcessExitCode(exitCode);
}
```

flcdrg: I like it. I'll update it to do that.

devlead: @flcdrg excellent, thanks! Just ping when you've addressed feedback and rebased against latest develop.

flcdrg: @devlead rebased and ready for review :-)

devlead (approved):
LGTM 👍

devlead:
@flcdrg your changes have been merged, thanks for your contribution 👍

Issue #741, opened 2016-03-08
Add IsRunningOnMacOs() alias


devlead:
For XPLAT builds it would be excellent to be able to check if script is running on a Mac.

Don't know best way to check this, Mono isn't relible, seen this code that pinvokes to get Darwin from uname
https://github.com/jpobst/Pinta/blob/master/Pinta.Core/Managers/SystemManager.cs#L162

devlead:
This is covered in #502

PR #2842, opened 2020-08-13
Add IsRunningOn for macOS and Linux 

mattleibow:
This is very helpful for when things run on just macOS or just Linux. The Unix variant is nice for the basic Windows/not-Windows, but in some cases, the build script needs to know if Xcode should be used or some other mac-specific tool.

This actually was supposedly "added" in #741, but it never actually was?

Did I miss something? If this was excluded for a specific reason, then feel free to close.

I am also looking at the code, and not sure if this fits exactly since the Windows variant is basically !Unix, however, I can iterate on this PR based on any feedback.

### DIFF
--- a/src/Cake.Common.Tests/Unit/EnvironmentAliasesTests.cs
+++ b/src/Cake.Common.Tests/Unit/EnvironmentAliasesTests.cs
@@ -226,5 +226,65 @@ namespace Cake.Common.Tests.Unit
                 Assert.Equal(expected, result);
             }
         }
+
+        public sealed class TheIsRunningOnLinuxMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Context_Is_Null()
+            {
+                // Given, When
+                var result = Record.Exception(() => EnvironmentAliases.IsRunningOnLinux(null));
+
+                // Then
+                AssertEx.IsArgumentNullException(result, "context");
+            }
+
+            [Theory]
+            [InlineData(PlatformFamily.Linux, true)]
+            [InlineData(PlatformFamily.OSX, false)]
+            [InlineData(PlatformFamily.Windows, false)]
+            public void Should_Return_Correct_Value(PlatformFamily family, bool expected)
+            {
+                // Given
+                var context = Substitute.For<ICakeContext>();
+                context.Environment.Returns(new FakeEnvironment(family));
+
+                // When
+                var result = EnvironmentAliases.IsRunningOnLinux(context);
+
+                // Then
+                Assert.Equal(expected, result);
+            }
+        }
+
+        public sealed class TheIsRunningOnMacMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Context_Is_Null()
+            {
+                // Given, When
+                var result = Record.Exception(() => EnvironmentAliases.IsRunningOnMacOs(null));
+
+                // Then
+                AssertEx.IsArgumentNullException(result, "context");
+            }
+
+            [Theory]
+            [InlineData(PlatformFamily.Linux, false)]
+            [InlineData(PlatformFamily.OSX, true)]
+            [InlineData(PlatformFamily.Windows, false)]
+            public void Should_Return_Correct_Value(PlatformFamily family, bool expected)
+            {
+                // Given
+                var context = Substitute.For<ICakeContext>();
+                context.Environment.Returns(new FakeEnvironment(family));
+
+                // When
+                var result = EnvironmentAliases.IsRunningOnMacOs(context);
+
+                // Then
+                Assert.Equal(expected, result);
+            }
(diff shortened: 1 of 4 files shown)


src/Cake.Common/EnviromentAliases.cs
```diff
@@ -186,6 +186,58 @@ public static bool IsRunningOnUnix(this ICakeContext context)
...
+        /// }
+        /// </code>
+        /// </example>
+        /// <param name="context">The context.</param>
+        /// <returns>
+        ///   <c>true</c> if the build script running on a macOS based system; otherwise <c>false</c>.
+        /// </returns>
+        [CakeMethodAlias]
+        [CakeAliasCategory("Platform")]
+        public static bool IsRunningOnMac(this ICakeContext context)
```
mattleibow, line 206: I went for "Mac" here as this is what seems to be used more commonly (and maybe ages better since OS "X" is "10" and maybe macOS finally decides to go to 11 in 50 years)

Also, the basic search of people implementing their own version is 3K "Mac" and 94 "OSX" (with various casing)

patriksvensson: Perhaps `IsRunningOnMacOs` is a better name here? What do you think @cake-build/cake-team

mattleibow: I left it off because the casing was weird 😬 but there are lots of ways to write that...

ecampidoglio: I also think that `IsRunningOnMacOs` is more appropriate. The term "Mac" refers to the line of computers (which comprises products like MacBook, Mac Pro and Mac mini), while macOS is the operating system that runs on those computers.

@mattleibow I think you're right in avoiding any reference to 'X' or "10", since the next version of [macOS (Big Sur)](https://en.wikipedia.org/wiki/MacOS_Big_Sur) coming out this fall will mark the move to 11. 🎉

devlead: Agree with @patriksvensson & @ecampidoglio 👍

devlead, line 206:
```suggestion
        public static bool IsRunningOnMacOs(this ICakeContext context)
```

devlead (changes requested):
In general LGTM 👍 One minor suggestion.

src/Cake.Common/EnviromentAliases.cs
```diff
@@ -186,6 +186,58 @@ public static bool IsRunningOnUnix(this ICakeContext context)
             return context.Environment.Platform.IsUnix();
         }
 
+        /// <summary>
+        /// Determines whether the build script running on a macOS based system.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// if (IsRunningOnMac())
```
mattleibow, line 194:
```suggestion
        /// if (IsRunningOnMacOs())
```

src/Cake.Common.Tests/Unit/EnvironmentAliasesTests.cs
```diff
@@ -226,5 +226,65 @@ public void Should_Return_Correct_Value(PlatformFamily family, bool expected)
...
+            }
+        }
+
+        public sealed class TheIsRunningOnMacMethod
+        {
+            [Fact]
+            public void Should_Throw_If_Context_Is_Null()
+            {
+                // Given, When
+                var result = Record.Exception(() => EnvironmentAliases.IsRunningOnMac(null));
```
mattleibow, line 266:
```suggestion
                var result = Record.Exception(() => EnvironmentAliases.IsRunningOnMacOs(null));
```

src/Cake.Common.Tests/Unit/EnvironmentAliasesTests.cs
```diff
@@ -226,5 +226,65 @@ public void Should_Return_Correct_Value(PlatformFamily family, bool expected)
...
+            [InlineData(PlatformFamily.OSX, true)]
+            [InlineData(PlatformFamily.Windows, false)]
+            public void Should_Return_Correct_Value(PlatformFamily family, bool expected)
+            {
+                // Given
+                var context = Substitute.For<ICakeContext>();
+                context.Environment.Returns(new FakeEnvironment(family));
+
+                // When
+                var result = EnvironmentAliases.IsRunningOnMac(context);
```
mattleibow, line 283:
```suggestion
                var result = EnvironmentAliases.IsRunningOnMacOs(context);
```

mattleibow:
Updated!

😬

devlead (approved):
LGTM 👍

devlead:
@mattleibow your changes have been merged, thanks for your contribution 👍

PR #2858, opened 2020-09-18
GH2857: Add support for .NET 5

devlead:
Adds support for .NET 5


devlead:
@cake-build/cake-team this should now be ready for review.

.github/workflows/build.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     name: Build
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
ecampidoglio, line 14: **suggestion (history):** you might want to pull this out into its own commit.

Allowing all builds in a matrix to run [even if one of them failed](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run) is useful in of itself (and has way lower impact), so I think it would make sense to decouple it from the .NET 5 support work. That way, we could bring it in sooner, if need be.

src/Cake.Core.Tests/Fixtures/NotFormattableFixture.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
ecampidoglio, line 5: **nitpick:** it looks like these import statements are unused.

src/Cake.Core.Tests/Fixtures/NotFormattableFixture.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Cake.Core.Tests.Fixtures
+{
+    internal sealed class NotFormattableFixture
+    {
+        public static NotFormattableFixture Default { get; } = new NotFormattableFixture();
```
ecampidoglio, line 11: **suggestion (naming):** I tend to avoid using the word "_default_" when naming things, since it doesn't really set any expectations.

If the role of this property is to expose a singleton, I would suggest calling it something more specific like `Instance`.

ecampidoglio (approved):
I left a couple of comments for some minor things, but otherwise LGTM. 👍

src/Cake.Core.Tests/Fixtures/NotFormattableFixture.cs
devlead: this should now be addressed

src/Cake.Core.Tests/Fixtures/NotFormattableFixture.cs
devlead: this should now be resolved

.github/workflows/build.yml
devlead: Makes sense, but no other 1.0 PRs pending atm so if people happy with this PR it can go in in this PR sooner rather than later. Otherwise, I can do a separate PR for this.

patriksvensson (approved):

patriksvensson:
@devlead Merged! Thank you for your contribution. Much appreciated! 👍

PR #2871, opened 2020-10-03
GH2870: Add helpers for adding multiple strings to ProcessArgumentBuilder

### LINKED ISSUE #2870: Add helpers for adding multiple strings to ProcessArgumentBuilder
[`ProcessArgumentBuilder`](https://github.com/cake-build/cake/blob/7ca5a28479cc7b7deb9a8e02ca495397c344f8e6/src/Cake.Core/IO/ProcessArgumentBuilder.cs#L16) has a [FromString](https://github.com/cake-build/cake/blob/7ca5a28479cc7b7deb9a8e02ca495397c344f8e6/src/Cake.Core/IO/ProcessArgumentBuilder.cs#L119) helper and implicit conversion [operator](https://github.com/cake-build/cake/blob/7ca5a28479cc7b7deb9a8e02ca495397c344f8e6/src/Cake.Core/IO/ProcessArgumentBuilder.cs#L109) that makes it trivial to work with a _single_ string but has no helpers for another common case of adding _multiple_ strings, assuming they should be handled homogeneously - either all naked or all quoted. Also, the single string case could use a helper that quotes its argument too, so overall I propose the following additional helpers:
* `FromStringQuoted`
* `FromStrings`
* `FromStringsQuoted`
* `ToProcessArguments` (extension)

gitfool:
Fixes #2870.

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/DotNetCore/Execute/DotNetCoreExecutorFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/DotNetCore/Execute/DotNetCoreExecutorFixture.cs
@@ -11,7 +11,7 @@ namespace Cake.Common.Tests.Fixtures.Tools.DotNetCore.Execute
     {
         public FilePath AssemblyPath { get; set; }
 
-        public string Arguments { get; set; }
+        public ProcessArgumentBuilder Arguments { get; set; }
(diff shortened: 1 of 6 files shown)


devlead (approved):
LGTM 👍

devlead:

PR #2876, opened 2020-10-07
(GH-2836) Link to website for example in README

### LINKED ISSUE #2836: Update README with more up-to-date "getting started" information
We should consider updating Cake's `README.md` with more up-to-date "getting started" information. I think the recommendation is to use the dotnet tool instead of bootstrappers.

pascalberger:
Remove example from README and link to website instead.

Fixes #2836


gep13 (approved):
LGTM!

PR #2881, opened 2020-10-11
GH2734: Tries to resolve assembly if not .resources file is found.

### LINKED ISSUE #2734: Can't resolve resource assemblies
Workaround in script:

```csharp
AppDomain.CurrentDomain.AssemblyResolve += (_, args) =>
{
    if (args.Name.Contains(".resources, "))
    {
        return System.Reflection.Assembly.Load(args.Name.Replace(".resources, ", ", "));
    }

    return null;
};
```

But we should add this to the global assembly resolve event.

### LINKED ISSUE #2743: Tool resolution for multiple names should be breadth first
Following up from https://github.com/cake-contrib/Cake.DotNetTool.Module/issues/20, I discovered the unexpected and inconsistent behavior that a tool with one name in the `PATH` was preferred over the tool with another name in the tool registry or tools directory.

Specifically, in this case, `GitVersion.exe` is already installed into the `PATH` by Chocolatey for the [Windows](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md) virtual environment used by GitHub Actions. During the Cake build, a different version is specified to be installed into the Cake tools subdirectory; it should have been used instead but was not, since the current tool resolution strategy does not take into account multiple tool executable names for each tool resolution source, and so is effectively depth first.

mholo65:
Fixes #2734  

Will need to find out a simple way to create a integration test for this. Maybe an addin with a resx file compiled to a satellite `.resource` file that's missing when addin is loaded.

### DIFF
--- a/src/Cake/Infrastructure/Scripting/ScriptAssemblyResolver.cs
+++ b/src/Cake/Infrastructure/Scripting/ScriptAssemblyResolver.cs
@@ -34,27 +34,40 @@ namespace Cake.Infrastructure.Scripting
             if (_resolvedNames.Add(name.Name))
             {
                 _log.Verbose($"Resolving assembly {args.Name}");
-                try
-                {
-                    var assembly = AppDomain.CurrentDomain.GetAssemblies()
-                        .FirstOrDefault(x => !x.IsDynamic && x.GetName().Name == name.Name)
-                        ?? Assembly.Load(name.Name);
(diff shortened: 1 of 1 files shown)


devlead (approved):
LGTM 👍

devlead:
@mholo65 your changes have been merged, thanks for your contribution 👍

PR #2910, opened 2020-10-20
GH-2897: Add tests for MyGetProvider

### LINKED ISSUE #2897: Add tests for MyGetProvider
Add tests for MyGetProvider

duracellko:
Fixes #2897 Add unit tests for MyGetProvider.

Additionally replacement of special characters was rewritten. Although original implementation (surprisingly) worked, it relied on implementation of Dictionary. Problem was that `|` should be replaced first, otherwise it would be replaced multiple times. However, `Dictionary` does not guarantee to enumerate items in the same order as they were added to the dictionary. So it was not guaranteed that `|` would be first token to replace.

I changed the replacement to iterate through characters of source string.

Also original code replaced `[` by `|['`. I assume it was type and correct replacement is `|[`

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Build/MyGetFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Build/MyGetFixture.cs
@@ -1,0 +1,34 @@
+﻿using Cake.Common.Build.MyGet;
+using Cake.Common.Tests.Fakes;
+using Cake.Core;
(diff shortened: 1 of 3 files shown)


devlead (approved):
LGTM 👍

devlead:
@duracellko your changes have been merged, thanks for your contribution 👍

PR #2924, opened 2020-10-31
GH2595: Installs prerelease package versions when explicitly specified

### LINKED ISSUE #2595: Misleading output message when trying to install prerelease package with the in-process nuget installer
### What You Are Seeing?
When loading a script package via the preprocessor directive, the loading fails because the behavior of the in process nuget installer changed. Before `0.34.1` it allowed to install nuget packages with prelease versions without explicitly adding the `&prerelease` statement.

The current implementation in `0.34.1` will result in a misleading output message rather stating the fact that it was not possible to find a release version with the given version number.

```
// works!
#load "nuget:?package=My.Cool.Script.Package&version=12.2.1.171"

// does not work, console just throws No scripts found in NuGet package My.Cool.Script.Package
#load "nuget:?package=My.Cool.Script.Package&version=12.3.1-rc0001.174"
```

### What is Expected?
```
// works!
#load "nuget:?package=My.Cool.Script.Package&version=12.2.1.171"

// works!
// or
// Unable to find release package in specified version.
#load "nuget:?package=My.Cool.Script.Package&version=12.3.1-rc0001.174"
```

`0.34.1`

`64`

`Windows`

agc93:
Got hit by this one today and based on the conversation in #2595, seems like the consensus is that we should honour explicitly specified prerelease versions, even without a `prerelease` flag. From what I can test, behaviour is otherwise unchanged.

---

fixes #2595

### DIFF
--- a/src/Cake.NuGet/Install/NuGetPackageInstaller.cs
+++ b/src/Cake.NuGet/Install/NuGetPackageInstaller.cs
@@ -133,8 +133,8 @@ namespace Cake.NuGet.Install
 
             if (packageIdentity.Version.IsPrerelease && !package.IsPrerelease())
             {
-                // TODO: Is this allowed? If not, log and return
-                return Array.Empty<IFile>();
+                // If a prerelease version is explicitly specified, we should install that with or without prerelease flag.
+                _log.Debug("Prerelease version string explicitly specified. Installing prerelease package version.");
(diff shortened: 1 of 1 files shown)


devlead (approved):
LGTM 👍

devlead:

PR #2925, opened 2020-11-02
Fix sentences which end with double full stop

pascalberger:
Fixes sentences in XML comments which end with double full stops.


gep13 (approved):
LGTM

gep13:
@pascalberger thanks for fixing these up!

PR #2946, opened 2020-11-15
GH2886/2904: Bring Frosting up to speed

patriksvensson:
Contains some unavoidable breaking changes.


devlead (changes requested):
Frosting Integration test needs to be updated to cater for the breaking changes.

devlead:
> ### Todo:
> 
>     * [ ]  Version number in template

Think this is already fixed by param `TemplateVersion`?
If running in CI that is passed
https://github.com/cake-build/cake/blob/11dc2434aef8ac4829c2e401074ef2495dddf2d3/build/parameters.cake#L120-L123

and build.csproj is poked
https://github.com/cake-build/cake/blob/11dc2434aef8ac4829c2e401074ef2495dddf2d3/src/Cake.Frosting.Template/Cake.Frosting.Template.csproj#L26-L28

devlead (approved):
After adjusted the following in integration tests
```diff
using Cake.Core;
using Cake.Frosting;
+using Microsoft.Extensions.DependencyInjection;
 
 public class Program : IFrostingStartup
 {
     public static int Main(string[] args)
-    {
-        // Create the host.
-        var host = new CakeHostBuilder()
-            .WithArguments(args)
+        => new CakeHost()
             .UseStartup<Program>()
-            .Build();
-
-        // Run the host.
-        return host.Run();
-    }
+            .Run(args);
 
-    public void Configure(ICakeServices services)
+    public void Configure(IServiceCollection services)
     {
         services.UseContext<Context>();
         services.UseLifetime<Lifetime>();
         services.UseWorkingDirectory("..");
    }
}
```
CI seems to be happier.

@patriksvensson as I commented earlier by the looks of it template versioning seems to work already?
Was there anything else you wanted to add before taking out of WIP?

Other than that this LGTM 👍

src/Cake.Frosting.Template/Cake.Frosting.Template.csproj
```diff
@@ -1,28 +1,30 @@
...
     <Content Include="templates\**\*" Exclude="templates\**\bin\**;templates\**\obj\**" />
-    <Content Include="../../build.config" Link="templates/cakefrosting/build.config" PackagePath="/content/templates/cakefrosting/build.config" />
-    <Content Include="build.ps1" Link="templates/cakefrosting/build.ps1"  PackagePath="/content/templates/cakefrosting/build.ps1" />
-    <Content Include="build.sh" Link="templates/cakefrosting/build.sh" PackagePath="/content/templates/cakefrosting/build.sh" />
     <Compile Remove="**\*" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="build.ps1" />
+    <None Remove="build.sh" />
```
pascalberger, line 25: Why are these two lines required?

pascalberger (changes requested):
One comment for a smaller issue. Beside that we should create issues for the breaking changes introduced with this PR so that they end up in release notes.

src/Cake.Frosting.Template/templates/cakefrosting/build.ps1
```diff
@@ -0,0 +1,2 @@
+dotnet run --project build/Build.csproj -- $args
```
pascalberger, line 1: We no longer want to download .NET as part of the bootstrapper? What's the reason for this?
Otherwise https://github.com/cake-build/cake/pull/2960 would be a solution using global.json.

devlead: Should the standard bootstrappers install .NET?
Maybe a good thing to keep these as simple as possible.
The template itself already requires .NET Core / .NET 5 to be installed.
I kinda like the simplicity of what's suggested here.
Maybe better to document `dotnet new globaljson` and have more advanced bootstrappers in our resources repo?

pascalberger: I was never team "Install .NET in bootstrapper" 😉 It's just what most people seem to prefer nowadays, and I wanted to bring it up to discussion, since it's quite a hidden change. Template requiring .NET to be installed probably isn't an argument since people / machine running build in most cases is not the same which did run the template.

@gep13 Did you already though about how bootstrappers in resources repo should look like for .NET tool and .NET Core regarding installation of .NET?

gep13: @pascalberger I haven't come to a final decision about how these should look yet.  I am still thinking that installing .Net locally into the repository folder, based on what is in the global.json is a good idea though.  I liked what was done with the `jsonfile` parameter, and I thought about putting an if statement round that to either pass it in or not, based on the global.json file being present.

src/Cake.Frosting.Template/Cake.Frosting.Template.csproj
patriksvensson: 🤷‍♂️

src/Cake.Frosting.Template/templates/cakefrosting/build.ps1
patriksvensson: I don't think that the frosting bootstrapper should install .NET. It should just be a easy way to run a build which exists in another directory.

devlead: Then it maybe should push / change / pop directory so it's executed in script directory?

patriksvensson: That's not needed since the working directory already is the script directory. Working directory should imo not be set by a script but be explicitly stated by the person calling the script

src/Cake.Cli/Cake.Cli.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyName>Cake.Cli</AssemblyName>
+    <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
```
pascalberger, line 4: Since this is also used in Cake.Core it should also target `net46`

pascalberger (changes requested):
The newly introduced Cake.Cli library needs to be added to https://github.com/cake-build/cake/blob/a2f339b682a48a67496756ac27207cbaea56b382/build/parameters.cake#L107 to be published.

src/Cake.Cli/Cake.Cli.csproj
devlead: It's not used in `Cake.Core`, it references `Cake.Core` though.

For `net46` won't work because of dependencies, it could and probably should target `net461` too as `Cake.exe` does.

I can add that.

devlead:
> 
> 
> The newly introduced Cake.Cli library needs to be added to
> 
> https://github.com/cake-build/cake/blob/a2f339b682a48a67496756ac27207cbaea56b382/build/parameters.cake#L107
> to be published.

Good catch I'll add that.

devlead:
@pascalberger Added Cake.Cli NuGet & net461

pascalberger (approved):
I didn't go through every single change, but overall it LGTM. 

I still want to have proper documentation out for breaking changes in release notes and on [website](https://cakebuild.net/docs/getting-started/upgrade) before we push this out.

devlead:

PR #2960, opened 2020-11-27
(GH-2956) Remove build.config from Frosting template

### LINKED ISSUE #2956: Wrong Cake version in build.config
### What You Are Seeing?
When creating a new Frosting project from a template a `build.config` file is included in the bootstrapping scripts, based on the [build.config file used to build Cake](https://github.com/cake-build/cake/blob/develop/build.config).

### What is Expected?
Cake version in the generated `build.config` should match the version of the template (e.g. for Cake.Frosting template 1.0 it should use Cake version 1.0)

1.0 RC1

64-Bit

Windows

No

- Install latest Cake Frosting Template
- Create a new project

pascalberger:
Removes `build.config` from Frosting template. Cake version from the `build.config` file was not used and .NET SDK version is now used from `global.json` file which has been added to the template. One side effect of this approach is that we always download the .NET install script on Windows, but this is something which we already did on Linux/macOS.

Also cleaned up some stuff which was in the bootstrapper script, but not used.

Fixes #2956

### DIFF
--- a/src/Cake.Frosting.Template/Cake.Frosting.Template.csproj
+++ b/src/Cake.Frosting.Template/Cake.Frosting.Template.csproj
@@ -17,9 +17,9 @@
 
   <ItemGroup>
     <Content Include="templates\**\*" Exclude="templates\**\bin\**;templates\**\obj\**" />
-    <Content Include="../../build.config" Link="templates/cakefrosting/build.config" PackagePath="/content/templates/cakefrosting/build.config" />
     <Content Include="build.ps1" Link="templates/cakefrosting/build.ps1"  PackagePath="/content/templates/cakefrosting/build.ps1" />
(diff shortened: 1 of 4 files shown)


pascalberger:
When running the build.sh on WSL/Ubuntu I get the folllowing error:

image

But work for @gep13 on macOS.

gep13:
@cake-build/cake-team Is anyone else in a position to take this change for a spin?  As @pascalberger mentioned, this is working for me on my Mac.

pascalberger:
> Isn't this taken care of in #2946?

I can't say since there's not much of information what exactly changed in the PR and a lot of files have changed. On a quick glance the file seems to be removed, which might solve it  (but I don't see how it is done after #2946)

pascalberger:
Had another look at #2946 and agree that the way bootstrapping is done in #2946 will make this obsolete

PR #2969, opened 2020-12-05
GH-2966: Enable AnsiConsoleRenderer when running on TeamCity 9.1 and later

augustoproiete:
Update the `AnsiDetector` to enable the `AnsiConsoleRenderer` when Cake is running on modern versions of TeamCity (i.e. 9.1+)

The colors don't looks amazing with the default theme given Cake is setting the dark background color. We may want to change as to not set the background color or perhaps introduce separate color palettes for the different CI systems.

image

---

Partially closes ... #2966

PR #2968, opened 2020-12-05
GH-2966: Enable AnsiConsoleRenderer when running on Azure Pipelines

augustoproiete:
Update the `AnsiDetector` to enable the `AnsiConsoleRenderer` when Cake is running on Azure Pipelines

image

---

Partially closes ... #2966 (one more PR to follow)

Issue #2966, opened 2020-12-04
Enable AnsiConsoleRenderer in TeamCity and Azure Pipelines


KatoStoelen:
I see that AnsiConsoleRenderer is enabled for GitHub Actions (and terminals that supports it).

Since both TeamCity (since 9.1) and Azure Pipelines now supports ANSI coloring, it would make the world much more colorful if AnsiConsoleRenderer is enabled for those build systems as well.

augustoproiete:
@KatoStoelen I've opened PRs to enable the `AnsiConsoleRenderer` for Azure Pipelines (#2968) and TeamCity (#2969).

In the meantime (as a workaround) if you want to enable colors _today_, you can "cheat" by setting an environment variable that will tell Cake that ANSI is supported:

- If your builds **are** running on Windows, set an environment variable `ConEmuANSI` with the value `On`
- If your builds are **not** running on Windows, set an environment variable `TERM` with the value `ansi`

This works for any CI system, including TeamCity and Azure Pipelines.

---

Just be aware that the colors on TeamCity don't look as nice as you'd expect because Cake is sets a black background color on the messages, which are displayed on a white background in TeamCity if you're using the default theme.

![image](https://user-images.githubusercontent.com/177608/101235993-9d350980-36a3-11eb-9d9f-ba0901e86474.png)

---

Also, some messages such as the summary report at the end of the build (with green forecolor on a regular console) are not emitting ANSI color codes and instead are using `Console.ForegroundColor` thus they won't have colors even with the `AnsiConsoleRenderer` enabled.

KatoStoelen:
Thanks, @augustoproiete!

Went looking for ways to customize `ConsolePalette`, but did not find any (apart from using Cake.Frosting and a custom `IConsole`). Hence, I started to think that it might be better to make this an opt-in feature. Either that, or make ConsolePalette customizable, or add some `BuildSystem` specific palettes (might cause issues for people using custom themes).

I guess with your current PRs, there is no way to opt-out of ANSI coloring?

augustoproiete:
@KatoStoelen Yes, as you probably noticed the [colors are all hard-coded in AnsiConsoleRenderer.](https://github.com/cake-build/cake/blob/4214b5201ccb35f76b49ae77eb7180875006282f/src/Cake.Core/Diagnostics/Console/AnsiConsoleRenderer.cs#L25-L63) at the moment, so supporting custom color palettes would be a new feature to be added.

[...]

PR #2970, opened 2020-12-05
GH-2966: Update AnsiDetector to support the TERM env. variable on Windows

augustoproiete:
When I was working on #2968 and #2969 I noticed that the `TERM` environment variable that would allow forcing Cake to enable ANSI is not checked when running on Windows.

This PR is just changing the order of the conditions so that the check for the `TERM` variable is done before checking if running on Windows.

---

Relates to #2966

### DIFF
--- a/src/Cake.Core/Diagnostics/Console/AnsiDetector.cs
+++ b/src/Cake.Core/Diagnostics/Console/AnsiDetector.cs
@@ -55,6 +55,16 @@ namespace Cake.Core.Diagnostics
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


augustoproiete:
This PR relates to #2968 and #2969 and touches the same area of the code, so I expect merge conflicts to arise after any of them are merged. It's a simple rebase which I can do as PRs are merged.

I decided to do separate PRs to make sure they don't block each other as they are separate improvements that can be independently merged.

devlead (approved):
LGTM 👍

To save CI time I'll cherry-pick into #2968

Issue #2976, opened 2020-12-08
Support custom ConsolePalette/ANSI colors


KatoStoelen:
Regarding #2966.

In build systems that have a build log with white background, the current palette is not... the prettiest of palettes. Would be great to have a way of customizing the colors. Could be by choosing between different "official" palettes, or by explicitly setting the different colors of the palette.

I suspect that you could achieve this using Cake.Frosting and a custom `IConsole` where you handle ANSI coloring yourself, but it would be nice utilize the built in ANSI coloring and do this without Cake.Frosting.

pascalberger:
Related issue https://github.com/cake-build/cake/issues/2852

augustoproiete:
> Could be by choosing between different "official" palettes, or by explicitly setting the different colors of the palette

Agreed. It would be nice if there were official palettes to choose from. Maybe start with two, one similar to the current one that looks good on dark backgrounds and an alternative one that looks good on light backgrounds.

Perhaps (as a separate issue) also add an extension point that can be used to set a different color palette from a module?

secretMeedo:
I fixed a similar issue for my company. Please let me know if this issue is still open for contribution so I can investigate it further.

abdelghani29:
Hello, can I try working on this issue?

PR #2984, opened 2020-12-12
GH-2852: Use the default background color when writing to the Console

### LINKED ISSUE #2852: Terminal output colours
I'm using [Windows Terminal](https://github.com/microsoft/terminal) with [PowerShell 7](https://github.com/PowerShell/PowerShell), which normally has a black background, but I've configured it to use the now classic blue background:

image

The text with black background looks a little jarring. Are the terminal output colours looking as expected / intended?

augustoproiete:
Quick fix to remove the black background color when writing messages to the console, and use the default background color instead.

It seems to work well for any dark background. For white backgrounds (e.g. TeamCity) it will require a different color palette as some messages are written using white foreground color.

/cc @patriksvensson as per chat via e-mail

---

- Closes #2852
- Relates to #2976 and #2966

### DIFF
--- a/src/Cake.Core/Diagnostics/Console/ConsoleColorEx.cs
+++ b/src/Cake.Core/Diagnostics/Console/ConsoleColorEx.cs
@@ -1,9 +1,0 @@
-﻿using System;
-
-namespace Cake.Core.Diagnostics.Console
-{
-    internal static class ConsoleColorEx
-    {
-        public const ConsoleColor Default = (ConsoleColor)(-1);
-    }
-}
--- a/src/Cake.Core/Diagnostics/Console/ConsoleColorEx.cs
+++ b/src/Cake.Core/Diagnostics/Console/ConsoleColorEx.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace Cake.Core.Diagnostics.Console
+{
+    internal static class ConsoleColorEx
+    {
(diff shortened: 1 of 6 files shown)


src/Cake.Core/Diagnostics/Console/ConsoleColorEx.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Cake.Core.Diagnostics.Console
+{
+    internal static class ConsoleColorEx
+    {
+        public const ConsoleColor Default = (ConsoleColor)(-1);
```
augustoproiete, line 7: I considered a couple of other options including changing references to `ConsoleColor` to `Nullable<ConsoleColor>` (and have `null` to indicate default) as well as creating a separate `CakeConsoleColor` enum with all elements of `System.ConsoleColor` + a default option, but both approaches would introduce a binary breaking change to the public API, including `IConsole` and friends, which I don't think is desirable for the v1 timeframe...

mholo65: I'd add this to Constants https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Constants.cs

augustoproiete: Good idea. I moved it to `Cake.Core/Constants.cs`

Issue #2975, opened 2020-12-07
Add support for opting out of ANSI coloring via NO_COLOR env. variable


augustoproiete:
Cake is able to [detect if ANSI coloring is supported](https://github.com/cake-build/cake/blob/4214b5201ccb35f76b49ae77eb7180875006282f/src/Cake.Core/Diagnostics/Console/AnsiDetector.cs#L43) in a number of cases, but once it detects that color _is_ supported, there's no way of opting out of it at the moment.

Some ideas:
- Check for the existence of an environment variable ([`NO_COLOR`](https://no-color.org) seems to be a standard) to opt out of ANSI coloring
- Add a new command-line switch e.g. `--no-color` and/or `--color=false` to opt out of ANSI coloring

---

Common options used to determine if ANSI coloring should be disabled used by some apps on GitHub:

- Command-line arguments:
    - `--no-color`
    - `--no-colors`
    - `--color=false`
    - `--color=never`

- Environment variables:
    - `TERM=dumb`
    - `FORCE_COLOR=false`
    - `FORCE_COLOR=0`

pascalberger:
Configuration should IMHO be possible as [other configuration options](https://cakebuild.net/docs/running-builds/configuration/set-configuration-values) through environment variable, configuration file and argument.

@patriksvensson Is this something which should be done in Spectre.CLI?

patriksvensson:
@pascalberger Makes sense to respect `NO_COLOR` since that's a pretty common convention. Other than that, the rest could be done in Cake.

augustoproiete:
Sounds good.

For the configuration options, I suggest adding a new configuration value under `Settings` called `Color`, which can be one of the following values:

- **`auto`** - Let AnsiDetector decide if color is supported or not based on environment (**_default_**)
- **`never`** - **opt out** of coloring, regardless of environment (_skip_ the logic to detect if color is supported or not)
- **`always`** - **opt in** on coloring, regardless of environment (_skip_ the logic to detect if color is supported or not)

The `color` option would be available as ...
- An argument passed into Cake: `--color <value>`
- A configuration file value: `Color=value` (under the `Settings` section)
- An environment variable: `CAKE_SETTINGS_COLOR=value`

... and follow the [same priority order as other configuration values](https://cakebuild.net/docs/running-builds/configuration/set-configuration-values).

---

[...]

PR #2999, opened 2020-12-13
GH-2975: Add Settings_Color configuration option

augustoproiete:
WIP. Opening a draft PR for feedback at this point to discuss next steps as this seems to require change on the timing when the Cake configuration is read.

Adding this new configuration option requires access to `ICakeConfiguration` at a much earlier stage in the Cake run lifecycle, more specifically at the time of resolution of `DefaultCommand` given that it needs an `ICakeLog` which in turn requires an `IConsoleRenderer` - and at this point the decision about ANSI vs non-ANSI console is made.

Does it make sense to register an `ICakeConfiguration` singleton [at the very start](https://github.com/cake-build/cake/blob/58fb66206ca843eb46ecac5799d931a262d1d32c/src/Cake/Program.cs#L83-L90), instead of reading the config later during [bootstrapping](https://github.com/cake-build/cake/blob/32f44da7d8c2996cbc00e43bc665f7921b91d23d/src/Cake/Features/Bootstrapping/BootstrapFeature.cs#L40-L42) or [building](https://github.com/cake-build/cake/blob/32f44da7d8c2996cbc00e43bc665f7921b91d23d/src/Cake/Features/Building/BuildFeature.cs#L59-L61), or is there a better way?


augustoproiete:
Meeting of 2022-06-12 - We decided that it's OK to write to the Console in case of errors reading the configuration

FreshKhalifa:
Great your airdrop tokens bonus is ready
You’re to send the gas fee of
$500 to the bot generated address from the wallet you’re to receive your
bonuse.

Issue #3018, opened 2021-01-02
Cake displays raw ANSI output after running specific executables


nsmithdev:
### What You Are Seeing?
ASCI character are returned to console for any output after running specific executables.
It works fine in the vscode console but exabits the problem in windows cmd console even when using PowerShell.
The same code used to work fine when running the older powershell version of cake but started happening when I switch to the dotnet tool version.
So far I can consistently reproduce with SqlCmd.EXE but it does not happen with something like ping which seems to works fine.
I'm passing an invalid argument for a simple poc that does not require a sql connection.
When the exe is successful for example change `-???` to just `-?` the cake output after the exe is run is still incorrect.

```
Sqlcmd: '??': Unexpected argument. Enter '-?' for help.
←[41m←[37;1mAn error occurred when executing task '←[0m←[41;1m←[37;1mDefault22←[0m←[41m←[37;1m'.←[0m
←[41m←[37;1mError: ←[0m←[41;1m←[37;1mOne or more errors occurred. (Restore database failed)←[0m
←[41m←[37;1m    ←[0m←[41;1m←[37;1mRestore database failed←[0m
```

### What is Expected?
To not see raw ASCI characters 

```
Sqlcmd: '??': Unexpected argument. Enter '-?' for help.
An error occurred when executing task 'Default22'.
Error: One or more errors occurred. (Restore database failed)
    Restore database failed
```

### What version of Cake are you using?
1.0.0-rc0002

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Window
 
### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)

```
Task("Default22").Does(() => {
    var settings = new ProcessSettings{ 
      Arguments = @"-???"
    };
    var exitCode = -1;
    using(var process = StartAndReturnProcess(@"SqlCmd.EXE", settings))
    {
        process.WaitForExit();
        exitCode = process.GetExitCode();
    }
    if(exitCode == 1){
        throw new Exception("Restore database failed");
    }
    Information("BlaBla");
    Information("Exit code: {0}", 5);
    Information("Exit code: {0}", exitCode);
    Information("BlaBla");
});
RunTarget("Default22");
```

```
where SqlCmd.EXE
C:\Program Files\Microsoft SQL Server\Client SDK\ODBC\170\Tools\Binn\SQLCMD.EXE

dotnet cake .\build.cake
```

[...]

PR #3021, opened 2021-01-04
GH-3018: Restore original Console Mode (if any) after running external processes

augustoproiete:
As described in https://github.com/cake-build/cake/issues/3018, external processes spawned by Cake have the ability to modify the current console mode, affecting the console output written by Cake itself, as is the case of SQLCMD disabling ANSI escape codes support.

This PR updates the `ProcessRunner` to take a snapshot of the current console mode (if any i.e. When running on Windows and with a console handle available) and restores it back to its original value in case the process that was executed changed the console mode.

---

Closes https://github.com/cake-build/cake/issues/3018

### DIFF
--- a/src/Cake.Core/IO/ConsoleMode.cs
+++ b/src/Cake.Core/IO/ConsoleMode.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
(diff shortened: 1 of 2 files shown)


devlead (approved):
LGTM 👍

devlead:

PR #3028, opened 2021-01-13
GH3020: Migrate from Spectre.Cli->Console

### LINKED ISSUE #3020: Migrate to Spectre.Console
Spectre.Cli has been moved into Spectre.Console library. We should also [migrate](https://spectresystems.github.io/spectre.console/cli/migration).

With Spectre.Console additional console features are made available, some of which can replace existing addins ([Figlet](https://spectresystems.github.io/spectre.console/widgets/figlet). We should make use of the additional console features provided by Spectre.Console by providing new or enhance existing aliases.

devlead:
* fixes #3020

### DIFF
--- a/src/Cake.Cli/Cake.Cli.csproj
+++ b/src/Cake.Cli/Cake.Cli.csproj
@@ -20,6 +20,6 @@
 
     <ItemGroup>
       <PackageReference Include="Autofac" Version="6.1.0" />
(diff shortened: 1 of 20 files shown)


augustoproiete:
Code LGTM.

@patriksvensson confirmed it's exactly the same code in Spectre.Console as Spectre.Cli and the only difference is that it's one assembly instead of two.

gep13 (approved):
LGTM

Issue #3038, opened 2021-01-18
Tool resolving in Frosting tasks


pascalberger:
### What You Are Seeing?
I can restore a .NET Tool using the Cake.DotNet.Module and then call it in the build context constructor. But when calling the same tool in the Setup or a task an error is thrown:

```
Error: GitVersion: Could not locate executable.
```

### What is Expected?
It should be possible to call .NET Tools also from setup or tasks.

### What version of Cake are you using?
1.0.0-rc0002

### Are you running on a 32 or 64 bit system?
64-Bit

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
No

### How Did You Get This To Happen? (Steps to Reproduce)

1. Create a new Froasting project and setup the host like this:

   ```
   return new CakeHost()
       .UseContext<BuildContext>()
       .UseModule<DotNetToolModule>()
       .UseTool(new Uri("dotnet:?package=GitVersion.Tool&version=5.6.3"))
       .Run(args);
   ```
2. Call `context.GitVersion();` in the constructor of the `BuildContext` class
3. Call `context.GitVersion();` in a task

pascalberger:
This is caused by working directory and probably a duplicate of https://github.com/cake-build/cake/issues/2904

patriksvensson:
@pascalberger Can't reproduce. I'm closing the issue.

![image](https://user-images.githubusercontent.com/357872/105614185-13aed780-5dc8-11eb-80c0-7867f4176ea2.png)

pascalberger:
I'll add a repro repository later

pascalberger:
Here's a repository where it can be reproduced: https://github.com/pascalberger/cake-gh-3038

For the reproduction to work it is important that GitVersion is not available on the system (e.g. installed as .NET global tool or through Chocolatey). In this case the behavior can still be seen when running with diagnostic verbosity where it is visible that global instance is called when running from a task instead of the version installed in the build script (see this GitHub actions run: https://github.com/pascalberger/cake-gh-3038/runs/1758010781?check_suite_focus=true#step:5:36).

augustoproiete:
> Here's a repository where it can be reproduced: https://github.com/pascalberger/cake-gh-3038

I can repro on my machine:

```
.\build.ps1 --verbosity=diagnostic
Tool GitVersion.Tool is already installed, with required version.
Building version 0.1.0...
Registering task: Default

========================================
Default
========================================
Executing task: Default
[...]

Issue #2904, opened 2017-02-22
(Frosting) Tool installer should respect configuration


patriksvensson:
Frosting should support [configuration](https://cakebuild.net/docs/running-builds/configuration/set-configuration-values).

dieterv:
@gep13 Frosting uses an internal `ToolInstaller` which does not respect the [tools path configuration option](https://cakebuild.net/docs/running-builds/configuration/default-configuration-values#tools-path).
Instead it is currently hardcoded to place tools into `_environment.WorkingDirectory.Combine("tools").MakeAbsolute(_environment)` which usually ends up being somthing like `project-root/tools/`.

[Something like this](https://github.com/dieterv/cake/commit/53b244fbad5d7c7112b91710103a683e72b27cb0) might be a first step to solve this. I did notice that `Cake.Core.Constants` is not accessible though and have no idea what would be preferred, making the constants public or duplicating them in the Cake.Frosting project somewhere?

Note that this would also require the `NuGetModule` changes from the `release/1.0.0` branch, commit daf0142d8a18a52c8c29e05c387f92239a7c94c5 I think, to be actually useful. Without that, we're stuck creating our own `NuGetModule` in our `Program.Configure` using an emtpy `CakeConfiguration` object for now, as mentioned [here](https://github.com/cake-build/cake/discussions/2931).

patriksvensson:
@dieterv This has been fixed in #2946 which will be merged shortly.

devlead:
Fixed by #2946

dieterv:
@patriksvensson upgraded to 1.0.0-rc0002 and it looks like Frosting does not yet respect the settings from the [Paths] section of my cake.config file. Digging a bit deeper and it looks like the internal ToolInstaller still uses a hardcoded `_environment.WorkingDirectory.Combine("tools").MakeAbsolute(_environment);` path, see
https://github.com/cake-build/cake/blob/main/src/Cake.Frosting/Internal/ToolInstaller.cs#L30

Additionally, it no longer seems to take the working directory change into account as configured on the CakeHost instance as was the case in rc0001.
```
        // Create and run the host.
        return new CakeHost()
            .UseWorkingDirectory("../..")
            .UseContext<Context>()
            .UseLifetime<Lifetime>()
            .Run(args);
```

pascalberger:
> Additionally, it no longer seems to take the working directory change into account as configured on the CakeHost instance as was the case in rc0001.

[...]

PR #3048, opened 2021-01-23
Cake Frosting should respect configuration

patriksvensson:
This pull request makes sure that Cake.Frosting respect any user configuration related to the tool installation path.

### DIFF
--- a/src/Cake.Frosting/Extensions/CakeHostExtensions.cs
+++ b/src/Cake.Frosting/Extensions/CakeHostExtensions.cs
@@ -6,6 +6,7 @@ using System;
 using Cake.Core.Composition;
 using Cake.Core.IO;
 using Cake.Core.Packaging;
+using CoreConstants = Cake.Core.Constants;
 
 namespace Cake.Frosting
 {
@@ -152,7 +153,7 @@ namespace Cake.Frosting
         /// <param name="host">The <see cref="CakeHost"/> to configure.</param>
         /// <param name="uri">The tool URI.</param>
         /// <returns>The same <see cref="CakeHost"/> instance so that multiple calls can be chained.</returns>
-        public static CakeHost UseTool(this CakeHost host, Uri uri)
+        public static CakeHost InstallTool(this CakeHost host, Uri uri)
         {
             return host.ConfigureServices(services => services.UseTool(uri));
         }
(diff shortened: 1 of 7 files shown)


src/Cake.Frosting/Extensions/CakeHostExtensions.cs
```diff
@@ -152,7 +153,7 @@ public static CakeHost UsePackageInstaller<TPackageInstaller>(this CakeHost host
         /// <param name="host">The <see cref="CakeHost"/> to configure.</param>
         /// <param name="uri">The tool URI.</param>
         /// <returns>The same <see cref="CakeHost"/> instance so that multiple calls can be chained.</returns>
-        public static CakeHost UseTool(this CakeHost host, Uri uri)
+        public static CakeHost InstallTool(this CakeHost host, Uri uri)
```
augustoproiete, line 156: LGTM. I'm assuming v1.0 we're OK with breaking public APIs

patriksvensson: Yeah, we're already breaking some APIs 😅

augustoproiete:
I tested this one locally on my machine, and everything works (in the sense that tools are installed), but it doesn't honor the Tools Path in `cake.config`. The tools are being installed in a `tools` folder, but my `cake.config` tells it to use a `.cake` folder.

I tried:
- Having a single `cake.config` side-by-side with the Frosting `build.csproj`
- Having a single `cake.config` at the root of the project

Test code is on this branch:
https://github.com/augustoproiete-forks/cake-build--cake/tree/feature/GH-2904-test1/src/build

patriksvensson:
Frosting does not support cake.config.

gep13:
Based on the comments in the associated issue, it seems that there is an expectation that the cake.config file will be honoured.

patriksvensson:
Ok, then we need to add support for cake.config which has never been supported in Frosting before, and I don't think it should tbh.

gep13:
@patriksvensson the title of the issue that this PR is associated with is:

> (Frosting) Tool installer should respect configuration

If not cake.config, what configuration is this issue referring to? There isn't any details in the issue.

patriksvensson:
Yeah, maybe I'm wrong here.

src/Cake.Frosting/Extensions/CakeHostExtensions.cs
pascalberger: We should create an additional issue for it though so that it will be listed in release notes.

augustoproiete: @pascalberger +1 https://github.com/cake-build/cake/issues/3050 & https://github.com/cake-build/website/issues/1504

pascalberger (approved):
LGTM

pascalberger:
This fixes https://github.com/cake-build/cake/issues/3038. I removed link to https://github.com/cake-build/cake/issues/2904 since it does not add support for configuration and this is something which we decided to move post 1.0.

pascalberger:

PR #3059, opened 2021-01-28
GH2904: Frosting cake.config&tool config support

### LINKED ISSUE #2904: (Frosting) Tool installer should respect configuration

devlead:
* Frosting utilize CakeConfigurationProvider, enables config
    * Frosting IServiceCollection.UseToolPath
    * cake.config file
    * environment variables
    * command line arguments
* Tools path should be resolved using configuration
* fixes #2904

### DIFF
--- a/src/Cake.Core/Configuration/CakeConfigurationProvider.cs
+++ b/src/Cake.Core/Configuration/CakeConfigurationProvider.cs
@@ -4,6 +4,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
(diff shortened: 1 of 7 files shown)


pascalberger (changes requested):
We should move this post 1.x and discuss first what approach we want to take with configuration in Frosting: Support it or not, only cake.config or another more extensible approach, if cake.config which settings, have update to documentation in place, etc. To do this properly would move a 1.0 release further out and we're currently at a place where we're ready to ship if we don't open a whole new can of worms.

PR #3095, opened 2021-02-07
GH2904: Frosting cake.config&tool config support

### LINKED ISSUE #2904: (Frosting) Tool installer should respect configuration

devlead:
* Frosting utilize CakeConfigurationProvider, enables config
    * Frosting IServiceCollection.UseToolPath
    * cake.config file
    * environment variables
    * command line arguments
* Tools path should be resolved using configuration
* fixes #2904

### DIFF
--- a/src/Cake.Core/Configuration/CakeConfigurationProvider.cs
+++ b/src/Cake.Core/Configuration/CakeConfigurationProvider.cs
@@ -4,6 +4,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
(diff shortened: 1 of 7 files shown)


augustoproiete (approved):
LGTM!

Tested installing tools with Frosting with and without a `cake.config`, and confirmed that it now honors the tool path defined in the `cake.config`

Also confirmed `cake.config` continues to be optional, and the tool path continues to default to `./tools` if the there's no `cake.config` (or if the setting is not specified in it)

PR #3114, opened 2021-02-10
Add GetArgumentOrDefault to extensions

BlythMeister:
Whilst this is a simple function, it makes intent on usage cleaner.


BlythMeister:
Realised i should be using the Argument alias which already has this functionality :)

augustoproiete:
@BlythMeister There's room for improvement though. What do you think about [this](https://github.com/augustoproiete/Cake.Args)?

BlythMeister:
To be honest, I have found the context.Argument("key", "default") to be perfect for my use case

Issue #3092, opened 2021-02-06
Add LaunchDirectory property to ICakeEnvironment


augustoproiete:
`ICakeEnvironment.Environment` currently exposes `WorkingDirectory` which returns the directory path where the `.cake` script is located.

This issue is about adding a `LaunchDirectory` property to `ICakeEnvironment` that returns the original working directory where the Cake runner was invoked from, which could be a different folder than where the `.cake` script is located.

Examples:

| Cake Runner Call | WorkingDirectory (today) | LaunchDirectory (future) |
| --- | --- | --- |
| `C:\Projects>dotnet cake ProjectA\build\deploy.cake` | `C:\Projects\ProjectA\build` | `C:\Projects` |
| `C:\Projects\ProjectA>dotnet cake scripts\deploy.cake` | `C:\Projects\ProjectA\build` | `C:\Projects\ProjectA` |
| `C:\Projects\ProjectA\build>dotnet cake deploy.cake` | `C:\Projects\ProjectA\build` | `C:\Projects\ProjectA\build` |

---

Proposed by @mattleibow on discussion https://github.com/cake-build/cake/discussions/3076

Marusyk:
I will try to prepare a PR for this. Please assign it to me

PR #3132, opened 2021-02-13
GH-3092: Add LaunchDirectory property to ICakeEnvironment

Marusyk:
Added `LaunchDirectory` property to `ICakeEnvironment` that returns the original working directory where the Cake runner was invoked from, which could be a different folder than where the .cake script is located.

Issue: #3092


src/Cake.Core/CakeEnvironment.cs
```diff
@@ -48,6 +51,7 @@ public CakeEnvironment(ICakePlatform platform, ICakeRuntime runtime)
 
             // Get the working directory.
             WorkingDirectory = new DirectoryPath(System.IO.Directory.GetCurrentDirectory());
+            LaunchDirectory = Environment.CurrentDirectory;
```
augustoproiete, line 54: I haven't tested this yet, but I suspect that getting the `Environment.CurrentDirectory` here is too late (as it would reflect the directory where the Cake script is running). If this works, that would mean `WorkingDirectory` would also work for the same purpose, isn't it?

Marusyk: Sorry, it's not ready for review because I'm trying to find a way how to test it

devlead: Maybe an `ILaunchInfo` (with `LaunchDirectory` proprty)  non lazy instance that's registered and injected via DI, that would be as early as possible, singleton(so it wouldn't change) and easy to mock in tests.

Marusyk: Little bit confused about how to add ILaunchInfo to FakeEnvironment and test it. Could you please suggest because FakeEnvironment doesn't use DI.

augustoproiete: @Marusyk You can create a `FakeLaunchInfo` class and add a property to `FakeEnvironment` similar to how it's done for [`Platform`](https://github.com/cake-build/cake/blob/develop/src/Cake.Testing/FakeEnvironment.cs#L35) and [`Runtime`](https://github.com/cake-build/cake/blob/develop/src/Cake.Testing/FakeEnvironment.cs#L44), for example.

Marusyk: Thanks a lot. Could you please point me where is the proper place to test it?

src/Cake.Core/ICakeEnvironment.cs
```diff
@@ -20,6 +18,12 @@ public interface ICakeEnvironment
         /// <value>The working directory.</value>
         DirectoryPath WorkingDirectory { get; set; }
 
+        /// <summary>
+        /// Gets the launch info.
+        /// </summary>
+        /// <value>The launch info.</value>
+        ILaunchInfo LaunchInfo { get; }
```
augustoproiete, line 25:
```suggestion
        DirectoryPath LaunchDirectory { get; }
```

The `ILaunchInfo` is just an implementation detail to give you a way for you to store the `LaunchDirectory` early in the process, and inject into the places that need it. From the point of view of the API, for build scripts to access it, let's expose it in the same way that we do for `WorkingDirectory` above.

src/Cake.Testing/FakeEnvironment.cs
```diff
@@ -43,6 +43,15 @@ public sealed class FakeEnvironment : ICakeEnvironment
...
         public FakeRuntime Runtime { get; }
 
+        /// <inheritdoc/>
+        ILaunchInfo ICakeEnvironment.LaunchInfo => LaunchInfo;
+
+        /// <summary>
+        /// Gets launch info.
+        /// </summary>
+        /// <value>The runtime Cake is running in.</value>
+        public FakeLaunchInfo LaunchInfo { get; }
```
augustoproiete, line 53: Is there a technical reason you're having to implement this twice?

src/Cake.Core/CakeEnvironment.cs
```diff
@@ -22,6 +22,9 @@ public DirectoryPath WorkingDirectory
             set { SetWorkingDirectory(value); }
         }
 
+        /// <inheritdoc/>
+        public ILaunchInfo LaunchInfo { get; }
```
augustoproiete, line 26:
```suggestion
        /// <inheritdoc/>
        public DirectoryPath LaunchDirectory
        {
            get { return launchInfo.LaunchDirectory; }
        }
```

The `ILaunchInfo` is just an implementation detail to give you a way for you to store the `LaunchDirectory` early in the process, and inject into the places that need it. From the point of view of the API, for build scripts to access it, let's expose it in the same way that we do for `WorkingDirectory` above.

augustoproiete (changes requested):
Thanks, looking promising!

src/Cake.Core/IO/ILaunchInfo.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Cake.Core.IO
+{
+    /// <summary>
+    /// Represents a launch info.
+    /// </summary>
+    public interface ILaunchInfo
```
augustoproiete, line 10: Let's move this file to src/Cake.Core to be side-by-side with CakeEnvironment which is the class that uses it

src/Cake.Core/IO/LaunchInfo.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Cake.Core.IO
+{
+    /// <inheritdoc/>
+    public sealed class LaunchInfo : ILaunchInfo
```
augustoproiete, line 10: Lets make this class `internal` as build scripts shouldn't need to use this directly.

augustoproiete, line 10: Let's also move this file to src/Cake.Core to be side-by-side with CakeEnvironment which is the class that uses it

src/Cake.Testing/FakeEnvironment.cs
```diff
@@ -43,6 +43,15 @@ public sealed class FakeEnvironment : ICakeEnvironment
         /// <value>The runtime Cake is running in.</value>
         public FakeRuntime Runtime { get; }
 
+        /// <inheritdoc/>
+        ILaunchInfo ICakeEnvironment.LaunchInfo => LaunchInfo;
```
augustoproiete, line 47: Is there a technical reason you're having to implement this twice? [See comment](https://github.com/cake-build/cake/pull/3132/files#r587071298)

src/Cake.Core/CakeEnvironment.cs
augustoproiete: Given that you're introducing a new property in `src/Cake.Core/CakeEnvironment.cs` the best place to add unit tests would be `src/Cake.Core.Tests/Unit/CakeEnvironmentTests.cs` (which doesn't exist yet, so you'll need to create it)

src/Cake.Core/IO/LaunchInfo.cs
devlead: @augustoproiete Would bridge / frosting need this, in that case public

augustoproiete: @devlead Hmm... Good point. I thought the `ILaunchInfo` interface was enough and they could have their own implementations, but we're sharing a lot of concrete implementations already.

[Frosting uses it](https://github.com/cake-build/cake/pull/3132/files#diff-a4fe3eef0c77283c6481738eb652efa3f56018d9a316dbb4575b0474038abd9fR85), so it's safe to assume Bridge would too.

If everything will be public, maybe better to rename to `ICakeLaunchInfo`/`CakeLaunchInfo` for consistency? (CakeEnvironment, CakePlatform, CakeRuntime, ...)

src/Cake.Testing/FakeEnvironment.cs
Marusyk: I followed the same pattern as `Runtime` and `Platform`

src/Cake.Testing/FakeEnvironment.cs
```diff
@@ -52,6 +55,10 @@ public FakeEnvironment(PlatformFamily family, bool is64Bit = true)
         {
             Platform = new FakePlatform(family, is64Bit);
             Runtime = new FakeRuntime();
+
+            var launchInfo = new FakeLaunchInfo();
+            LaunchDirectory = launchInfo.LaunchDirectory;
```
augustoproiete, line 60: I believe you'll need to expose `FakeLaunchInfo` in this `FakeEnvironment` in order to manipulate it during your unit tests (as per your previous comment on following the pattern of Runtime). Sorry if I caused any confusion with my question about why you needed the two implementations - it makes sense to expose the `Fake` to your tests

src/Cake.Testing/FakeEnvironment.cs
```diff
@@ -43,6 +43,9 @@ public sealed class FakeEnvironment : ICakeEnvironment
         /// <value>The runtime Cake is running in.</value>
         public FakeRuntime Runtime { get; }
 
+        /// <inheritdoc/>
+        public DirectoryPath LaunchDirectory { get; set; }
```
augustoproiete, line 47:
```suggestion
        public DirectoryPath LaunchDirectory { get; }
```

I believe this should match the interface exactly (so no setter), and you can use `FakeLaunchInfo` to set up your mock

src/Cake.Core/Modules/CoreModule.cs
```diff
@@ -36,6 +36,7 @@ public void Register(ICakeContainerRegistrar registrar)
             registrar.RegisterType<DefaultExecutionStrategy>().As<IExecutionStrategy>().Singleton();
 
             // Environment
+            registrar.RegisterType<LaunchInfo>().As<ILaunchInfo>().Singleton();
```
augustoproiete, line 39: I could be wrong, but I don't think you need to register `LaunchInfo` again here - I'd expect the same instance to flow through to this scope. Could you take a look?

If it's really needed, then I think we need to find a way to get the instance created at the start and pass it through here, so you can register the same instance, because creating a new instance of `LaunchInfo` here might be too late because bootstrapping could already have happened and changed the working directory.

src/Cake/Program.cs
```diff
@@ -96,6 +96,7 @@ private ITypeRegistrar BuildTypeRegistrar()
             builder.RegisterType<InfoFeature>().As<ICakeInfoFeature>().SingleInstance();
 
             // Core
+            builder.RegisterType<LaunchInfo>().As<ILaunchInfo>().SingleInstance();
```
augustoproiete, line 99: This registration looks good, but it doesn't actually causes `LaunchInfo` to be instantiated at this time - it will only be instantiated later when some dependency needs it (i.e. `CakeEnvironment`).

Now, this _might_ work currently because the `CakeEnvironment` is instantiated early in the pipeline which should cause the `LaunchInfo` to be instantiated early enough as well - but it's not 100% guaranteed it will continue like that in the future (e.g. if we introduce other dependencies)

I'd prefer to play safe here and create the instance of `LaunchInfo` ourselves at the very beginning, and register it into the container, so that know for sure that we captured the `CurrentDirectory` at the right time.

e.g.
```csharp
private ITypeRegistrar BuildTypeRegistrar()
{
    var builder = new ContainerBuilder();

    var launchInfo = new LaunchInfo(); // We're 100% sure we captured the working directory within `launchInfo`
    builder.RegisterInstance(launchInfo).As<ILaunchInfo>();

     // ...
```

augustoproiete (changes requested):
Looking great

src/Cake.Core.Tests/Unit/CakeEnvironmentTests.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public sealed class CakeEnvironmentTests
+    {
+        public sealed class LaunchDirectoryProperty
+        {
+            [Theory]
+            [InlineData(PlatformFamily.Linux)]
+            [InlineData(PlatformFamily.OSX)]
+            [InlineData(PlatformFamily.Windows)]
+            public void Should_Not_Changes_When_WorkingDirectory_Changed(PlatformFamily family)
```
augustoproiete, line 20: nit: typo on `Changes` -> `Change`

src/Cake.Testing/FakeEnvironment.cs
Marusyk: I'm not sure how I can have two implementations because we don't have the `ILaunchInfo` property anymore in `ICakeEnvironment`

augustoproiete:
Hey @Marusyk are you planning on reopening/getting back to one?

Marusyk:
Hi @augustoproiete 
No, I didn't get answers to my questions here for a long time. I don't know how to implement some things after your review. 
I can continue to work on it if you help me

augustoproiete:
@Marusyk Cool, I'll reopen then. I take a deeper look in a few days as time permits, and let you know.

lastthyme (approved):

Marusyk:
Hi @augustoproiete
When we can return to this PR?

PR #3134, opened 2021-02-13
GH3133: Add IncludeNativeLibrariesForSelfExtract and IncludeAllContentForSelfExtract to DotNetCorePublishSettings

### LINKED ISSUE #3133: Add IncludeNativeLibrariesForSelfExtract and IncludeAllContentForSelfExtract to DotNetCorePublishSettings
`dotnet publish` in .NET 5 introduces two new settings: `IncludeNativeLibrariesForSelfExtract` and `IncludeAllContentForSelfExtract` which are not yet exposed in `DotNetCorePublishSettings` as of Cake v1.0.0.

### Docs

> Single-file doesn't bundle native libraries by default. On Linux, we prelink the runtime into the bundle and only application native libraries are deployed to the same directory as the single-file app. On Windows, we prelink only the hosting code and both the runtime and application native libraries are deployed to the same directory as the single-file app. This is to ensure a good debugging experience, which requires native files to be excluded from the single file. There is an option to set a flag, IncludeNativeLibrariesForSelfExtract, to include native libraries in the single file bundle, but these files will be extracted to a temporary directory in the client machine when the single file application is run.
> 
> Specifying IncludeAllContentForSelfExtract will extract all files before running the executable. This preserves the original .NET Core single-file deployment behavior.
> 

Source: https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file#other-considerations

### Workaround

As with every tool setting in Cake, `ArgumentCustomization` is your friend:

```csharp
DotNetCorePublish("./src/ExcelDnaUnpack/ExcelDnaUnpack.csproj", new DotNetCorePublishSettings
{
    Framework = "net5.0",
    Runtime = "win7-x64",
    Configuration = "Release",
    SelfContained = true,
    PublishSingleFile = true,
    PublishTrimmed = true,
    OutputDirectory = "./artifacts",
    ArgumentCustomization = args =>
        args.Append("-p:IncludeNativeLibrariesForSelfExtract=true")
});
```

augustoproiete:
Resolves #3133

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Publish/DotNetCorePublisherTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/DotNetCore/Publish/DotNetCorePublisherTests.cs
@@ -261,6 +261,20 @@ namespace Cake.Common.Tests.Unit.Tools.DotNetCore.Publish
                 // Then
                 Assert.Equal("publish -p:IncludeNativeLibrariesForSelfExtract=true", result.Args);
             }
+
+            [Fact]
+            public void Should_Add_IncludeAllContentForSelfExtract()
+            {
+                // Given
+                var fixture = new DotNetCorePublisherFixture();
+                fixture.Settings.IncludeAllContentForSelfExtract = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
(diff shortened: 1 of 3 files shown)


devlead (approved):
LGTM 👍

devlead:

PR #3152, opened 2021-02-16
GH-3151: Add events to ICakeEngine to invoke at the end of Setup/Teardown

### LINKED ISSUE #3151: Add support for Engine event hooks after execution as well as before
Currently the CakeEngine will fire events before running:

* Setup
* TaskSetup
* TaskTeardown
* Teardown

Add support for event hooks after each of these methods have fired.

This is to allow Modules (such as Cake.BuildSystems.Module) to provide a custom engine which performs an action after the method has fired as well as before.

BlythMeister:
This allows for any logic required after each section to be executed.

Fixes #3151

### DIFF
--- a/src/Cake.Core/CakeEngine.cs
+++ b/src/Cake.Core/CakeEngine.cs
@@ -27,16 +27,48 @@ namespace Cake.Core
         public IReadOnlyList<ICakeTaskInfo> Tasks => _tasks;
 
         /// <inheritdoc/>
+#pragma warning disable 618
         public event EventHandler<SetupEventArgs> Setup;
+#pragma warning restore 618
 
         /// <inheritdoc/>
+        public event EventHandler<BeforeSetupEventArgs> BeforeSetup;
+
+        /// <inheritdoc/>
+        public event EventHandler<AfterSetupEventArgs> AfterSetup;
+
+        /// <inheritdoc/>
+#pragma warning disable 618
         public event EventHandler<TeardownEventArgs> Teardown;
+#pragma warning restore 618
 
         /// <inheritdoc/>
+        public event EventHandler<BeforeTeardownEventArgs> BeforeTeardown;
+
+        /// <inheritdoc/>
+        public event EventHandler<AfterTeardownEventArgs> AfterTeardown;
+
+        /// <inheritdoc/>
+#pragma warning disable 618
         public event EventHandler<TaskSetupEventArgs> TaskSetup;
+#pragma warning restore 618
 
         /// <inheritdoc/>
+        public event EventHandler<BeforeTaskSetupEventArgs> BeforeTaskSetup;
+
+        /// <inheritdoc/>
+        public event EventHandler<AfterTaskSetupEventArgs> AfterTaskSetup;
+
+        /// <inheritdoc/>
+#pragma warning disable 618
         public event EventHandler<TaskTeardownEventArgs> TaskTeardown;
+#pragma warning restore 618
+
+        /// <inheritdoc/>
+        public event EventHandler<BeforeTaskTeardownEventArgs> BeforeTaskTeardown;
+
+        /// <inheritdoc/>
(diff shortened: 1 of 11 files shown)


augustoproiete:
@BlythMeister Just to give you some feedback on this one, we discussed this PR internally and given that it introduces a breaking change to the public API (i.e. `ICakeEngine`) which would impact all extensions, we'll consider it for Cake v2.0.0 later this year, at which time we'll engage with you to understand more about the use-case of these events and test some examples together.

pascalberger:
Added a comment to the issues where to discuss the requirements and design first: https://github.com/cake-build/cake/issues/3151#issuecomment-945202748

nils-a (dismissed):
@BlythMeister thank you for the contribution 👍 

I added some requests for changes.

BlythMeister:
Yeah perfectly fine with me.  It's been on my to-do list...but work has been super busy!

gep13 (approved):
LGTM!

src/Cake.Core/CakeEngine.cs
```diff
@@ -232,16 +264,35 @@ private async Task RunTask(ICakeContext context, IExecutionStrategy strategy, Ca
...
+                    foreach (var setup in _actions.Setups)
+                    {
+                        PerformTaskSetup(context, strategy, targetTask, false);
+                        try
+                        {
+                            strategy.PerformSetup(setup, new SetupContext(context, targetTask, tasks));
+                        }
+                        catch (Exception e)
+                        {
+                            Console.WriteLine(e);
```
augustoproiete, line 285: This should not write to the Console. We should Log the exception to the logger - or remove the Try Catch entirely and let it bubble up

augustoproiete: Apologies if I'm missing something here, but my understanding is that the Setup runs only once at the start and before any TaskSetup, and TaskSetup runs on every task - Here it seems we're running TaskSetup multiple times and _before_ Setup, instead of after.

src/Cake.Core/ICakeEngine.cs
```diff
@@ -20,25 +20,73 @@ public interface ICakeEngine
...
 
+        /// <summary>
+        /// Raised at the start of task teardown after each task has been run.
+        /// </summary>
+        event EventHandler<BeforeTaskTeardownEventArgs> BeforeTaskTeardown;
+
+        /// <summary>
+        /// Raised at the end of task teardown after each task has been run.
+        /// </summary>
+        event EventHandler<AfterTaskTeardownEventArgs> AfterTaskTeardown;
```
augustoproiete, line 88: Do we want to expose these events in the interface `ICakeEngine`?

augustoproiete (changes requested):
Looks good. Added a couple of comments

BlythMeister:
Who is to resolve the comments?
I'm now not sure what I did originally and what was modified.

If you want to wait for me, I'm not sure what timeline that will be, as I am super busy at work at the moment.

src/Cake.Core/ICakeEngine.cs
BlythMeister: These need to be on the interface to allow for things like the teamcity logger to log "end blocks" as service messages.

src/Cake.Core/CakeEngine.cs
pascalberger: I removed the try/catch, and removed the `PerformTaskSetup` call, so that logic is not changed

pascalberger:
> Who is to resolve the comments? I'm now not sure what I did originally and what was modified.
> 
> If you want to wait for me, I'm not sure what timeline that will be, as I am super busy at work at the moment.

@BlythMeister That's fine. I took care of resolving the comments

augustoproiete (approved):
LGTM!

augustoproiete:
@BlythMeister your changes have been merged, thanks for your contribution :+1:

Issue #2903, opened 2020-10-16
Integrate Cake.DotNetTool.Module


pascalberger:
To install .NET tools the Cake.DotNetTool.Module can be used. Using .NET tools in a build becomes a more and more common pattern. We should think about integrating functionality of Cake.DotNetTool.Module in Core Cake Engine.

Originally reported in https://github.com/cake-build/cake/issues/2685

devlead:
Fixed by #3207

PR #3207, opened 2021-02-27
GH2685/2903: Integrate Cake.DotNetTool.Module 

### LINKED ISSUE #2685: Better dotnet tool integration
It would be great to have better dotnet tool integration, especially now that we have [local tools in .NET Core 3.0](https://andrewlock.net/new-in-net-core-3-local-tools/#local-tools-in-net-core-3-0). I've been pondering what a _minimal_ Cake build could look like. My [BoardGameGeek.Dungeon](https://github.com/gitfool/BoardGameGeek.Dungeon) repo can serve as the use case.

Current Cake related files:
- [cake.ps1](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/cake.ps1) bootstraps Cake on Windows
- [cake.sh](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/cake.sh) bootstraps Cake on Linux
- [tools/packages.config](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/tools/packages.config) locks Cake version

For [Azure Pipelines](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/.azure-pipelines.yml#L30) and [GitHub Actions](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/.github/workflows/build.yml#L28) builds I avoid the Cake bootstrap by using a [Docker image](https://github.com/gitfool/cake-docker/blob/master/Dockerfile#L10) with the Cake tool already installed. However, all builds still have to bootstrap Cake _modules_ in order to use dotnet tools in the Cake build itself, one of the most common being [GitVersion](https://github.com/GitTools/GitVersion).

BoardGameGeek.Dungeon is built using [Cake.Dungeon](https://github.com/gitfool/Cake.Dungeon), and in the depths lives the [bootstrap.cake](https://github.com/gitfool/Cake.Dungeon/blob/master/scripts/bootstrap.cake) script with the following:
```c#
#module nuget:?package=Cake.DotNetTool.Module&version=0.4.0
...
#tool dotnet:?package=GitVersion.Tool&version=5.1.2
```
This is the first barrier to diet Cake nirvana as it requires Cake to be invoked twice, first to install the `Cake.DotNetTool.Module` to extend Cake to support `#tool`, then again for the actual build, using something like the following:
```
cake --bootstrap && cake
```
Not too bad, although more verbose in [PowerShell](https://github.com/PowerShell/PowerShell) until v7 drops next year with long overdue support for [pipeline chain operators](https://github.com/PowerShell/PowerShell/releases/tag/v7.0.0-preview.5), but really it's surely about time this module moved into Cake core, and when it does we'll want it to play nice with dotnet local tools. (The two step bootstrap will still be required if you're using other Cake modules, but I'd argue that's much less common, and using dotnet [local] tools is standard now, especially if writing cross-platform builds.)

Anyway, given support for dotnet local tools, I would be able to delete the above Cake related files and replace them with a single `dotnet-tools.json` manifest file:
```json
{
  "version": 1,
  "isRoot": true,
  "tools": {
    "cake.tool": {
      "version": "0.36.0",
      "commands": [
        "dotnet-cake"
      ]
    }
  }
}
```
Then to invoke the build on a clean machine I'd use the following:
```
dotnet tool restore && dotnet cake --bootstrap && dotnet cake
```
... or when the dotnet tool module is moved to Cake core:
```
dotnet tool restore && dotnet cake
```
... and when the Cake dotnet tool is already installed (on a local dev machine):
```
dotnet cake
```
Also, for this to integrate better with dotnet _local_ tools, the `#tool` directive should use the local tool directly, presumably invoking `dotnet <tool>`, rather than extracting and copying what was previously a global tool into the Cake tools directory alongside addins and modules. (I'm hoping such a change would also avoid the quirk resulting from building the same directory in Windows and then WSL or Docker with a host mount, or vice versa, where you first need to explicitly delete any such copied dotnet tools which would be for the wrong platform.)

So in summary, a minimal Cake build would simply leverage dotnet local tools to bootstrap Cake itself, and given the dotnet tool module was moved to Cake core, a common case requiring an extra module bootstrap step could most likely be avoided, and the `#tool` directive would use the dotnet local tool directly, rather than copying a platform specific executable into its own tools directory.

devlead:
Fixes  #2685 and #2903

- [x] Filter out CI files (GitHub/AppVeyor/Recipe etc.)
- [x] Rebase into Cake repo
- [x] Move into Cake solution/src
- [x] Adhere to Cake code style
- [x] Load by default in Scripting & Frosting
- [x] Adapt/add integration tests
- [x] Warn if Cake.DotNetTool.Module referenced as module
- [x] Fix posix builds
- [x] Push to NuGet
- [ ] Add Cake as owner of package


augustoproiete (approved):
LGTM!

Tested with a number of .NET tools and it works as expected:

```csharp
#tool dotnet:?package=GitVersion.Tool&version=5.1.2
#tool dotnet:?package=exceldna-unpack&version=2.1.0
#tool dotnet:?package=minver-cli&version=2.4.0
#tool dotnet:?package=dotnet-reportgenerator-globaltool&version=4.8.6
```

### Test 1 - First run without any tools installed installs them as expected:

image

### Test 2 - Second run with the tools already installed displays informational message that tools are already installed as expected
(perhaps something we could make verbose in a future PR?)

image

### Test 3 - Deleted tools folder, added cake.config with different path for tools folder, installs as expected honoring the cake.config tools path

cake.config

```ini
[Paths]
Tools=./MyCustomTools
Addins=./MyCurtomAddins
Modules=./MyCustomModules
```

image

augustoproiete:
Thanks a lot for taking this over the line @devlead! And to @gep13 for the actual implementation!

Another successful incubator project that made it into the mainline 🚀

PR #3212, opened 2021-02-28
Add support for dotnet local tools

### LINKED ISSUE #3209: DotNetTool Module: Add support for dotnet local tools
> Creating a new issue for tracking https://github.com/cake-contrib/Cake.DotNetTool.Module/issues/18 created by @gitfool given that GitHub doesn't allow transferring issues from one organization to another.

---

@gitfool wrote:

Note: I'm not sure this is a good idea and I'm currently stuck determining what the implementation looks like given local tools seem to be managed by the dotnet CLI.

I'll create a work-in-progress PR with how far I've got, including fixes for some minor bugs I found.

gitfool:
Fixes #3209; migrated from https://github.com/cake-contrib/Cake.DotNetTool.Module/pull/19.


src/Cake.DotNetTool.Module/DotNetToolContentResolver.cs
```diff
@@ -73,6 +73,10 @@ public IReadOnlyCollection<IFile> GetFiles(PackageReference package, PackageType
                         return GetToolFiles(new DirectoryPath(_environment.GetEnvironmentVariable("USERPROFILE")).Combine(".dotnet/tools"), package);
                     }
                 }
+                else if (package.Parameters.ContainsKey("local"))
+                {
+                    return new List<IFile>(); // TODO: implementation
```
gitfool, line 78: This is the hard part. Anyone got any suggestions?

src/Cake.DotNetTool.Module/DotNetToolPackageInstaller.cs
```diff
@@ -207,7 +215,8 @@ private List<DotNetToolPackage> GetInstalledTools(string toolLocation)
                     {
                         Id = match.Groups["packageName"].Value,
                         Version = match.Groups["packageVersion"].Value,
-                        ShortCode = match.Groups["packageShortCode"].Value
+                        ShortCode = match.Groups["packageShortCode"].Value,
+                        Manifest = match.Groups["packageManifest"].Value
```
gitfool, line 219: I've done no testing so far, but expecting this to be empty and not throw when not matched.

src/Cake.DotNetTool.Module/DotNetToolPackageInstaller.cs
```diff
@@ -196,7 +204,7 @@ private List<DotNetToolPackage> GetInstalledTools(string toolLocation)
             var installedTools = isInstalledProcess.GetStandardOutput().ToList();
             var installedToolNames = new List<DotNetToolPackage>();
 
-            const string pattern = @"(?<packageName>[^\s]+)\s+(?<packageVersion>[^\s]+)\s+(?<packageShortCode>[^`s])";
+            const string pattern = @"(?<packageName>[^\s]+)\s+(?<packageVersion>[^\s]+)\s+(?<packageShortCode>[^\s]+)(?:\s+(?<packageManifest>[^\s]+))?";
```
gitfool, line 207: Optionally match the package manifest, when available, for local tools.

This regex had a bug; `` `s `` should be `\s` for non-whitespace match and it also didn't match more than one character.

src/Cake.DotNetTool.Module/DotNetToolPackageInstaller.cs
```diff
@@ -237,7 +246,7 @@ private void RunDotNetTool(PackageReference package, DirectoryPath toolsFolderDi
             {
                 _log.Warning("dotnet exited with {0}", exitCode);
                 var output = string.Join(Environment.NewLine, process.GetStandardError());
-                _log.Verbose(Verbosity.Diagnostic, "Output:\r\n{0}", output);
+                _log.Verbose(Verbosity.Diagnostic, "Output:{0}{1}", Environment.NewLine, output);
```
gitfool, line 249: There was another minor bug here. Changed to use `Environment.NewLine` when composing output message.

gitfool:
@augustoproiete it would be good to get some eyes on this and any ideas on how to actually implement it. 😛

src/Cake.DotNetTool.Module/DotNetToolContentResolver.cs
devlead: This is a good question, getting and executing tools from global  NuGet cache is possible quick PoC gist here
https://gist.github.com/devlead/559d9b082c9cd81c5c5d52f94d68380c

Don't know if that helps, but potentially one could return entry points 🤔

PR #3214, opened 2021-02-28
GH-3213: Add iconUrl fallback to Cake and Cake.CoreCLR NuGet packages

### LINKED ISSUE #3213: Add iconUrl fallback to Cake and Cake.CoreCLR packages
The Cake and Cake.CoreCLR packages have an embedded icon, but don't have the `iconUrl` set, which means that clients that don't support the embedded icon don't have an icon to fallback to, and display the default NuGet icon.

augustoproiete:
Closes #3213


devlead (approved):
LGTM 👍

Issue #3223, opened 2021-03-02
Feature request: Environment variable substitution in cake.config


epaulsen:
Given that [nuget.config supports environment variable substitution ](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#environment-variables-in-configuration), adding the same to cake.config would help out in scenarios where you have more than one package repository(we have a separate package repository for an off-shore contractor).

We would like to control which package repository is used via environment variables, but cannot for cake build, because cake.config does not support this.  

I'd be happy to create a PR for this if needed.

gep13:
@epaulsen can you elaborate on exactly what you are trying to achieve here?

What options are you attempting to control within the cake.config file?  Cake already supports providing parameters via command line argument, cake.config and also environment variables.

epaulsen:
Here's what we are trying to do:
```
[Nuget]
Source=nuget/local;https://packages.nuget.org/api/v2;https://%PACKAGE_HOME%/api/nuget/InternalNugetTools
UseInProcessClient=true

```
`https://%PACKAGE_HOME%/api/nuget/InternalNugetTools` point to one of two internal internal artifact repositories, we would like to control which via environment variable.  

But if there is another and perhaps better way of doing this please let me know.

gep13:
The NuGet Source parameter can be controlled, in it's entirity, through a single environment variable called `CAKE_NUGET_SOURCE`.

https://cakebuild.net/docs/running-builds/configuration/default-configuration-values#nuget-download-url

But you can set it to whatever you want.  You can read more about this here:

https://cakebuild.net/docs/running-builds/configuration/set-configuration-values

epaulsen:
Thanks for the links, I must have missed that in the docs.  
I've resolved our problem by adding two environment variables, `CAKE_NUGET_SOURCE` and another one with just the hostname (`PACKAGE_HOME` in my example above).
We needed both because the format in nuget.config is different from cake.config.

Not as elegant as it would have been if cake.config supported environment variable substitution(like nuget does), so I'll leave this issue open for now.  Feel free to close it if this is not something you would like to support.

Thanks for the help!

augustoproiete:
Thanks @epaulsen we're considering adding environment variable expansion to `cake.config` values in a future release. Would you be interested in sending a PR?

epaulsen:
@augustoproiete Yes absolutely!  We're using cake extensively, and I'd be happy contribute 😄 
[...]

PR #3234, opened 2021-03-05
Issue #3223: Adding environment variable substitution support for cake.config

epaulsen:
This pullrequests adds basic environment variable substitution to `cake.config`, as requested in #3223 

I've used the windows syntax for environment variables, as that is what nuget does for `nuget.config`.
([See Microsoft docs on this topic](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#environment-variables-in-configuration))
The effect is that environment variable substitution will work cross-platform, but windows environment variable syntax must be used.  This might look somewhat odd in unix environments, but can be justified with the .ini syntax in cake.config 😄 

This should probably be documented somehow, but I can't find any documentation in this repo?

### DIFF
--- a/src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
+++ b/src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
@@ -37,11 +37,11 @@ namespace Cake.Core.Configuration.Parser
             using (var stream = file.OpenRead())
             using (var reader = new StreamReader(stream))
             {
-                return Read(reader.ReadToEnd());
+                return Read(_environment, reader.ReadToEnd());
             }
         }
 
-        private static IDictionary<string, string> Read(string text)
+        private static IDictionary<string, string> Read(ICakeEnvironment environment, string text)
         {
             var tokens = ConfigurationTokenizer.Tokenize(text);
             var section = string.Empty;
@@ -56,7 +56,7 @@ namespace Cake.Core.Configuration.Parser
                         break;
                     case ConfigurationTokenKind.Value:
                         var pair = ParseKeyAndValue(tokens, section);
-                        result[pair.Key] = pair.Value;
+                        result[pair.Key] = environment.ExpandEnvironmentVariables(pair.Value);
                         break;
                     default:
                         throw new InvalidOperationException("Encountered unexpected token.");
--- a/src/Cake.Core.Tests/Unit/Configuration/Parser/ConfigurationParserTests.cs
+++ b/src/Cake.Core.Tests/Unit/Configuration/Parser/ConfigurationParserTests.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using Cake.Core.Configuration.Parser;
 using Cake.Core.Tests.Properties;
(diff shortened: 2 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3234) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3234) before we can accept your contribution.<br/><br/>:x: epaulsen [sign now](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3234)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/cake-build/cake?pullRequest=3234) it.</sub>

src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
```diff
@@ -37,7 +37,7 @@ public ConfigurationParser(IFileSystem fileSystem, ICakeEnvironment environment)
             using (var stream = file.OpenRead())
             using (var reader = new StreamReader(stream))
             {
-                return Read(reader.ReadToEnd());
+                return Read(EnvironmentVariableSubstitution.Substitute(_environment, reader.ReadToEnd()));
```
augustoproiete, line 40: I took a cursory look at your `EnvironmentVariableSubstitution.Substitute` and it seems that it's re-implementing an existing .NET feature -> [`Environment.ExpandEnvironmentVariables`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.expandenvironmentvariables?view=net-5.0) is that correct?

If that's true, could we use the built-in function instead? Otherwise, what's different/better about this implementation?

augustoproiete:
Thanks for the PR @epaulsen!

Re: docs, I think a good place to explain we support environment variable expansion would be in the docs of the website: https://cakebuild.net/docs/running-builds/configuration/set-configuration-values

The site is on a separate repo: https://github.com/cake-build/website

The config docs are here, in markdown: https://github.com/cake-build/website/tree/master/input/docs/running-builds/configuration

src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
epaulsen: Oh.My.Lord.  Well, this is awkward... 
I've been programming .net for the last twenty years, and I wasn't aware that the framework had the functionalty built-in.
Oh well, at least I learned something by doing this 😁 

I'll fix this, I think using built-in functionality would be better.

epaulsen: Rewritten to use built-in functionality instead

src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
```diff
@@ -37,7 +37,7 @@ public ConfigurationParser(IFileSystem fileSystem, ICakeEnvironment environment)
             using (var stream = file.OpenRead())
             using (var reader = new StreamReader(stream))
             {
-                return Read(reader.ReadToEnd());
+                return Read(Environment.ExpandEnvironmentVariables(reader.ReadToEnd()));
```
augustoproiete, line 40: The scope of the expansion here seems a little too broad, and I'm not 100% sure if it will work for expanding multiple variables on all platforms, considering different line-endings, and the tests don't cover that.

With the current scope, the following scenarios would be possible:

```ini
[Section]
%A%=%B%
```

or even

```ini
[%A%]
%B%=%C%
```

I'm not sure we'd want to support expansion on the whole file rather than on invididual `values` of each field... We probably only want this one scenario to work:

```ini
[Section]
Key=%VALUE%
```

@cake-build/cake-team What do you think?

I think it would be safer to perform the expansion [directly on each value](https://github.com/cake-build/cake/blob/04a98f8e25f4f676f99edb9aef9ac3e724eb809d/src/Cake.Core/Configuration/Parser/ConfigurationParser.cs#L59) as we parse individual entries:

```csharp
case ConfigurationTokenKind.Value:
     var pair = ParseKeyAndValue(tokens, section);
     result[pair.Key] = _environment.ExpandEnvironmentVariables(pair.Value);
     break;
```

The other good thing is that the [`ConfigurationParser` already has access to `ICakeEnvironment`](ExpandEnvironmentVariables), so we can use that instead of calling `Environment` directly.

augustoproiete (changes requested):
Looking good!

src/Cake.Core/Configuration/Parser/ConfigurationParser.cs
epaulsen: Ok, that's reasonable and should cover my needs.
Should be fairly easy to implement.

epaulsen: Done, now changed to using environment variable substitution for values only.  
Sorry about the force-push, I am not used to the workflow for pullrequests on Github repositories.

epaulsen:
@augustoproiete - Done(I think).  
There's a build failure on AppVeyor which seems unrelated to this PR.

Should I squash the commits in to a single commit or leave it as-is?

augustoproiete:
Thanks @epaulsen! I think this is it!
If you could squash the commits and rebase on top of `develop` that would help (and might even fix the AppVeyor build).

---

@cake-build/cake-team Could I get another pair of eyes on this one?

epaulsen:
> Thanks @epaulsen! I think this is it!
> If you could squash the commits and rebase on top of `develop` that would help (and might even fix the AppVeyor build).

Rebased and squashed into single commit 👍 
But that didn't fix the appveyor build, which seems to be failing on wrong .net core sdk in `global.json`
Fairly certain that I didn't break that, so ignoring it for now.

augustoproiete (approved):
LGTM!

The AppVeyor build issue is unrelated. @devlead suggested that it's probably caused because of the branch name having only numbers (the branch name is used to generate version numbers and it seems to be generating a version number not considered valid)

augustoproiete:
@epaulsen your changes have been merged, thanks for your contribution :+1:

Issue #3282, opened 2021-03-16
GitVersion Tool: Rename verbosity values to match GitVersion values


augustoproiete:
> Follow-up from #3281

The [`Verbosity` argument of GitVersion](https://github.com/GitTools/GitVersion/blob/5.6.6/src/GitVersion.Core/Logging/Verbosity.cs) changed in more recent releases and the existing values in [Cake's enumeration](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Common/Tools/GitVersion/GitVersionVerbosity.cs) no longer match the accepted values by GitVersion.

This issue is about updating the values in the Cake enumeration to match the exact values values expected in newer versions of GitVersion. This would be a breaking change in the Cake's API and would be included in a future major version of Cake.

| [Cake]  | Rename to     | [GitVersion] | 
| ------- |:-------------:| ------------ |
| `None`  | :arrow_right: | `Quiet`      |
| `Error` | :arrow_right: | `Minimal`    |
| `Warn`  | :arrow_right: | `Normal`     |
| `Info`  | :arrow_right: | `Verbose`    |
| `Debug` | :arrow_right: | `Diagnostic` |

[Cake]: https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Common/Tools/GitVersion/GitVersionVerbosity.cs
[GitVersion]: https://github.com/GitTools/GitVersion/blob/5.6.6/src/GitVersion.Core/Logging/Verbosity.cs

---

Related:

- [Unable to set `none` flag for verbosity](https://github.com/GitTools/GitVersion/issues/1949)
- [Setting default branching strategy fails with an unexpected error](https://github.com/cake-build/cake/discussions/3274#discussioncomment-486088)

vijaya-lakshmi-venkatraman:
Hi, 
I am new to CAKE and would like to know more about the proposed fix. 
Is the change required only on GitVersionVerbosity.cs?

augustoproiete:
Hi @vijaya-lakshmi-venkatraman! Welcome to the Cake project 👋🏻 

Yes, the goal is to rename the enumeration values in [GitVersionVerbosity](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Common/Tools/GitVersion/GitVersionVerbosity.cs) and (if needed) update the `summary` XML comment on the enumeration values to reflect/describe the new name.

[...]

PR #3321, opened 2021-04-24
Renamed GitVersionVerbosity values to match GitVersion

joshuacookdev:
Addresses #3282.

Nothing too major, just a simple enum renaming as requested. I'd like to address [this question](https://github.com/cake-build/cake/issues/3282#issuecomment-826026499) before merging, if possible, but if you guys just want to merge and move on, that's fine with me.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/GitVersion/GitVersionRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/GitVersion/GitVersionRunnerTests.cs
@@ -356,11 +356,11 @@ namespace Cake.Common.Tests.Unit.Tools.GitVersion
             }
 
             [Theory]
-            [InlineData(GitVersionVerbosity.None, nameof(Verbosity.Quiet))]
-            [InlineData(GitVersionVerbosity.Debug, nameof(Verbosity.Diagnostic))]
-            [InlineData(GitVersionVerbosity.Info, nameof(Verbosity.Verbose))]
-            [InlineData(GitVersionVerbosity.Warn, nameof(Verbosity.Normal))]
-            [InlineData(GitVersionVerbosity.Error, nameof(Verbosity.Minimal))]
+            [InlineData(GitVersionVerbosity.Quiet, nameof(Verbosity.Quiet))]
+            [InlineData(GitVersionVerbosity.Diagnostic, nameof(Verbosity.Diagnostic))]
+            [InlineData(GitVersionVerbosity.Verbose, nameof(Verbosity.Verbose))]
+            [InlineData(GitVersionVerbosity.Normal, nameof(Verbosity.Normal))]
+            [InlineData(GitVersionVerbosity.Minimal, nameof(Verbosity.Minimal))]
             public void Should_Add_Verbosity_To_Arguments_If_Set(GitVersionVerbosity verbosity, string arg)
             {
                 // Given
--- a/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
+++ b/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs
@@ -174,23 +174,23 @@ namespace Cake.Common.Tools.GitVersion
             {
                 switch (settings.Verbosity.Value)
                 {
-                    case GitVersionVerbosity.None:
+                    case GitVersionVerbosity.Quiet:
                         builder.Append("-verbosity");
                         builder.Append(nameof(Verbosity.Quiet));
                         break;
-                    case GitVersionVerbosity.Debug:
+                    case GitVersionVerbosity.Diagnostic:
                         builder.Append("-verbosity");
                         builder.Append(nameof(Verbosity.Diagnostic));
                         break;
-                    case GitVersionVerbosity.Info:
+                    case GitVersionVerbosity.Verbose:
                         builder.Append("-verbosity");
                         builder.Append(nameof(Verbosity.Verbose));
                         break;
-                    case GitVersionVerbosity.Warn:
+                    case GitVersionVerbosity.Normal:
                         builder.Append("-verbosity");
                         builder.Append(nameof(Verbosity.Normal));
                         break;
-                    case GitVersionVerbosity.Error:
+                    case GitVersionVerbosity.Minimal:
                         builder.Append("-verbosity");
                         builder.Append(nameof(Verbosity.Minimal));
                         break;
--- a/src/Cake.Common/Tools/GitVersion/GitVersionVerbosity.cs
+++ b/src/Cake.Common/Tools/GitVersion/GitVersionVerbosity.cs
@@ -12,26 +12,26 @@ namespace Cake.Common.Tools.GitVersion
         /// <summary>
         /// No messages will be logged.
         /// </summary>
-        None,
+        Quiet,
 
         /// <summary>
         /// Log error messages.
         /// </summary>
-        Error,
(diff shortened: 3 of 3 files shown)


src/Cake.Common.Tests/Unit/Tools/GitVersion/GitVersionRunnerTests.cs
```diff
@@ -356,11 +356,11 @@ public void Should_Tolerate_Bad_Json_Set()
...
-            [InlineData(GitVersionVerbosity.None, nameof(Verbosity.Quiet))]
-            [InlineData(GitVersionVerbosity.Debug, nameof(Verbosity.Diagnostic))]
-            [InlineData(GitVersionVerbosity.Info, nameof(Verbosity.Verbose))]
-            [InlineData(GitVersionVerbosity.Warn, nameof(Verbosity.Normal))]
-            [InlineData(GitVersionVerbosity.Error, nameof(Verbosity.Minimal))]
+            [InlineData(GitVersionVerbosity.Quiet, nameof(Verbosity.Quiet))]
+            [InlineData(GitVersionVerbosity.Diagnostic, nameof(Verbosity.Diagnostic))]
+            [InlineData(GitVersionVerbosity.Verbose, nameof(Verbosity.Verbose))]
+            [InlineData(GitVersionVerbosity.Normal, nameof(Verbosity.Normal))]
+            [InlineData(GitVersionVerbosity.Minimal, nameof(Verbosity.Minimal))]
```
augustoproiete, line 363: These `InlineData` items should now use the new names int `GitVersionVerbosity`, e.g. 

```suggestion
            [InlineData(GitVersionVerbosity.Minimal, nameof(GitVersionVerbosity.Minimal))]
```

augustoproiete (changes requested):
@joshuacookdev Thanks for PR! One small change needed (see comment) and then it's good to go.

src/Cake.Common.Tests/Unit/Tools/GitVersion/GitVersionRunnerTests.cs
joshuacookdev: Apologies, I figured you still wanted to verify that it mapped to the Cake verbosity level. Updating now.

augustoproiete: No worries! This particular test is checking if we are writing out the correct string when executing the tool. e.g.

```powershell
gitversion.exe -verbosity Minimal
```

So there's no mapping from Cake to GitVersion here. It's all GitVersion. There are other tests in the same file that do perform a mapping, for example if the user does not specify the verbosity they want for GitVersion - then we use the Verbosity of Cake to map to the best option for GitVersion.

joshuacookdev: That makes sense to me - question then in this case:
If you no longer care to have the Cake Verbosity level, it would be better to remove the `nameof(Verbosity.x)` and just do the name of in the assert, no?

joshuacookdev: i.e.
```csharp
[Theory]
[InlineData(GitVersionVerbosity.Quiet)]
[InlineData(GitVersionVerbosity.Diagnostic)]
[InlineData(GitVersionVerbosity.Verbose)]
[InlineData(GitVersionVerbosity.Normal)]
[InlineData(GitVersionVerbosity.Minimal)]
public void Should_Add_Verbosity_To_Arguments_If_Set(GitVersionVerbosity verbosity)
{
    // Given
    var fixture = new GitVersionRunnerFixture();
    fixture.Settings.Verbosity = verbosity;

    // When
    var result = fixture.Run();

    // Then
    Assert.Equal($"-verbosity {nameof(verbosity)}", result.Args);
}
```

augustoproiete: I don't see how you'd do the `nameof` in the assert, as `nameof` is resolved at compile time.

joshuacookdev: You're most definitely right - I was meaning do a ToString() in the assert, just got it confused.

Something like:
```csharp
[Theory]
[InlineData(GitVersionVerbosity.Quiet)]
[InlineData(GitVersionVerbosity.Diagnostic)]
[InlineData(GitVersionVerbosity.Verbose)]
[InlineData(GitVersionVerbosity.Normal)]
[InlineData(GitVersionVerbosity.Minimal)]
public void Should_Add_Verbosity_To_Arguments_If_Set(GitVersionVerbosity verbosity)
{
    // Given
    var fixture = new GitVersionRunnerFixture();
    fixture.Settings.Verbosity = verbosity;

    // When
    var result = fixture.Run();

    // Then
    Assert.Equal($"-verbosity {verbosity.ToString()}", result.Args);
}
```
Compiles, passes, and prevents accidentally comparing GitVersionVerbosity to other Verbosity enums.

joshuacookdev: I pushed a commit with the changes as you requested, just food for thought on testing that the enums are being added correctly to Args. I can adjust if you decide you like the ToString() method.

augustoproiete: Thanks! LGTM as-is.

Nothing particularly _wrong_ with the `ToString` approach, but for the sake of consistency with the other tests, the `nameof` seems slightly better to me given that the tests in this file (and many others in Cake codebase) follow this pattern of `TestName(input, expected)`.

It's also unlikely that this would ever happen, but if GitVersion decides in the future to add an option that uses a special character, or for example two words, say, `-verbosity "info only"`, then the `ToString` wouldn't match given that C# doesn't allow spaces, so the enum would have be `InfoOnly` or `Info_Only`. If we leave the `expected` open for any kind of string input, we could still easily test:

```csharp
// ...
[InlineData(GitVersionVerbosity.Minimal, nameof(GitVersionVerbosity.Minimal))]
[InlineData(GitVersionVerbosity.InfoOnly, "info only")]
```

augustoproiete (changes requested):
LGTM. Could you squash the commits and push --force?

src/Cake.Common.Tests/Unit/Tools/GitVersion/GitVersionRunnerTests.cs
joshuacookdev: Good reasoning. Thanks for the feedback on the work.

Looking forward to helping out more!

joshuacookdev:
@augustoproiete Done - thanks for the feedback.

augustoproiete (approved):
LGTM! Will be merged for Cake v2.0

augustoproiete:
I reached out to @joshuacookdev via DM on Twitter to help with rebasing his PR (as we don't have permission to do ourselves). If we don't hear from him in the next couple of days I'll open a new PR with the same commit so we can merge

gep13:
@augustoproiete @joshuacookdev just having a quick look at this.... do we need to update the descriptions for the enumerated values as well? I don't think they align properly anymore.

arturcic:
@augustoproiete I will check and will let you know

PR #3336, opened 2021-05-06
GH2168: Support for `DirectoryPath` and `FilePath` when reading arguments

### LINKED ISSUE #2168: TypeConverter to enable Argument<DirectoryPath>(...)
This code works so long as the argument `--artifact-dir` is not present:

```cake
var artifactDirectory = Argument<DirectoryPath>("artifact-dir", "Artifacts");
```

If it is present, Cake throws A System.Exception:

> TypeConverter cannot convert from System.String.

This would be easy to fix by adding a TypeConverter!

- Could we also do this for `FilePath`?
- Any other Cake types we could convert from string?
- I thought there was an issue around here somewhere to add a generic `EnvironmentVariable<DirectoryPath>(...)` method. Should I start one, since https://github.com/cake-build/cake/issues/1878 is specifically about enums?

augustoproiete:
Add support for using `DirectoryPath` and `FilePath` types directly when reading arguments passed to the Cake build.

eg.

```csharp
var outputFolder = Argument<DirectoryPath>("outputFolder");
var logFile = Argument<FilePath>("logFile");
```

---

Closes #2168

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -669,6 +669,8 @@ Task("Run-Integration-Tests")
                     .AppendSwitchQuoted("--customarg", " ", "hello")
                     .AppendSwitchQuoted("--multipleargs", "=", "a")
                     .AppendSwitchQuoted("--multipleargs", "=", "b")
+                    .AppendSwitchQuoted("--testAssemblyDirectoryPath", "=", cakeAssembly.GetDirectory().FullPath)
+                    .AppendSwitchQuoted("--testAssemblyFilePath", "=", cakeAssembly.FullPath)
             });
     }
     catch(Exception ex)
--- a/src/Cake.Common.Tests/Unit/ArgumentAliasesTests.cs
+++ b/src/Cake.Common.Tests/Unit/ArgumentAliasesTests.cs
@@ -1,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Core;
+using Cake.Core.IO;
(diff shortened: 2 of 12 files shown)


devlead (changes requested):
This LGTM 👍 

But I would like some integration tests to test these new features.

Probably something like below would be bare minimum

### /build.cake

```diff
                    .AppendSwitchQuoted("--multipleargs", "=", "b")
+                    .AppendSwitchQuoted("--testAssemblyFilePath", "=", cakeAssembly.FullPath)
+                    .AppendSwitchQuoted("--testAssemblyDirectoryPath", "=", cakeAssembly.GetDirectory().FullPath)
            });
```

### /tests/integration/Cake.Common/ArgumentAliases.cake

```cake
Task("Cake.Common.ArgumentAliases.Argument.FilePathArgument")
    .Does(() =>
{

    // Given
    var testAssemblyPath = Context
                            .Environment
                            .ApplicationRoot
                            .CombineWithFilePath(
#if NETCOREAPP
                                "Cake.dll"
#else
                                "Cake.exe"
#endif
                            );

    // When
    var arg = Argument<FilePath>("testAssemblyFilePath");

    // Then
    Assert.Equal(testAssemblyPath.FullPath, arg.FullPath);
});


Task("Cake.Common.ArgumentAliases.Argument.DirectoryPathArgument")
    .Does(() =>
{

     // Given, When
    var arg = Argument<DirectoryPath>("testAssemblyDirectoryPath");

    // Then
    Assert.Equal(Context.Environment.ApplicationRoot.FullPath, arg.FullPath);
});
```

devlead (approved):
LGTM 👍

devlead:

PR #3339, opened 2021-05-07
GH2654: Add Support for TestParams flag in NUnit3

### LINKED ISSUE #2654: NUnit3Settings should support TestParam
### What You Are Seeing?
I see a warning `--params is deprecated and will be removed in a future release. Please use --testparam instead.' when using NUnit.ConsoleRunner 3.10.0 and using the `Params` property of the NUnit3Settings class.

### What is Expected?
Either redirect the values from `Params` to the `--testparam` parameter or provide a dedicated `TestParams` property.

0.33

64bit

Windows

No

Execute Nunit3 unit tests with parameters

### Output Log
```
Executing: C:\xyz\tools\NUnit.ConsoleRunner.3.10.0\tools\nunit3-console.exe
--params is deprecated and will be removed in a future release. Please use --testparam instead.
```

joshuacookdev:
Resolves #2654.

I wasn't sure if it was better to add TestParams as its own property, or to convert the Params to the new `--testparam:key=value` syntax in NUnit v3.10+. I opted for creating a new property so that we have support for both, but am willing to change to converting the output if that's deemed better. 

As it stands, the only issue that I can see with this is that we might make it confusing to allow both, but this was allows us to support multiple versions of the runner, and lets Cake scripts migrate as they're ready (for now) without any breaking changes.

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/NUnit/NUnit3RunnerTests.cs
@@ -244,6 +244,12 @@ namespace Cake.Common.Tests.Unit.Tools.NUnit
                     ["two"] = "2",
                     ["three"] = "3"
                 };
+                fixture.Settings.TestParams = new Dictionary<string, string>
+                {
+                    ["uno"] = "1",
+                    ["dos"] = "2",
+                    ["tres"] = "3"
+                };
 
                 // When
                 var result = fixture.Run();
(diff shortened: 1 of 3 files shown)


joshuacookdev:
Sorry, had to fix a typo.

joshuacookdev:
Didn't realize that including an obsolete tag would fail all of those builds...

devlead (changes requested):
@joshuacookdev could you please rebase and squash commits against latest develop branch.

devlead (approved):
LGTM 👍

devlead:
@joshuacookdev your changes have been merged, thanks for your contribution 👍

PR #2650, opened 2019-10-27
GH2099: Cache compiled script on disk

tstewart65:
This pull request is what started RFC-02.  At the time I was not able to submit a pull request for the changes.

**Differences from RFC-02**

1. The cache argument was removed.  A new section was added to the cake.config file [Cache] with 2 possible properties Enabled and Path.  Enabled defaults to false and Path defaults cache under the tools folder.
2. I originally used a time stamp to determine if the cache was valid or not.  This worked for the way I was using Cake but would not work for all cases.  I have changed this to use a hash like pull request 2584 used.  I did change it to use SHA1CryptoServiceProvider instead of MD5 since MD5 is not FIPS compliant.

I have been using several versions of Cake with this  applied and have had no issues.

@mholo65 posted this in the RFC Comments
"Caching compilation is not alone enough. We must also cache registered tools resolved via #tool directive. Addins (and their dependencies) must also be loaded into app domain / load context in same order as the first time when resolved via #addin directive."

For the way this is implemented I feel just caching the compilation is enough.  I need all of the lines of the script in order to do the hash for determining if anything has changed.  The tools and addins are still being found like usual.  The main changes were made in RoslynScriptSession and in there it just determines if it needs to compile the script or run the cached DLL.

PR #3381, opened 2021-05-29
GH2099: Also takes care of RFC-02.

tstewart65:
This is the same as pull request 2650 but has been updated to match the changes made in Cake version 1.0.0.

I am hoping that since version 1.0.0 has been released this can be merged.

### DIFF
--- a/src/Cake/Commands/DefaultCommandSettings.cs
+++ b/src/Cake/Commands/DefaultCommandSettings.cs
@@ -59,5 +59,9 @@ namespace Cake.Commands
         [CommandOption("--info")]
         [Description("Displays additional information about Cake.")]
         public bool ShowInfo { get; set; }
+
+        [CommandOption("--regeneratecache")]
+        [Description("Forces the script cache to be regenerated if caching is enabled.")]
+        public bool Recompile { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Cake/Commands/DefaultCommandSettings.cs
```diff
@@ -59,5 +59,9 @@ public sealed class DefaultCommandSettings : CommandSettings
         [CommandOption("--info")]
         [Description("Displays additional information about Cake.")]
         public bool ShowInfo { get; set; }
+
+        [CommandOption("--regeneratecache")]
```
gitfool, line 63: Use `--recompile` instead or as an alias for consistency?

Also, `--regenerate-cache` would be more in line with cli conventions; i.e. separate words with dashes.

tstewart65: I had --recompile in a previous pull request and they wanted that changed at that time.  That is when I made it --regeneratecache.  If we get a consensus on what is desired I will change it.

bjorkstromm: I'd say `--invalidate-cache`.

pascalberger:
Closed in favor of #2650. #2650 has been updated with code from this PR.

PR #3388, opened 2021-06-06
Added Arguments() method to retrieve command line arguments as IDictionary

### LINKED ISSUE #2690: Consider adding some kind of "GetArguments()" alias, similar to the EnvironmentVariables() one
Right now, there is a way to get all of the Environment Variables using the [EnvironmentVariables](https://cakebuild.net/api/Cake.Common/EnvironmentAliases/97336C76) alias.  It returns an IDictionary<string, string>.

However, there does not seem to be one to get all of the arguments (unless I am missing something, in which case I apologize).

If the alias is to be consistent with EnvironmentVariables() it should probably return a copy of a dictionary, not a direct reference to the underlying dictionary Cake uses.

**Example Use Case:**
The use case for this could be if one wants to call [CakeExecuteScript](https://cakebuild.net/api/Cake.Common.Tools.Cake/CakeAliases/5FCD85B6) and pass-through the current process's arguments to it (or append arguments of its own), there's no way to do this right now.

```C#
// Arguments() does not exist right now.
// To be consistent with EnvironmentVariables() it should probably a new dictionary
// that is safe to be modified, not a reference to Cake's underlying one.
IDictionary<string, string> currentArgs = Arguments();
currentArgs["additionalArg"] = "myOtherArg";

CakeSettings settings = new CakeSettings
{
    Arguments = currentArgs,
    EnvironmentVariables = EnvironmentVariables()
};

CakeExecuteScript( "somewhereElse/build.cake", settings );
```

sunshykin:
Fixes #2690
Tried my first issue, hope I did it ok.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3388) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3388) before we can accept your contribution.<br/><br/>:x: sunshykin [sign now](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3388)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/cake-build/cake?pullRequest=3388) it.</sub>

augustoproiete:
Thanks a lot for sending the PR @sunshykin and apologies for not being able to provide feedback sooner.

In principle this PR looks good to me, and just need a couple of adjustments:

- Cake already has plumbing for parsing arguments and making them available to builds via `Context.Arguments`, which exposes [`ICakeArguments`](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Core/ICakeArguments.cs#L12), implemented by [CakeArguments](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Core/CakeArguments.cs#L10), so this is a more fitting place for implementing this new API

- Newer versions of Cake (released after [#2690](https://github.com/cake-build/cake/issues/2690) was created) added support for multiple values with the same name/key, so it's possible for an argument to have more than one value, thus instead of returning `IDictionary<string, string>` we should return `IDictionary<string, ICollection<string>>`

- Cake uses [Spectre.Console](https://spectreconsole.net) to parse the arguments, and we can leverage this directly in [CakeArguments](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Core/CakeArguments.cs#L10) without having to parse it ourselves... So your implementation could look something as simple as the example below:

```csharp
public IDictionary<string, ICollection<string>> GetArguments()
{
    var arguments = _arguments
        .ToDictionary(x => x.Key, x => (ICollection<string>)x.Value.ToList());

    return arguments;
}
```

src/Cake.Core/CakeEnvironment.cs
```diff
@@ -81,6 +82,12 @@ public string GetEnvironmentVariable(string variable)
                     dictionary => dictionary);
         }
 
+        /// <inheritdoc/>
+        public IDictionary<string, string> GetArguments()
```
augustoproiete, line 86: - [ ] Move implementation to [CakeArguments](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Core/CakeArguments.cs#L10)
- [ ] Return type should be `IDictionary<string, ICollection<string>>` to allow for multiple values for the same argument

src/Cake.Core/CommandLineArguments/CommandLineArgumentsParser.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Cake.Core.CommandLineArguments
+{
+    internal static class CommandLineArgumentsParser
+    {
+        public static IDictionary<string, string> Parse(string[] arguments)
```
augustoproiete, line 8: Let's leverage the parsing of the arguments that is already done for us via Spectre.Console, which is available inside of [CakeArguments](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Core/CakeArguments.cs#L10), so we don't need to implement additional logic to parse the arguments

src/Cake.Core/ICakeEnvironment.cs
```diff
@@ -43,9 +43,15 @@ public interface ICakeEnvironment
...
         /// </summary>
-        /// <returns>The environment variables as IDictionary&lt;string, string&gt;. </returns>
+        /// <returns>The environment variables as IDictionary&lt;string, string&gt;.</returns>
         IDictionary<string, string> GetEnvironmentVariables();
 
+        /// <summary>
+        /// Gets all command line arguments.
+        /// </summary>
+        /// <returns>The command line arguments as IDictionary&lt;string, string&gt;.</returns>
+        IDictionary<string, string> GetArguments();
```
augustoproiete, line 53: - [ ] Move API contract to [ICakeArguments](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Core/ICakeArguments.cs#L12)
- [ ] Return type should be `IDictionary<string, ICollection<string>>` to allow for multiple values for the same argument

src/Cake.Testing/FakeEnvironment.cs
```diff
@@ -109,6 +111,12 @@ public string GetEnvironmentVariable(string variable)
             return new Dictionary<string, string>(_environmentVariables, StringComparer.OrdinalIgnoreCase);
         }
 
+        /// <inheritdoc/>
+        public IDictionary<string, string> GetArguments()
```
augustoproiete, line 115: This will be available in the fixtures (might need tweaking a fixture or create a new one)

augustoproiete (changes requested):

src/Cake.Testing/FakeEnvironment.cs
sunshykin: Checked, that `CakeContextFixture` has `Arguments = Substitute.For<ICakeArguments>();` line. Do I need to add a new Fixture or some tests with that?

src/Cake.Core/ICakeEnvironment.cs
sunshykin: Done

src/Cake.Core/CommandLineArguments/CommandLineArgumentsParser.cs
sunshykin: Done

src/Cake.Core/CakeEnvironment.cs
sunshykin: Done

sunshykin:
@augustoproiete Done with comments, please take a look.

augustoproiete:
Thanks @sunshykin! We'll review over the next few days

augustoproiete (approved):
LGTM!

augustoproiete:
@sunshykin your changes have been merged, thanks for your contribution :+1:

PR #3400, opened 2021-06-27
#1663 - Fix CopyFiles when empty enumeration is passed

### LINKED ISSUE #1663: CopyFiles alias throws exception if empty enumeration is passed
### What You Are Seeing?
The [CopyFiles](http://cakebuild.net/api/Cake.Common.IO/FileAliases/8B962704) alias currently throws an "Sequence contains no matching element" exception if an empty enumeration is passed to copy.

### What is Expected?
Either the alias should not throw (since there is nothing to copy) or an `ArgumentException` should be thrown.

0.20.0

64-Bit

Windows

coder2213:
Fix #1663 

Added if condition to filter out the empty enumerations passed.
Not raising any errors or exceptions as confirmed on the issue discussion.

### DIFF
--- a/src/Cake.Common/IO/FileCopier.cs
+++ b/src/Cake.Common/IO/FileCopier.cs
@@ -106,42 +106,45 @@ namespace Cake.Common.IO
                 throw new DirectoryNotFoundException(message);
             }
 
-            if (preserverFolderStructure)
+            if (filePaths.Any())
             {
-                var commonPath = string.Empty;
-                var separatedPath = filePaths
-                    .First(str => str.ToString().Length == filePaths.Max(st2 => st2.ToString().Length)).ToString()
-                    .Split(new string[] { "/" }, StringSplitOptions.RemoveEmptyEntries)
-                    .ToList();
-
-                foreach (string pathSegment in separatedPath)
+                if (preserverFolderStructure)
                 {
-                    if (commonPath.Length == 0 && filePaths.All(str => str.ToString().StartsWith(pathSegment)))
+                    var commonPath = string.Empty;
+                    var separatedPath = filePaths
+                        .First(str => str.ToString().Length == filePaths.Max(st2 => st2.ToString().Length)).ToString()
+                        .Split(new string[] { "/" }, StringSplitOptions.RemoveEmptyEntries)
+                        .ToList();
+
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3400) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3400) before we can accept your contribution.<br/><br/>:x: coder2213 [sign now](https://cla.dotnetfoundation.org/cake-build/cake?pullRequest=3400)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/cake-build/cake?pullRequest=3400) it.</sub>

coder2213:
Failure on mac seems to be known issue. #2564
I think this is not caused due to my code changes.

src/Cake.Common/IO/FileCopier.cs
```diff
@@ -106,42 +106,45 @@ public static void CopyFiles(ICakeContext context, IEnumerable<FilePath> filePat
                 throw new DirectoryNotFoundException(message);
             }
 
-            if (preserverFolderStructure)
+            if (filePaths.Any())
```
augustoproiete, line 109: We only need to check if the the list is empty when `preserveFolderStructure` is `false`, so let's move this inside the `preserveFolderStructure` condition to avoid an unnecessary enumeration. Let's also invert the new `if` statement to reduce nesting (i.e. if list is empty, give up and return)

augustoproiete (changes requested):
Thanks @coder2213 Just a small change, and this should be good to go

src/Cake.Common/IO/FileCopier.cs
coder2213: @augustoproiete 
This is done.
Please check.

augustoproiete: Thanks, that's it. Cheers. Will merge after CI finishes

augustoproiete:
@coder2213 your changes have been merged, thanks for your contribution :+1:

Issue #3237, opened 2021-03-06
Allow setting MSBuildToolVersion using custom string - Part 1


augustoproiete:
The `MSBuildSettings` extension [`UseToolVersion`](https://github.com/cake-build/cake/blob/v1.0.0/src/Cake.Common/Tools/MSBuild/MSBuildSettingsExtensions.cs#L33-L47) expects a valid value from the [MSBuildToolVersion](https://cakebuild.net/api/Cake.Common.Tools.MSBuild/MSBuildToolVersion/) enumeration.

```csharp
public static MSBuildSettings UseToolVersion(this MSBuildSettings settings, MSBuildToolVersion version)
{
    // ...
}
```

Add an overload which accepts a custom string, to make it easier for using newer versions of MSBuild before a new version of Cake is released, including an updated version of the enumeration.

```csharp
public static MSBuildSettings UseToolVersion(this MSBuildSettings settings, string version)
{
    // ...
}
```

Marusyk:
Could you please clarify how custom string can be interpreted to `MSBuildToolVersion`?
```
public MSBuildToolVersion ToolVersion { get; set; }
```
and how base on a custom sring we can return MSBuild path?

augustoproiete:
Hey @Marusyk, thanks for looking into this one.

Good question. When I created this issue, I was thinking of tackling this in two parts:

1. :green_circle: (This issue) Add an overload of `UseToolVersion` that can accept a string and stop there - i.e. Leaving the `ToolVersion` property using the enum as it is today, and **release in Cake v1.x** (as it's brand new extension method, so non-breaking)

2. :orange_circle: (Your question, Future issue) Change the `ToolVersion` to use a "smart" enum that can be converted to/from `string`, and **release in Cake v2.x** (as it's changing the public API, so it's breaking)

What do you think?

To do item 1 :green_circle:, we could add an internal `string` backing field to `MSBuildSettings`, change the implementation of the `ToolVersion` to read/write from this field converting the values to/from the enum using the [existing translations](https://github.com/cake-build/cake/blob/8bc15841e97944291946a89f335562244965247e/src/Cake.Common/Tools/DotNetCore/MSBuild/MSBuildArgumentBuilderExtensions.cs#L233-L254) (but keeping the enum in the public API), and then implement the `UseToolVersion` overload that simply sets the value to the internal string backing field to any value the user wants.

[...]

PR #3435, opened 2021-08-27
setup useToolVersion override

narcopanda:
Added override for UseToolVersion method to take in a string param also added a helper method to map said string to the correct MSBuildToolVersion enum. It's a little big since I wasn't sure on what type of string we should expect. 
https://github.com/cake-build/cake/issues/3237

This is my first time contributing to a open source project so apologies for any mistakes with the process

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsExtensionsTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/MSBuild/MSBuildSettingsExtensionsTests.cs
@@ -67,6 +67,55 @@ namespace Cake.Common.Tests.Unit.Tools.MSBuild
                 // Then
                 Assert.Equal(settings, result);
             }
+
+            [Fact]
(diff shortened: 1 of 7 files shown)


augustoproiete:
Want to give that a try? We already have a [UseToolsVersion](https://github.com/cake-build/cake/blob/v1.1.0/src/Cake.Common/Tools/MSBuild/MSBuildSettingsExtensions.cs#L39) that accepts an enum, so the main value proposition of this new overload would be to allow users to use an arbitrary string that Cake doesn't yet know about - assuming that can be used to locate the right MSBuild.

narcopanda:
sure

narcopanda:
I added to 2 new enums to MSBuildToolVersion one that is for a custom .net framework and another for a custom visual studio.
I also updated the MSBuildVersion enums to contain the new enums so that we can match it

devlead (approved):
LGTM 👍

devlead:
@narcopanda your changes have been merged, thanks for your contribution 👍

Issue #2104, opened 2018-03-29
Make possibility to set Process Exit Code


Gargony:
Currently cake process can return exit code 0 or 1 only.
Will useful if I can return custom Process Exit Code from task.
e.g. `throw CakeExitCodeException(-1)`

devlead:
Yip that could be useful, as a workaround you could currently use `System.Environment.Exit(-1);`

![image](https://user-images.githubusercontent.com/1647294/38080740-8ff4f1f2-3342-11e8-8787-19614957b23b.png)

patriksvensson:
This doesn't have to be a new exception btw. We could add a constructor overload to the CakeException to contain an exit code, since it's always used to indicate a problem within Cake (that will lead to an exit if not handled)

devlead:
@patriksvensson exactly just an overload constructor with 1 as default.

PR #3446, opened 2021-08-30
GH2104: Added custom exit code to CakeException

narcopanda:
added overload constructor that sets the exit code of the app and has the default code set to 1
I wasn't sure how to make a unit test(s) for this since setting the exit code kills the app

https://github.com/cake-build/cake/issues/2104

### DIFF
--- a/src/Cake.Core/CakeException.cs
+++ b/src/Cake.Core/CakeException.cs
@@ -18,6 +18,11 @@ namespace Cake.Core
     public sealed class CakeException : Exception
     {
         /// <summary>
+        /// Gets custom exit code.
+        /// </summary>
+        public int ExitCode { get; } = 1;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="CakeException"/> class.
         /// </summary>
         public CakeException()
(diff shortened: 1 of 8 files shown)


src/Cake.Core/CakeException.cs
```diff
@@ -43,6 +43,15 @@ public CakeException(string message, Exception innerException)
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CakeException"/> class.
+        /// </summary>
+        /// <param name="exitCode">Exit code</param>
+        public CakeException(int exitCode = 1)
+        {
+            System.Environment.Exit(exitCode);
```
devlead, line 52: We don't want to exit from the constructor, as this will kill the build regardless of what error handling exists in the script.

What we want is additional constructors overloads with a exitCode parameter, having it stored in a read-only `int ExitCode` property where its default value is `1`.
Something like `public int ExitCode { get; } = 1;`

And constructors like
```` csharp
        public CakeException(int exitCode)
        {
            ExitCode = exitCode;
        }

        public CakeException(int exitCode, string message)
            : base(message)
        {
            ExitCode = exitCode;
        }

        public CakeException(int exitCode, string message, Exception innerException)
            : base(message, innerException)
        {
            ExitCode = exitCode;
        }
````
We want new overloads and not default parameters as we don't want to break ABI with existing code.

devlead (changes requested):

src/Cake.Core/CakeException.cs
narcopanda: I just updated the pr based on your comments. Sorry it took so long and sorry for the confusion

devlead (approved):
LGTM,  I rebased, and added an integration test.

devlead:
@narcopanda your changes have been merged, thanks for your contribution 👍

PR #3457, opened 2021-09-02
(#3452) added an option to append --build / --no-build to InspectCodeSettings

### LINKED ISSUE #3452: Missing option in InspectCodeSettings: `--build` and `--no-build` flags
### Prerequisites

### Cake runner

Cake .NET Tool, Cake Frosting, Cake runner for .NET Framework, Cake runner for .NET Core

### Cake version

1.2.0

### Operating system

N/A

N/A

### CI Server

_No response_

### What are you seeing?

Running `InspectCode` with `JetBrains.ReSharper.CommandLineTools` version 2021.2.0 or later produces the following warning:

```
Warning: Starting from version 2021.2, InspectCode builds the target solution before starting the analysis to make sure it only finds relevant code issues.
To explicitly accept the new behavior and suppress this warning, use the '--build' option.
To match the behavior in previous versions and skip the build, use '--no-build'.
```

However, currently `InspectCodeSettings` does not have a property to pass those flags.

### What is expected?

I propose adding a new property to `InspectCodeSettings`: `BuildBehaviour` which is an enum containing of three values: `Undefined`, `Build` and `NoBuild`.

The default should be `Undefined` which will result in no change to the generated command arguments and hence be backwards compatible with existing usages.
Setting `BuildBehaviour` to `Build` should add the `--build` flag to the generated command arguments, and setting  `BuildBehaviour` to `NoBuild` should add the `--no-build` flag to the generated command arguments.

### Steps to Reproduce

Select any build that uses `InspectCode` and bump  `JetBrains.ReSharper.CommandLineTools` to version 2021.2.0 or later.

### Output log

_No response_

nils-a:
fixes #3452

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/InspectCode/InspectCodeRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/InspectCode/InspectCodeRunnerTests.cs
@@ -408,6 +408,48 @@ namespace Cake.Common.Tests.Unit.Tools.InspectCode
                 // Then
                 Assert.Equal("\"/config=/Working/config.xml\"", result.Args);
             }
+
+            [Fact]
+            public void Should_Not_Contain_Build_Or_NoBuild_If_Build_Is_Not_Set()
+            {
+                // Given
+                var fixture = new InspectCodeRunFixture();
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.DoesNotContain("--build", result.Args);
+                Assert.DoesNotContain("--no-build", result.Args);
+            }
+
+            [Fact]
+            public void Should_Contain_Build_If_Build_Is_Set_To_True()
+            {
+                // Given
+                var fixture = new InspectCodeRunFixture();
+                fixture.Settings.Build = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Contains("--build", result.Args);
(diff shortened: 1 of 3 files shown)


src/Cake.Common.Tests/Unit/Tools/InspectCode/InspectCodeRunnerTests.cs
```diff
@@ -408,6 +408,48 @@ public void Should_Use_Provided_Config_File()
                 // Then
                 Assert.Equal("\"/config=/Working/config.xml\"", result.Args);
             }
+
+            [Fact]
+            public void Should_Contain_build_or_noBuild_if_Build_is_not_set()
```
augustoproiete, line 413:
```suggestion
            public void Should_Not_Contain_build_nor_noBuild_if_Build_is_not_set()
```

augustoproiete (changes requested):
LGTM. Small nitpick on the of one of the tests to reflect what it's testing

src/Cake.Common.Tests/Unit/Tools/InspectCode/InspectCodeRunnerTests.cs
nils-a: Good catch. Fixed.

augustoproiete (approved):
LGTM

PR #3468, opened 2021-09-20
GH3467: Update .NET SDK to 5.0.401

### LINKED ISSUE #3467: Update .NET SDK to 5.0.401

devlead:
* fixes #3467


.github/workflows/build.yml
```diff
@@ -21,20 +21,20 @@ jobs:
...
         with:
-          dotnet-version: '3.1.409'
+          dotnet-version: '3.1.x'
 
-      - name: Install .NET Core SDK 5.0.203
+      - name: Install .NET Core SDK 5.0.x
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '5.0.203'
+          dotnet-version: '5.0.x'
```
augustoproiete, line 37: Are we sure we want to use floating versions of the .NET SDK (even if patch)? If we do this we can no longer say we have deterministic builds. :thinking:

My initial reaction is that we should keep a hard-coded version at all times and keep manually updating (hopefully dependabot or similar will be able to send PRs in the future to make this easier)

devlead: version from global json is installed afterwards.

augustoproiete: :thinking: Doesn't that mean we are always running on the latest feature installed (due to rollForward policy)?

devlead: Not when we change it to 6.0 ;) But perhaps the 5.x should be removed altogether until we've got 6.x in global json. Think it was added because of some issues with CI.

augustoproiete:
@devlead your changes have been merged, thanks for your contribution :+1:

Issue #2048, opened 2018-02-20
DotNetCoreToolSettings.WorkingDirectory is not respected when running DotNetCoreTool


derwasp:
### What You Are Seeing?
Working directory is taken from [here](https://github.com/cake-build/cake/blob/edd75e172fccd18de5801b82446c103e040d4ff8/src/Cake.Common/Tools/DotNetCore/Tool/DotNetCoreToolRunner.cs#L60), while it seems like `DotNetCoreToolSettings` should override it.

### What is Expected?
`DotNetCoreToolSettings.WorkingDirectory is respected`

### What version of Cake are you using?
0.25.0

### Are you running on a 32 or 64 bit system?
64
### What environment are you running on?  Windows? Linux? Mac?
Windows

### How Did You Get This To Happen? (Steps to Reproduce)
```
void PackageLambda(FilePath project, string outputFile)
{
    var lambdaCmd = string.Format("package --output-package {0} --configuration Release --framework netcoreapp1.0", outputFile);
    var settings
        = new DotNetCoreToolSettings{
            WorkingDirectory = "my @@ random ## invalid //// directory",
            DiagnosticOutput = true
        };
    DotNetCoreTool(project, "lambda", lambdaCmd, settings);
}
```
The code above should fail, because the working dir is invalid, but it works just fine.

GeertvanHorrik:
Running into this as well with 0.30.0, but just checked the source code and it's still the same on the latest main branch.

GeertvanHorrik:
A workaround for now is to pass in `null` for the first argument, like this:

```
var arguments = new List<string>();

arguments.Add("--dry-run");

if (failOnChanges)
{
    arguments.Add("--check");
}

DotNetCoreTool(null, "format", string.Join(" ", arguments),
    new DotNetCoreToolSettings
    {
        WorkingDirectory = "./src/"
    });
```

kashifsoofi:
Can I please work on this?

PR #3500, opened 2021-10-04
Respect WorkingDirectory if set

kashifsoofi:


devlead (changes requested):
I suspect this relates to #2048, but don't think this sorts the issue.

It is about using the `DotNetCoreToolSettings settings.WorkingDirectory` if set.

src/Cake.Common/Tools/DotNetCore/Tool/DotNetCoreToolRunner.cs
```diff
@@ -53,7 +53,8 @@ public void Execute(FilePath projectPath, string command, ProcessArgumentBuilder
 
             var processSettings = new ProcessSettings();
 
-            if (projectPath != null)
+            if (processSettings.WorkingDirectory == null &&
+                projectPath != null)
             {
                 processSettings.WorkingDirectory = projectPath.GetDirectory();
             }
```
devlead, line 60: Probably something more like this would be more in line with solving the issue.
```suggestion
            var processSettings = new ProcessSettings
            {
                WorkingDirectory = settings.WorkingDirectory ?? projectPath?.GetDirectory()
            };
```

devlead (changes requested):

src/Cake.Common/Tools/DotNetCore/Tool/DotNetCoreToolRunner.cs
kashifsoofi: Aha, yes that makes more sense

augustoproiete (approved):
LGTM!

augustoproiete:
@kashifsoofi your changes have been merged, thanks for your contribution :+1:

PR #3504, opened 2021-10-04
GH-3501: Add sources attribute to DotNetCorePackSettings

### LINKED ISSUE #3501: Add NuGet Sources argument to DotNetCorePackSettings
Similar to https://github.com/cake-build/cake/issues/2165 and https://github.com/cake-build/cake/issues/2119, the [`DotNetCorePackSettings`](https://github.com/cake-build/cake/blob/v1.2.0/src/Cake.Common/Tools/DotNetCore/Pack/DotNetCorePackSettings.cs) currently does not expose the `Sources` property, which is used during implicit restores as per the [docs](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack#implicit-restore):

> _You don't have to run `dotnet restore` because it's run implicitly by all commands that require a restore to occur, such as `dotnet new`, `dotnet build`, `dotnet run`, `dotnet test`, `dotnet publish`, and `dotnet pack`. To disable implicit restore, use the `--no-restore` option_.
> 
> _(...)_
> 
> _This command supports the `dotnet restore` options when passed in the long form (for example, `--source`). Short form options, such as `-s`, are not supported_.

eoehen:
Add sources attribute to DotNetCorePackSettings.

Close #3501


augustoproiete (approved):
LGTM!

augustoproiete:
@eoehen your changes have been merged, thanks for your contribution :+1:

PR #3516, opened 2021-10-05
GH3515: added NuGet.org.md as a "readme" for NuGet.org

nils-a:
and added it to all packages, except Cake.Cake.DotNetTool.Module.

This first version is a copy of the current README.md with only
non-working images removed, as they won't be rendered on
NuGet.org. (Namely this was the build-status and coverage sections.)


NuGet.org.md
```diff
@@ -0,0 +1,74 @@
+﻿# Cake
```
devlead, line 1: To no not add more noise in the root folder, I think we should place this one in the `nuspec` folder where icon etc. exists.

devlead (changes requested):

NuGet.org.md
nils-a: done.

devlead (approved):
LGTM 👍

devlead:
@nils-a your changes have been merged, thanks for your contribution 👍

PR #3557, opened 2021-10-07
GH-2075: Modify alias DotNetCoreNuGetPush to allow using string or FilePath.

### LINKED ISSUE #2075: Add overloads for DotNetCore*() methods taking FilePath instead of string
### What You Are Seeing?

I recently got a compilation exception because I was trying to pass a `FilePath` that I'd found through `GetFiles().Single()` into `DotNetCoreNuGetPush()`.

### What is Expected?

The argument passed to `DotNetCoreNuGetPush()` is a file path, but is of type `string`. It seems off that I need to do `myFilePath.ToString()` to pass it in. I would expect this method to take a `FilePath` type.

Of course, backwards compatibility is a concern - but could `FilePath` overload be added? The [DotNetCore*()](https://cakebuild.net/dsl/dotnetcore/) methods in general seem to vary between taking a FilePath or a string - I can't work out any pattern here?

This may have initially be confused by the inaccurate docs, see: https://github.com/dotnet/docs/issues/4585

0.26.0

64 bit

Windows
Nope

```
var filePath = GetFiles("*.nupkg").Single();
DotNetCoreNuGetPush(filePath);
```

dharillo:
resolves #2075 

The change will allow using FilePath or string as parameter for this alias. 
The implicit operator of the FilePath class will convert the string value into a FilePath automatically.

### DIFF
--- a/src/Cake.Common/Tools/DotNetCore/DotNetCoreAliases.cs
+++ b/src/Cake.Common/Tools/DotNetCore/DotNetCoreAliases.cs
@@ -867,25 +867,29 @@ namespace Cake.Common.Tools.DotNetCore
         /// Pushes one or more packages to a server.
         /// </summary>
         /// <param name="context">The context.</param>
-        /// <param name="packageName">Name of package to push.</param>
+        /// <param name="packageFilePath"><see cref="FilePath"/> of the package to push.</param>
         /// <example>
         /// <code>
-        /// DotNetCoreNuGetPush("*.nupkg");
+        /// // With FilePath instance
+        /// var packageFilePath = GetFiles("*.nupkg").Single();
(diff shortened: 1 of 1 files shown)


dharillo:
I seem to be making a mess with the CI builds, sorry. The first time that I opened the pull request, the appveyor workflow failed because the version extracted from the branch name was an invalid version name. I renamed the branch not knowing that the original pull request will be closed (sorry).

Now, TeamCity is failing because it is finding two pull requests with the same commit id. I think that it could be solved if I made another commit to this branch, but I will wait until receiving some feedback from you guys.

Sorry again

dharillo:
Thank you! @augustoproiete

Issue #3569, opened 2021-10-09
Set Cake runner project as the default Startup Project


augustoproiete:
![image](https://user-images.githubusercontent.com/177608/136666358-f32407bb-1322-4d60-8ebe-43e95b5dc928.png)

When opening the Cake solution on Visual Studio, the `Cake.Core` project is set as the Startup Project, and upon trying to run or debug Cake, we (understandably) get an error

```
A project with an Output Type of Class Library cannot be started directly

In order to debug this project, add an executable project to this solution which references the library project. Set the executable project as the startup project.
```

![image](https://user-images.githubusercontent.com/177608/136666479-0075cd1e-529f-46c2-8089-5d62e1574f63.png)

PR #3570, opened 2021-10-09
(#3569) Set Cake runner project as the default Startup Project

augustoproiete:
Closes https://github.com/cake-build/cake/issues/3569


gep13:
@augustoproiete thanks for getting this updated!

PR #3675, opened 2021-11-06
GH1111: Update DotNetRestore alias XML Docs to match official docs

### LINKED ISSUE #1111: DotNetCoreRestore: dotnet restore no longer supports globbing

augustoproiete:
Matches [official docs on `dotnet restore`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#arguments) as of this writing:

image

---

Closes #1111

### DIFF
--- a/src/Cake.Common/Tools/DotNet/DotNetAliases.cs
+++ b/src/Cake.Common/Tools/DotNet/DotNetAliases.cs
@@ -151,10 +151,10 @@ namespace Cake.Common.Tools.DotNet
         /// Restore all NuGet Packages in the specified path.
         /// </summary>
         /// <param name="context">The context.</param>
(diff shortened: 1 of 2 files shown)


devlead (approved):
LGTM 👍

Issue #3697, opened 2021-11-14
Error: The requested service 'Cake.Commands.DefaultCommandSettings' has not been registered


augustoproiete:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake .NET Tool

### Cake version

2.0.0-rc0001

### Operating system

N/A

### Operating system architecture

N/A

### CI Server

N/A

### What are you seeing?

When attempting to F5/Debug the `Cake` project on Visual Studio 2022 (also on 2019), we get the error below:

```

Autofac.Core.Registration.ComponentNotRegisteredException
  HResult=0x80131500
  Message=The requested service 'Cake.Commands.DefaultCommandSettings' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method to resolve an optional dependency.
  Source=Autofac
  StackTrace:
   at Autofac.ResolutionExtensions.ResolveService(IComponentContext context, Service service, IEnumerable`1 parameters)
   at Autofac.ResolutionExtensions.Resolve(IComponentContext context, Type serviceType, IEnumerable`1 parameters)
   at Autofac.ResolutionExtensions.Resolve(IComponentContext context, Type serviceType)
   at Cake.Infrastructure.Composition.AutofacTypeResolver.Resolve(Type type) in C:\...\cake\src\Cake\Infrastructure\Composition\AutofacTypeResolver.cs:line 22
   at Spectre.Console.Cli.TypeResolverAdapter.Resolve(Type type) in /_/src/Spectre.Console/Cli/Internal/TypeResolverAdapter.cs:line 25
```

![image](https://user-images.githubusercontent.com/177608/141666354-ac74073d-e8bf-4580-8620-89b0924513db.png)

### What is expected?

No exception thrown / Autofac should be able to resolve `DefaultCommandSettings`

### Steps to Reproduce

- Open the Cake solution
- Debug the execution of an existing `.cake` file

PR #3698, opened 2021-11-14
GH3697: Register DefaultCommandSettings with Autofac

augustoproiete:
Closes https://github.com/cake-build/cake/issues/3697

### DIFF
--- a/src/Cake/Program.cs
+++ b/src/Cake/Program.cs
@@ -71,6 +71,9 @@ namespace Cake
         {
             var builder = new ContainerBuilder();
 
+            // Commands
+            builder.RegisterType<DefaultCommandSettings>();
+
             // Converters
             builder.RegisterType<Cli.FilePathConverter>();
             builder.RegisterType<VerbosityConverter>();


nils-a:
@augustoproiete Could we re- work the AutofacTypeResolver to not throw in this case? It would be the more appropriate solution, I think.
 https://github.com/spectreconsole/spectre.console/pull/620 makes the requirements for ITypeResolver a bit clearer.

devlead (approved):
LGTM 👍

augustoproiete:
> Could we re- work the AutofacTypeResolver to not throw in this case? It would be the more appropriate solution, I think. [spectreconsole/spectre.console#620](https://github.com/spectreconsole/spectre.console/pull/620) makes the requirements for ITypeResolver a bit clearer.

@nils-a I looked at our [`AutofacTypeResolver`](https://github.com/cake-build/cake/blob/v2.0.0-rc0001/src/Cake/Infrastructure/Composition/AutofacTypeResolver.cs) and conceptually it seemed to make sense that it throws given that we were not registering the `DefaultCommandSettings` with our container... Is this something Spectre.Console should do automatically for us and we wouldn't need to register `CommandSettings` derived classes?

PR #3700, opened 2021-11-14
(#3699) ITypeResolvers do not throw on Resolve

### LINKED ISSUE #3699: `AutofacTypeResolver` should not throw when resolving unregistered services
We first saw the problem in https://github.com/cake-build/cake/pull/3585, and added a workaround for that in https://github.com/cake-build/cake/pull/3698.

This is the same issue that surfaced in https://github.com/spectreconsole/spectre.console/issues/555.

When implementing an `ITypeResolver` in Spectre.Console, it should never throw but return `null` for unresolved services. This expectation has been a bit unclear in Spectre.Console, but was recently made a bit clearer in https://github.com/spectreconsole/spectre.console/pull/620 in the [documentation of `ITypeResolver`](https://github.com/nils-a/spectre.console/blob/fd181e274beda6de6a40e8f36dd0643a7874c31f/src/Spectre.Console/Cli/ITypeResolver.cs#L14).

Also, in Spectre.Console.Testing https://github.com/spectreconsole/spectre.console/pull/620/commits/8765dc90025cb9cab6b50bd26bd851356c6798b7, added [`TypeRegistrarBaseTests`](https://github.com/nils-a/spectre.console/blob/8765dc90025cb9cab6b50bd26bd851356c6798b7/src/Spectre.Console.Testing/Cli/TypeRegistrarBaseTests.cs) which can be used to unit test custom implementations of `ITypeRegistrar`/`ITypeResolver`.

We should modify `AutofacTypeResolver` so it adheres to the expectations of Spectre.Console and does not throw on resolving unregistered services.

nils-a:
fixes #3699

### DIFF
--- a/src/Cake.Frosting/Internal/TypeResolver.cs
+++ b/src/Cake.Frosting/Internal/TypeResolver.cs
@@ -18,6 +18,11 @@ namespace Cake.Frosting.Internal
 
         public object Resolve(Type type)
         {
+            if (type == null)
+            {
+                return null;
+            }
+
             return _provider.GetService(type);
         }
     }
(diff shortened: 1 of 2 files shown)


augustoproiete (approved):
LGTM

PR #3710, opened 2021-11-23
GH3709: Add Arguments ICollection fallback

### LINKED ISSUE #3709: Arguments alias should support ICollection<T> as default value
Currently, multiple arguments only support a single fallback value
```cake
ICollection<T> Arguments<T>(this ICakeContext context, string name, T defaultValue)
```
but would make total sense for it to also support a collection as a fallback too
```cake
ICollection<T> Arguments<T>(this ICakeContext context, string name, ICollection<T> defaultValues)
```

I'll send a PR for this shortly.

devlead:
*  fixes #3709


gep13 (approved):
LGTM!

PR #3734, opened 2021-12-06
GH3733: Show relative path of addin assemblies that are being loaded

### LINKED ISSUE #3733: Show relative path of addin assemblies that are being loaded
When Cake loads assemblies of an addin, it displays a diagnostic message that looks like this:

```
The addin Cake.MyAddin will reference Cake.MyAddin.dll.
```

Addins often target multiple TFMs and have a folder structure that looks similar like this:

```
tools\Addins\Cake.MyAddin.2.0.0\lib\net6.0\Cake.MyAddin.dll
tools\Addins\Cake.MyAddin.2.0.0\lib\net5.0\Cake.MyAddin.dll
tools\Addins\Cake.MyAddin.2.0.0\lib\netcoreapp3.1\Cake.MyAddin.dll
```

The current diagnostic message does not show ***which*** of these assemblies were loaded, which is very useful when troubleshooting compatibility issues.

### Proposal

Show relative path of the assembly being loaded. E.g.:

```
The addin Cake.MyAddin will reference tools/Addins/Cake.MyAddin.1.0.0/lib/net6.0/Cake.MyAddin.dll.
```

augustoproiete:
Closes #3733

### DIFF
--- a/src/Cake.Core/Scripting/ScriptProcessor.cs
+++ b/src/Cake.Core/Scripting/ScriptProcessor.cs
@@ -112,7 +112,19 @@ namespace Cake.Core.Scripting
                     // Reference found assemblies.
                     foreach (var assembly in assemblies)
                     {
-                        _log.Debug("The addin {0} will reference {1}.", addin.Package, assembly.Path.GetFilename());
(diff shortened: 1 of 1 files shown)


nils-a (approved):
LGTM!

PR #3791, opened 2022-01-07
Add alias for dotnet store command 

### LINKED ISSUE #3488: Add alias for dotnet store command
`dotnet store` - Stores the specified assemblies in the runtime package store. Introduced with .NET Core 2.x SDK.

`dotnet store` stores the specified assemblies in the runtime package store. By default, assemblies are optimized for the target runtime and framework. For more information, see the runtime package store topic.

Docs: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-store

Marusyk:
Add alias for dotnet store command

Fixes #3488

Please review

### DIFF
--- a/src/Cake.Common/Tools/DotNet/Store/DotNetStoreSettings.cs
+++ b/src/Cake.Common/Tools/DotNet/Store/DotNetStoreSettings.cs
@@ -1,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Core.IO;
+
+namespace Cake.Common.Tools.DotNet.Store
+{
+    /// <summary>
+    /// Contains settings used by <see cref="DotNetStorer" />.
+    /// </summary>
+    public class DotNetStoreSettings : DotNetSettings
+    {
+        /// <summary>
+        /// Gets or sets the output directory.
+        /// </summary>
+        public DirectoryPath OutputDirectory { get; set; }
+
+        /// <summary>
+        /// Gets or sets the version of the installed Shared Framework to use to run the application.
+        /// </summary>
+        public string FrameworkVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to skip the optimization phase.
+        /// </summary>
+        public bool SkipOptimization { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to skip symbol generation.
+        /// </summary>
+        /// <remarks>
+        /// Currently, you can only generate symbols on Windows and Linux.
+        /// </remarks>
+        public bool SkipSymbols { get; set; }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/Cake.Common/Tools/DotNet/Store/DotNetStoreSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public bool SkipOptimization { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to skip symbol generation.
+        /// </summary>
+        /// <remarks>
+        /// Currently, you can only generate symbols on Windows and Linux.
+        /// </remarks>
+        public bool SkipSymbols { get; set; }
```
devlead, line 35: Args in latest SDK missing
* `--use-current-runtime                    Use current runtime as the target runtime.`
* `--working-dir <WORKING_DIR>          The working directory used by the command to execute.`

Marusyk: There is `WorkingDirectory` from `ToolSettings`

devlead: `WorkingDirectory` agree just noticed there was an option for it unsure if there's any semantical difference or not.

`--use-current-runtime` seems to be a common option

https://github.com/dotnet/sdk/blob/df5c30db37d58380f5705250d76552979e4f3375/src/Cli/dotnet/CommonOptions.cs#L59

Marusyk:
Could you please help me with 

> error NETSDK1016: Unable to find resolved path for 'C:\Users\VssAdministrator\.nuget\packages\microsoft.netcore.app.runtime.win-x64\6.0.1\runtimes\win-x64\native\..\..\..\tools\crossgen.exe'. [D:\a\1\s\tests\integration\temp\Cake.Common\Tools\DotNet\hwapp\hwapp.csproj]

devlead:
> Could you please help me with
> 
> > error NETSDK1016: Unable to find resolved path for 'C:\Users\VssAdministrator.nuget\packages\microsoft.netcore.app.runtime.win-x64\6.0.1\runtimes\win-x64\native......\tools\crossgen.exe'. [D:\a\1\s\tests\integration\temp\Cake.Common\Tools\DotNet\hwapp\hwapp.csproj]
> 
> ```
> var path = Paths.Temp.Combine("./Cake.Common/Tools/DotNet");
> var project = path.CombineWithFilePath("hwapp/hwapp.csproj");
> 
> // When
> DotNetStore(project.FullPath, "net6.0", "win-x64");
> ```

Done some testing and it seems crossgen.exe only ships up to .NET 5, so changing `net6.0` to i.e. `netcoreapp3.1` seems to make it happy. Wonder if  `dotnet store` command is supported post .NET 5.0 🤔

pascalberger:
@Marusyk Thank you very much for you pull request for this issue 👍  

We discussed it again in the team and came to the conclusion that we don't want to add it to Cake as the feature seems no longer to be shipped with .NET 6 and we don't see an actual demand for it to be supported in Cake, beside for completeness of supporting all `dotnet` commands. Considering that we would need to add integration test for it and don't see any use cases for the feature we don't think it is worth the additional effort required to get this PR merged.

Please let me know if there are any other aspects which we might have missed to consider or if you want to continue to work on this.

PR #3833, opened 2022-03-04
Add entireProcessTree parameter to Process.Kill and Process.WaitForExit methods

MohsenAra:
I Use Process tools for start npm scripts 
but in some scenario i need to kill process and subprocess 
but process not kill sub process 
i want add entireProcessTree prameter to kill process and wait for exit

### DIFF
--- a/src/Cake.Common/Tools/DotCover/DotCoverProcessRunner.cs
+++ b/src/Cake.Common/Tools/DotCover/DotCoverProcessRunner.cs
@@ -24,7 +24,7 @@ namespace Cake.Common.Tools.DotCover
             {
             }
 
-            public bool WaitForExit(int milliseconds , bool entireProcessTree = false)
+            public bool WaitForExit(int milliseconds, bool entireProcessTree = false)
             {
                 return true;
             }
--- a/src/Cake.Common/Tools/DotCover/DotCoverProcessRunner.cs
+++ b/src/Cake.Common/Tools/DotCover/DotCoverProcessRunner.cs
@@ -24,7 +24,7 @@ namespace Cake.Common.Tools.DotCover
             {
             }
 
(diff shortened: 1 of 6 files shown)


gep13:
@MohsenAra thank you for taking the time to raise this PR!

Did you happen to see our [Contributing document](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md) where we ask to get sign off on a potential change before making changes?  This would have been preferred in this scenario since the changes that you are introducing are breaking in the sense that they break ABI compatability.

We have discussed this as a team, and we think that there is a need going forward to change the default process kill method to include all subprocesses, but this is something that would be undertaken in the next major release of Cake.  Would you be in a position to raise this issue?

In the interim, our suggestion would be that you create your own Cake Module, that provide your own implementation of the IProcess interface, which would then give you the functionality that you require.  You can find out more information about writing modules here:

https://cakebuild.net/docs/extending/modules/

MohsenAra:
OK
thanks

MohsenAra:
I make this extension method for that my be work for anyone 
public static class ProcessHelper
    {
        public static bool WaitForExit(this IProcess process, int milliseconds, bool entireProcessTree)
        {
            var mainProcess = GetProcess(process);
            if (mainProcess.WaitForExit(milliseconds))            
                return true;
            mainProcess.Refresh();
            if (!mainProcess.HasExited)
            {
                mainProcess.Kill(entireProcessTree);
            }
            return false;
        }

        public static Process GetProcess(this IProcess process)
        {
            var field = process.GetType()
                .GetField("_process", BindingFlags.Instance | BindingFlags.NonPublic);
            var mainProcess = (Process)field.GetValue(process);
            return mainProcess;
        }
    }

Issue #2099, opened 2018-03-23
Cache compiled script on disk


patriksvensson:
Would be nice if we could cache compiled script to disk if it haven't changed.

guillaume86:
FYI I experimented a bit with caching the compiled script. I'd like to use Cake as a task runner and the 4 sec penalty seemed a bit much for running tasks that take under 1 sec on their own.

I just used `Emit(path)` to generate an assembly and figured out how to call it by reflection, and it seems encouraging:

```
# normal run
Analyzing build script...
Processing build script...

========================================
Default
========================================
Executing task: Default
Hello World!
Finished executing task: Default

Task                          Duration
--------------------------------------------------
Default                       00:00:00.0114282
--------------------------------------------------
Total:                        00:00:00.0114282


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 5
Milliseconds      : 448
Ticks             : 54489236
TotalDays         : 6,30662453703704E-05
TotalHours        : 0,00151358988888889
TotalMinutes      : 0,0908153933333333
TotalSeconds      : 5,4489236
TotalMilliseconds : 5448,9236

# compile run (build cache and run)
Analyzing build script...
Processing build script...

========================================
Default
========================================
Executing task: Default
Hello World!
Finished executing task: Default

Task                          Duration
--------------------------------------------------
Default                       00:00:00.0146321
--------------------------------------------------
Total:                        00:00:00.0146321


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 5
Milliseconds      : 347
Ticks             : 53478949
TotalDays         : 6,1896931712963E-05
TotalHours        : 0,00148552636111111
TotalMinutes      : 0,0891315816666667
TotalSeconds      : 5,3478949
TotalMilliseconds : 5347,8949

# cached run
Analyzing build script...
Processing build script...

========================================
Default
========================================
Executing task: Default
Hello World!
Finished executing task: Default

Task                          Duration
--------------------------------------------------
Default                       00:00:00.0095663
--------------------------------------------------
Total:                        00:00:00.0095663


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 689
Ticks             : 6890002
TotalDays         : 7,97453935185185E-06
TotalHours        : 0,000191388944444444
TotalMinutes      : 0,0114833366666667
TotalSeconds      : 0,6890002
TotalMilliseconds : 689,0002
```

[...]

PR #3863, opened 2022-04-06
GH2099: Fix and add script cache integration test

devlead:
* fix `ScriptCacheHashPath` not being checked correctly
* add environment variable drive cache path example
* relates to #2099


devlead:
@GeertvanHorrik here's an example, believe it achieves what you're after
https://github.com/cake-build/cake/blob/4cc8c5ed8dbf4c72f1718223101b0fd9db49e9fe/tests/integration/resources/Cake/ScriptCache/Config/cake.config#L1-L2

GeertvanHorrik:
Looks good to me!

Issue #3859, opened 2022-04-04
`PathCollapser.Collapse` breaks UNC paths


nils-a:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake .NET Tool

### Cake version

2,1,0

### Operating system

Linux, Windows

### Operating system architecture

64-Bit

### What are you seeing?

Given the following Cake script:

```cs
var target = Argument("target", "Default");

Task("Default")
.Does(() => {
   var one = new DirectoryPath("\\\\server\\share\\folder\\..");
   var two = one.Collapse();
   Information($"1: {one.FullPath} IsUNC:{one.IsUNC}");
   Information($"2: {two.FullPath} IsUNC:{two.IsUNC}");
});

RunTarget(target);
```

The output is:
```text
========================================
Default
========================================
1: \\server\share\folder\.. IsUNC:True
2: //server/share IsUNC:False
```

### What is expected?

I would expect a collapsed UNC path to start with `\\` and not `//`, also I would expect `IsUNC` to return `true`.

### Steps to Reproduce

See script above.

Issue #3858, opened 2022-04-04
`PathCollapser.Collapse` shows wrong output for if `..` is the second segment in the path


nils-a:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake .NET Tool

### Cake version

2.1.0

### Operating system

Linux, Windows

### Operating system architecture

64-Bit

### What are you seeing?

Given the following script:
```cs
var target = Argument("target", "Default");

Task("Default")
.Does(() => {
   var one = new DirectoryPath("foo/../bar").Collapse();
   var two = new DirectoryPath("./foo/../bar").Collapse();
   Information("1: "+one);
   Information("2: "+two);
});

RunTarget(target);
```

The output is 
```text
========================================
Default
========================================
1: foo/bar
2: foo/bar
```

### What is expected?

For a path `foo/../bar` or `./foo/../bar` I would expect the output being `bar`.

### Steps to Reproduce

See Cake script above.

PR #3865, opened 2022-04-07
(#3858, #3859) Made PathCollapser aware of different formats

nils-a:
Especially different calculations of the root of a path
for windows, non-windows and UNC like paths.

To this end, I re-structured the unit test. There were no test
removed that previously existed, but only new tests added.

Also, instead of joining together the path with one specific
separator ('/'), I re-used the existing separator of the original
path. (This was the main-fix for #3859.)

### DIFF
--- a/src/Cake.Core.Tests/Unit/IO/DirectoryPathTests.cs
+++ b/src/Cake.Core.Tests/Unit/IO/DirectoryPathTests.cs
@@ -359,7 +359,6 @@ namespace Cake.Core.Tests.Unit.IO
                 [InlineData("/C/Data/Work", "/C/Data")]
                 [InlineData("/C/Data/Work/file.txt", "/C/Data/Work")]
                 [InlineData("/folder/foo/..", "/")]
-                [InlineData("/..", ".")] // a bit unexpected, but due to the way "Collapse" works.
                 public void Should_Return_Parent_Directory(string directoryPath, string parentPath)
                 {
                     // Given
@@ -374,6 +373,7 @@ namespace Cake.Core.Tests.Unit.IO
 
                 [NonWindowsTheory]
                 [InlineData("/")]
+                [InlineData("/..")]
                 public void Should_Return_Null_If_No_Parent(string directoryPath)
                 {
                     // Given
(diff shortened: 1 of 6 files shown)


src/Cake.Testing.Xunit/WindowsFact.cs
```diff
@@ -1,29 +0,0 @@
...
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using Cake.Core;
-using Cake.Core.Polyfill;
-using Xunit;
-
-namespace Cake.Testing.Xunit
-{
-    public sealed class WindowsFact : FactAttribute
```
devlead, line 11: As this is a public assembly ( https://www.nuget.org/packages/Cake.Testing.XUnit )
and this is a breaking change should we obsolete first?

devlead (changes requested):

src/Cake.Testing.Xunit/WindowsFact.cs
nils-a: @devlead you're right, I haven't thought of that. I changed the `WindowsFact` to `WindowsFactAttribute` out of habit.

The docs say
> By convention, attribute classes are named with a suffix of Attribute. An *attribute_name* can either include or omit this suffix.

So I'd suggest I revert the name change. This would leave only the change in inheritance, which isn't breaking, right?

devlead: why not make WindowsFact inherit from WindowsFactAttribute and obsolete WindowsFact with a warning.

nils-a: No joy. Having an attribute called `WindowsFact` and one called `WindowsFactAttribute` results in

>   [CS1614] 'WindowsFact' is ambiguous between 'WindowsFact' and 'WindowsFactAttribute'. Either use '@WindowsFact' or explicitly include the 'Attribute' suffix. 

So, it seems while this keeps ABI compatibility, it breaks the API.

devlead: Reasonable, and then we can rename it in 3.0 (add an issue for that).

devlead: renamed

devlead (approved):
LGTM 👍

devlead:
@nils-a  your changes have been merged, thanks for your contribution 👍

PR #3945, opened 2022-08-12
GH3931: Improve posix native library loading

### LINKED ISSUE #3931: Cake fails to load native libraries on Ubuntu 22.04
As discovered in https://github.com/actions/runner-images/issues/6029, Cake fails to load native libraries on Ubuntu 22.04

```
Verifying assembly 'Cake.Sonar, Version=1.1.29.0, Culture=neutral, PublicKeyToken=null'.
Error: System.DllNotFoundException: Unable to load shared library 'libdl' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: liblibdl: cannot open shared object file: No such file or directory
   at Cake.Core.Polyfill.AssemblyHelper.LoadUnixLibrary(String path, Int32 flags)
   at Cake.Core.Polyfill.AssemblyHelper.LoadAssembly(ICakeEnvironment environment, IFileSystem fileSystem, FilePath path) in C:\projects\cake\src\Cake.Core\Polyfill\AssemblyHelper.cs:line 54
```

Will need some investigation but iIssue probably in 

https://github.com/cake-build/cake/blob/c3f74155e88388b2354a5359a0fdde5fa5362ceb/src/Cake.Core/Polyfill/AssemblyHelper.cs#L74-L75

Which likely should on newer distros be
```csharp
[DllImport("libdl.so.2", EntryPoint = "dlopen")] 
private static extern IntPtr LoadUnixLibrary2(string path, int flags); 
```

we probably could have a flag for which to use, unset by default, set when successful or unset and below catch try the other and set if success, so it's a one-time runtime cost.
```csharp
catch (DllNotFoundException)
```

devlead:
* Better RID detection on OSX
* Support libdl 2
* Better OSX dylib support
* Native library integration tests
* Assembly load diagnostic logging
* Build & test on Ubuntu 22.04
* fixes #3931

### DIFF
--- a/src/Cake.Core/Polyfill/AssemblyHelper.cs
+++ b/src/Cake.Core/Polyfill/AssemblyHelper.cs
@@ -5,8 +5,8 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using Cake.Core.Diagnostics;
 using Cake.Core.IO;
-using Cake.Core.Reflection;
 
 namespace Cake.Core.Polyfill
 {
@@ -26,7 +26,7 @@ namespace Cake.Core.Polyfill
             return Assembly.Load(assemblyName);
         }
 
-        public static Assembly LoadAssembly(ICakeEnvironment environment, IFileSystem fileSystem, FilePath path)
+        public static Assembly LoadAssembly(ICakeEnvironment environment, IFileSystem fileSystem, ICakeLog log, FilePath path)
         {
             if (path == null)
             {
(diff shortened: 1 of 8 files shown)


bjorkstromm (approved):
LGTM!

Great work @devlead! Thanks for fixing this!

src/Cake.Core/Polyfill/AssemblyHelper.cs
```diff
@@ -26,7 +26,7 @@ public static Assembly LoadAssembly(AssemblyName assemblyName)
             return Assembly.Load(assemblyName);
         }
 
-        public static Assembly LoadAssembly(ICakeEnvironment environment, IFileSystem fileSystem, FilePath path)
+        public static Assembly LoadAssembly(ICakeEnvironment environment, IFileSystem fileSystem, ICakeLog log, FilePath path)
```
gep13, line 29: Should we regard this as a breaking change?

src/Cake.Core/Reflection/AssemblyLoader.cs
```diff
@@ -17,18 +18,21 @@ public sealed class AssemblyLoader : IAssemblyLoader
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AssemblyLoader"/> class.
         /// </summary>
         /// <param name="environment">The environment.</param>
         /// <param name="fileSystem">The file system.</param>
         /// <param name="verifier">The assembly verifier.</param>
-        public AssemblyLoader(ICakeEnvironment environment, IFileSystem fileSystem, IAssemblyVerifier verifier)
+        /// <param name="log">The cake log.</param>
+        public AssemblyLoader(ICakeEnvironment environment, IFileSystem fileSystem, IAssemblyVerifier verifier, ICakeLog log)
```
gep13, line 30: Should we regard this as a breaking change?

src/Cake.Core/Polyfill/AssemblyHelper.cs
devlead: No, it's an internal class

src/Cake.Core/Reflection/AssemblyLoader.cs
devlead: Don't think so in this case,  interface untouched,  created through IoC, only used before addins are initiated and Cake.Tool

src/Cake.Core/Polyfill/AssemblyHelper.cs
gep13: Sounds good to me!

src/Cake.Core/Reflection/AssemblyLoader.cs
gep13: Sounds good to me!

gep13:

PR #3721, opened 2021-11-29
GH2953: Allow setting MSBuild target via MSBuildSettings using a string

sbwaggoner:
PR to address this feature request:

https://github.com/cake-build/cake/issues/2953

PR #3954, opened 2022-08-28
GH2953: Feature addition target initializer msbuildsettings

FrankRay78:
Regarding the following (incomplete) PR which is now over 6 months old, https://github.com/cake-build/cake/pull/3721, I have rebased @sbwaggoner's original changes on to an up-to-date `develop` branch and made the change recommended by @devlead on 15 December 2021 during the PR review. Several unit tests have also been added to increase the code coverage of this area too.

Regarding the conditional StringSplitOptions approach recommended in the review, I tried implementing it as suggested but given .Net Core 3.1 StringSplitOptions enum doesn't include the StringSplitOptions.TrimEntries option, trimming still needed to be implemented manually in the code for that target framework. So the conditional compile logic didn't end up offering any benefit imho.

### DIFF
--- a/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs
+++ b/src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Cake.Core.Diagnostics;
 using Cake.Core.Tooling;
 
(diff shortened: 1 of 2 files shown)


src/Cake.Common/Tools/MSBuild/MSBuildSettings.cs
```diff
@@ -47,6 +48,24 @@ public sealed class MSBuildSettings : ToolSettings
...
+        {
+            get => string.Join(";", Targets);
+            set
+            {
+                Targets.Clear();
+
+                foreach (var target in value?.Split(';').Where(p => !string.IsNullOrEmpty(p?.Trim())) ?? Array.Empty<string>())
+                {
+                    Targets.Add(target.Trim());
+                }
```
devlead, line 65:
```suggestion
                Targets.Clear();

                if (string.IsNullOrWhiteSpace(value))
                {
                    return;
                }

                foreach (var target in value
                                            .Split(';')
                                            .Select(target => target.Trim())
                                            .Where(target => !string.IsNullOrEmpty(target)))
                {
                    Targets.Add(target);
                }
```

devlead (changes requested):
Thanks for submitting a PR @FrankRay78 👍  Added some suggestions.

devlead (approved):
LGTM 👍

devlead:
@FrankRay78 & @sbwaggoner  your changes have been merged, thanks for your contributions 👍

PR #3989, opened 2022-10-17
(build) Set next version to 3.0.0

gep13:
There are upcoming breaking changes in Cake, so we need to start building version 3.0.0, rather than 2.4.0.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,4 +1,4 @@
-next-version: 2.0.0
+next-version: 3.0.0
 branches:
   master:
     regex: main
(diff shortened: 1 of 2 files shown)


devlead (approved):
LGTM 👍 

We should also bump the potential breaking change
https://github.com/cake-build/cake/blob/7c2b3b2d5dc75a4248108abe5bc6a6dd6a4753a9/src/Cake.Core/Constants.cs#L14

gep13:
@devlead ah, good catch! I have added that change into this PR.

PR #4007, opened 2022-10-22
GH4006: Add missing GitVersion CLI options

### LINKED ISSUE #4006: Add missing GitVersion command line options
There are a number of newer command line options that can be passed through to GitVersion which are not currently implemented in the Cake settings.  While it is possible to use ArgumentCustomization to make use of these options, they should also be supported directly.

These options are:

- OutputFile
- ConfigFile
- UpdateProjectFiles
- EnsureAssemblyInfo
- UpdateWixVersionFile
- NoCache
- NoNormalize
- Diag

gep13:
Since the initial implementation in Cake, there are a number of new command line options for GitVersion. While it is possible to set these using ArgumentCustomization, these should be supported directly within the Cake settings class.

Fixes #4006


devlead (approved):
LGTM 👍

devlead:

PR #4022, opened 2022-10-25
GH3946: Add ISetupContext to Frosting lifetime

### LINKED ISSUE #3946: Access to ITaskSetupContext in Frosting
> Sorry to resurrect an old thread, but does someone know how to do this for a 
> "Frosting" project? i.e. to get access to `ISetupContext.TasksToExecute`.

> It is exposed in `FrostingTaskLifetime<T>.Setup` but that is before each task. 
> I can't find an equivalent for `FrostingLifetime<T>.Setup` which is before all tasks. 
> Maybe I'm looking in the wrong places?

_Originally posted by @lonix1 in https://github.com/cake-build/cake/issues/1772#issuecomment-1220288335_

A quick look at the code, it looks like currently, the SetupContext isn't passed to the Frosting setup, adding it to `IFrostingLifetime` / `IFrostingSetup` would be a breaking change, so maybe first fixed in the next mayor release, or at least some more thought will need to go into it.
https://github.com/cake-build/cake/blob/a1893a2bba26f49e089d66beeb78421cedf6126d/src/Cake.Frosting/Internal/FrostingEngine.cs#L83

like it is for the teardown
https://github.com/cake-build/cake/blob/a1893a2bba26f49e089d66beeb78421cedf6126d/src/Cake.Frosting/Internal/FrostingEngine.cs#L89

A dirty workaround, in this case, could be to register the setup task yourself.

This could be done by injecting a custom setup task in IoC and adding it and `ICakeEngine` to your own custom context. that could  look something like

```csharp
using System.Threading.Tasks;
using Cake.Core;
using Cake.Core.Diagnostics;
using Cake.Frosting;
using Microsoft.Extensions.DependencyInjection;

public static class Program
{
    public static int Main(string[] args)
    {
        return new CakeHost()
            .ConfigureServices(services => services.AddSingleton<CustomBuildSetup>())
            .UseContext<BuildContext>()
            .Run(args);
    }
}

public class BuildContext : FrostingContext
{
    public BuildContext(ICakeContext context, ICakeEngine cakeEngine, CustomBuildSetup customBuildSetup)
        : base(context)
    {
        cakeEngine.RegisterSetupAction(setupContext => customBuildSetup.Setup(this, setupContext));
    }
}

public class CustomBuildSetup
{
    public void Setup(BuildContext buildContext, ISetupContext setupContext)
    {
        buildContext.Log.Information("Custom setup found {0} tasks starting with {1}", setupContext.TasksToExecute.Count, setupContext.TargetTask.Name);
    }
}
```

devlead:
* fixes #3946

### DIFF
--- a/src/Cake.Frosting.Tests/Fakes/FakeLifetime.cs
+++ b/src/Cake.Frosting.Tests/Fakes/FakeLifetime.cs
@@ -11,7 +11,7 @@ namespace Cake.Frosting.Tests
         public int SetupCount { get; set; }
         public int TeardownCount { get; set; }
 
-        public override void Setup(ICakeContext context)
+        public override void Setup(ICakeContext context, ISetupContext info)
         {
             SetupCount++;
         }
--- a/src/Cake.Frosting/FrostingLifetime.cs
+++ b/src/Cake.Frosting/FrostingLifetime.cs
@@ -28,7 +28,8 @@ namespace Cake.Frosting
         /// If setup fails, no tasks will be executed but teardown will be performed.
         /// </summary>
         /// <param name="context">The context.</param>
-        public abstract void Setup(TContext context);
+        /// <param name="info">The setup information.</param>
+        public abstract void Setup(TContext context, ISetupContext info);
 
         /// <summary>
         /// This method is executed after all tasks have been run.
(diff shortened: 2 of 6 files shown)


gep13 (approved):
LGTM!

PR #4056, opened 2022-11-07
GH4055: Add task skip reason to task summary output

### LINKED ISSUE #4055: Add column to summary to include skip reason
There are times when running a build, when one or more tasks are skipped.  Currently, in order to know what caused the task to be skipped, you have to be running in diagnostic mode, and also you have to scroll up through the build output to see the skip reason.

I think it would be useful to include the skip reason for a task in the summary output that is reported at the end of a build.

gep13:
Via a new setting, --Settings_UseSpectreConsoleForConsoleOutput, it is
now possible to output the task headers, and task summary using
Spectre.Console.  This allows for the easy addition of a new column in
the task summary, which includes information about _why_ any given task
has been skipped.

The old CakeReportPrinter is still in play, but some implementation has
been moved from within the DefaultExecutionStrategy to the reporter
class.  Depending on the setting mentioned above, either the
Spectre.Console version of the reporter will be added to the IoC
container, or the old version will be in place.

Fixes #4055

### DIFF
--- a/src/Cake.Cli/Infrastructure/CakeSpectreReportPrinter.cs
+++ b/src/Cake.Cli/Infrastructure/CakeSpectreReportPrinter.cs
@@ -1,0 +1,162 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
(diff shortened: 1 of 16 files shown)


gep13:
Original Cake output:

image

Spectre.Console Output with no tasks being skipped:

image

Spectre.Console Output with a task that is being skipped:

image

devlead (approved):
LGTM 👍

devlead:

Issue #1317, opened 2016-10-25
CleanDirectory does not clean readonly files.


frozenskys:
### What You Are Seeing?

Trying to run `CleanDirectory()` on a folder that has had some `readonly` files copied into it using `CopyFiles()` The issue is that `CleanDirectory()` fails with access denied on the read only files.
### What is Expected?

`CleanDirectory()` deletes the files even if they are `readonly`.
### What version of Cake are you using?

0.16.2
### Are you running on a 32 or 64 bit system?

x64
### What environment are you running on?  Windows? Linux? Mac?

Windows
### Are you running on a CI Server?  If so, which one?

N\A
### How Did You Get This To Happen? (Steps to Reproduce)

Add some `readonly` xsd files to `./Documents/Schemas/` and run the following cake file twice...

``` csharp
Task("Default")
    .Does(() =>
    {
            var schemaFolder = Argument("schemaFolder", "./artifacts/schemas");
            EnsureDirectoryExists(schemaFolder);
            CleanDirectory(schemaFolder);
            var files = GetFiles("./Documents/Schemas/*.xsd");
            CopyFiles(files, schemaFolder);
        });

RunTarget("Default");
```
### Output Log

First Run

``` text
Preparing to run build script...
Running build script...
Analyzing build script...
Processing build script...
Compiling build script...

========================================
Default
========================================
Executing task: Default
Creating directory C:/Source/Main/artifacts/schemas
Cleaning directory C:/Source/Main/artifacts/schemas
Copying file main.xsd to C:/Source/Main/artifacts/schemas/main.xsd
Copying file sub.xsd to C:/Source/Main/artifacts/schemas/sub.xsd
Finished executing task: Default

Task                          Duration            
--------------------------------------------------
Default                       00:00:00.1268127    
--------------------------------------------------
Total:                        00:00:00.1268127   
```

Second Run

``` text
Preparing to run build script...
Running build script...
Analyzing build script...
Processing build script...
Compiling build script...

========================================
Default
========================================
Executing task: Default
Cleaning directory C:/Source/Main/artifacts/schemas

An Error occured when executing task 'default'.
Error: Access to the path 'C:/Source/Main/artifacts/schemas/main.xsd' is denied.
```

nrjohnstone:
Not sure if it should be modified to do this... Surely it would be better if you want this behavior to iterate over the files/dirs first and remove the readonly attribute?

devlead:
[...]

PR #4077, opened 2022-11-22
Ability to force clean readonly files

FrankRay78:
This PR successfully addresses #1317, 'CleanDirectory does not clean readonly files.'

I have familarised myself with #1574 when making this change (as @bjorkstromm mentioned) and in a similar manner, introduced a `CleanDirectorySettings` class which has a `Force` flag that defaults to false, but when enabled, will ensure the `DirectoryCleaner` will force clean readonly files.

I avoided making this a breaking change by creating overloads of the `DirectoryCleaner` methods that take an additional `CleanDirectorySettings` parameter, and did the same for the various `DirectoryAliases` methods. 

If it was, however, decided that every clean method should take a settings parameter, we can keep the overloaded methods and simply remove the initial methods that didn't accept `CleanDirectorySettings`.


augustoproiete (approved):
LGTM.

Addition to the API (`CleanDirectorySettings` & `Force`) follows prior art such as [`DeleteDirectorySettings `](https://github.com/cake-build/cake/blob/v3.0.0/src/Cake.Common/IO/DeleteDirectorySettings.cs#L26)

augustoproiete:
@FrankRay78 your changes have been merged, thanks for your contribution :+1:

PR #4110, opened 2023-01-21
GH-4109: Add PublishReadyToRun to DotNetRestoreSettings

### LINKED ISSUE #4109: Add PublishReadyToRun to DotNetRestoreSettings
### Discussed in https://github.com/orgs/cake-build/discussions/4108

<div type='discussions-op-text'>

<sup>Originally posted by **Trisped** January 19, 2023</sup>
According to https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/publish-readytorun-requires-restore-change#recommended-action :
> In .NET 6, dotnet restore followed by dotnet publish -p:PublishReadyToRun=true --no-restore will fail with the NETSDK1095 error. This is because the crossgen binary is now shipped as a separate NuGet package, and so needs to be part of the restore operation for publishing to succeed.

According to https://cakebuild.net/api/Cake.Common.Tools.DotNet.Restore/DotNetRestoreSettings/ , there is not a PublishReadyToRun property like there is in DotNetPublishSettings.

There is a work around (set NoRestore to false), but it requires restoring twice, if you have already restored (for example, so you could test before publishing).</div>

Marusyk:
Please review. 

Fixes #4109

### DIFF
--- a/src/Cake.Common.Tests/Unit/Tools/DotNet/Restore/DotNetRestorerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/DotNet/Restore/DotNetRestorerTests.cs
@@ -216,6 +216,20 @@ namespace Cake.Common.Tests.Unit.Tools.DotNet.Restore
                 // Then
                 Assert.Equal($@"restore --force-evaluate", result.Args);
             }
+
+            [Fact]
+            public void Should_Add_PublishReadyToRun()
+            {
+                // Given
+                var fixture = new DotNetRestorerFixture();
+                fixture.Settings.PublishReadyToRun = true;
+
+                // When
+                var result = fixture.Run();
+
+                // Then
+                Assert.Equal("restore -p:PublishReadyToRun=true", result.Args);
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


augustoproiete (approved):
LGTM

Marusyk:
so,  can we merge it?

augustoproiete:
Thanks for the ping @Marusyk . We're looking into the impact on addins, as this is a breaking change to the public API, to decide if we can merge it now for v3.1.0 or if it has to wait for the next major.

Marusyk:
@augustoproiete could you explain why is this a breaking change?

augustoproiete:
@Marusyk It's a binary breaking change that would only affect existing Cake addins. Addins that were compiled with an existing Cake version are referencing a type called `DotNetRestoreSettings` that **does not** have the property `PublishReadyToRun` and thus technically not the same type.

i.e. `CakevNext...DotNetRestoreSettings` might not be convertible to `Cakev3.0...DotNetRestoreSettings`

Marusyk:
but this is why we have versions, isn't it? I'm sure that Cake add-ins don't compile on the main branch and always use the latest source. 
adding an optional parameter is not a breaking change from API perspective

PR #4136, opened 2023-04-02
GH-4128: Fix Inconsistent determination of positional Argument when using context.Arguments versus context.Argument

### LINKED ISSUE #4128: Inconsistent determination of positional Argument when using context.Arguments versus context.Argument
### Prerequisites

### Cake runner

Cake Frosting

### Cake version

3.0.0

### Operating system

Windows

64-Bit

### CI Server

_No response_

### What are you seeing?

When I am running a task, and I have an IFrostingContext, I have 4 options to get an argument:
1) ArgumentAliases.Argument<T>(IFrostingContext, string) 
* Requires the argument and displays an error if there is none displayed
2) IFrostingContext.ICakeContext.GetArguments()
* Returns a dictionary of collections for all the arguments, which puts the onus on the user to get the argument
3) IFrostingContext.ICakeContext.GetArguments(string)
* Returns a collection of arguments for the provided string, which once again puts the onus on the user to get the Argument
4) ICakeContextExtensions.GetArgument(ICakeContext, string)
* Returns the "Last" instance of an argument provided to the CLI

Since 2 and 3 require the user to do extra work, we'll disregard those, unless they're helpful or informational. So to compare 1 vs 4:

1) Returns and requires an argument to be present. Returns the **FIRST** instance of the argument.
4) Returns and does not require an argument to be present. Returns the **LAST** instance of the argument

We need consistency, because there is an actual use-case for both. Sometimes I want to use ArgumentAliases.Argument<T> so that I don't have to write in my own Error Message, and mess with the AnsiConsole in the same way that that function messes with it.

### What is expected?

I would expect that when using:

ArgumentAliases.Argument<T>(IFrostingContext, string) 

and 

ICakeContextExtensions.GetArgument(ICakeContext, string)

I would get the same result.

### Steps to Reproduce

Create a task in which you pass in an argument. Then, try to retrieve the argument using the two methods described:
.\build.ps1 --required_single=0 --required_single=1

var foo = context.Argument<String>("required_single");
var bar = context.Arguments.GetArgument("required_single");

See that foo is 0, and bar is 1.

### Output log

_No response_

augustoproiete:
Closes #4128

### DIFF
--- a/src/Cake.Common/ArgumentAliases.cs
+++ b/src/Cake.Common/ArgumentAliases.cs
@@ -78,7 +78,7 @@ namespace Cake.Common
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var value = context.Arguments.GetArguments(name).FirstOrDefault();
+            var value = context.Arguments.GetArguments(name).LastOrDefault();
             if (value == null)
             {
                 const string format = "Argument '{0}' was not set.";
@@ -248,7 +248,7 @@ namespace Cake.Common
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var value = context.Arguments.GetArguments(name)?.FirstOrDefault();
+            var value = context.Arguments.GetArguments(name)?.LastOrDefault();
             return value == null
                 ? defaultValue
                 : Convert<T>(value);


augustoproiete:
Closing this in favor of https://github.com/cake-build/cake/pull/4139

PR #4148, opened 2023-04-14
GH2101: Disable report printing if verbosity of logger is Quiet

### LINKED ISSUE #2101: CakeReportPrinter should be disabled in Verbosity = Quiet
I expect no extra lines from cake-build in Verbosity Quiet mode.

deniszykov:
fix for https://github.com/cake-build/cake/issues/2101

### DIFF
--- a/src/Cake.Cli/Hosts/BuildScriptHost.cs
+++ b/src/Cake.Cli/Hosts/BuildScriptHost.cs
@@ -1,10 +1,11 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Cake.Cli.Infrastructure;
 using Cake.Core;
 using Cake.Core.Configuration;
 using Cake.Core.Diagnostics;
@@ -96,7 +97,7 @@ namespace Cake.Cli
             {
                 var report = await Engine.RunTargetAsync(_context, _executionStrategy, Settings).ConfigureAwait(false);
 
-                var noReportEnabled = _configuration.GetValue("Settings_NoReport") ?? bool.FalseString;
+                var noReportEnabled = _configuration.GetValue(Constants.Settings.NoReport) ?? bool.FalseString;
                 if (report != null && !report.IsEmpty && !noReportEnabled.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase))
                 {
                     _reportPrinter.Write(report);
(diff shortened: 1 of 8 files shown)


deniszykov:

src/Cake.Core/CakeReportPrinter.cs
```diff
@@ -30,13 +30,18 @@ public CakeReportPrinter(IConsole console, ICakeContext context)
...
             if (report == null)
             {
                 throw new ArgumentNullException(nameof(report));
             }
 
+            if (verbosity <= Verbosity.Quiet)
+            {
+                return;
+            }
+
```
augustoproiete, line 44: Thanks for the PR @deniszykov! As you can see in [this comment in the original issue](https://github.com/cake-build/cake/issues/2101#issuecomment-376160140) we wouldn't want to change the semantics of the `Quiet` verbosity in order to hide the report - this means users don't have a way to suppress the log messages and see the report anymore.

Could you add a new command-line argument to Cake called `--no-report` that hides the report? (regardless of the Verbosity)

augustoproiete (dismissed):

src/Cake.Core/CakeReportPrinter.cs
gep13: I have taken the liberty to update this PR with the suggested changes.

gep13:
@deniszykov rather than create a new PR, I have taken the liberty to update this PR using the suggested approach discussed in the original issue.

Please let me know if you have any concerns about me doing this.

deniszykov:
> @deniszykov rather than create a new PR, I have taken the liberty to update this PR using the suggested approach discussed in the original issue.
> 
> Please let me know if you have any concerns about me doing this.

I don't mind, and thanks. I was actually thinking of adding another logging level below Quiet.  But I solved my problem differently and forgot about this PR.

gep13:
@deniszykov said...
I don't mind, and thanks.

Perfect! I have started a process of reviewing some of the older PR's that we have, and I stumbled across your one.

src/Cake.Cli/Hosts/BuildScriptHost.cs
```diff
@@ -88,7 +96,8 @@ private async Task<CakeReport> internalRunTargetAsync()
             {
                 var report = await Engine.RunTargetAsync(_context, _executionStrategy, Settings).ConfigureAwait(false);
 
-                if (report != null && !report.IsEmpty)
+                var noReportEnabled = _configuration.GetValue("Settings_NoReport") ?? bool.FalseString;
```
gep13: @devlead I tried to use the `Constants` class here, rather than the hard coded `Settings_NoReport`, but it is marked `internal`.  Should I create a new `Constants` class in the Cake.Cli project?

devlead: @gep13 If it's just used in Cake.Cli it makes sense to add a new Constants class in Cake.Cli project.

devlead (approved):
LGMT 👍

devlead:
@deniszykov & @gep13  your changes have been merged, thanks for your contribution 👍

PR #4164, opened 2023-05-31
Add Azure Pipelines group logging commands

fl3pp:
Add Azure Logging Command to create collapsible groups. See: 

[MSDocs: Logging commands -> Formatting commands](https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#formatting-commands)

There were some changes in other logging commands as well. This is because they contain redundant whitespaces, which ADO does tolerate sometimes, but not always, as is the case for the `##[group]` command. It is better to avoid them if possible.


fl3pp:
> microsoft

@microsoft-github-policy-service agree [company="Hamilton Bonaduz AG"]

devlead (approved):
LGTM 👍

devlead:
@fl3pp your changes have been merged, thanks for your contribution 👍

fl3pp:
@devlead, great, thanks for taking time to review and correct 😊

PR #4229, opened 2023-10-11
Add DotNetListPackage alias for dotnet list package command

### LINKED ISSUE #4224: Add DotNetListPackage alias for dotnet list package command
`dotnet list package` - Lists the package references for a project or solution.

The `dotnet list package` command provides a convenient option to list all NuGet package references for a specific project or a solution.

Docs: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-list-package

Marusyk:
Add an alias for `dotnet list package` command

Fixes #4224

Here is the basic implementation. I need your suggestion on what we should do with the output.
If the format is `console` we will receive smth like this
image
- hard to parse into model.
If the format is json, there could be a several different output which is dependent on the input params
image
image
when error
image

If we allow only JSON, then I'll be able to create models and deserialize (for `console` format it is almost impossible) and return the result for `public ListResult DotNetListPackage(...)` but what if someone just wants to get the output and that's all. 
What do you think?

Thank you in advance

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/DotNet/Package/List/DotNetPackageListerFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/DotNet/Package/List/DotNetPackageListerFixture.cs
@@ -1,0 +1,49 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Common.Tools.DotNet.Package.List;
+
+namespace Cake.Common.Tests.Fixtures.Tools.DotNet.Package.List
+{
+    internal sealed class DotNetPackageListerFixture : DotNetFixture<DotNetPackageListSettings>
+    {
+        public string Project { get; set; }
+        public DotNetPackageList Result { get; set; }
+
+        public void GivenPackgeListResult()
+        {
+            ProcessRunner.Process.SetStandardOutput(new string[]
+            {
+                "{",
+                "  \"version\": 1,",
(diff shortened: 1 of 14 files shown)


devlead:
@Marusyk going the JSON route seems like the best option and an object model provides the most value in a Cake context.

devlead:
> ok, so should we allow only `--format json` and have the alias like `public ListPackage List(Project, Settings)`?

Yes, think that makes sense.

Marusyk:
Can someone help me find the model for that json result? 
Depending on the parameters, the result is different every time. I would like to know all the possible attributes that can be returned and make a model for them.

devlead:
Thin it's just a wrapper around NuGet.CommandLine.XPlat which probably means the code is somewhere around here

https://github.com/NuGet/NuGet.Client/tree/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/ListPackage

Marusyk:
Useful: https://github.com/NuGet/Home/blob/dev/accepted/2022/DotnetListPackageMachineReadableJsonOutput.md
It's not quite written in a schema way, but it contains all the examples.

Marusyk:
thanks @augustoproiete . can you help me? I can't figure out why the tests are failing

Marusyk:
can anyone help me to fix this PR?)

devlead:
> can anyone help me to fix this PR?)

Rebased against latest, fixed the failing unit tests and integration test (needed to ensure project restored before executing as integration test cleans between tfms).

devlead (approved):
LGTM 👍

devlead:
@Marusyk your changes have been merged, thanks for your contribution 👍

Issue #3291, opened 2021-03-25
Unable to retrieve target argument with Frosting


nathangiuliani:
### What You Are Seeing?
Unable to retrieve the value passed to the `target` command line argument by using the Arguments method. - e.g. `context.Argument<string>("target")` throws an exception, running `context.HasArgument("target")` returns false even when a target argument is provided, and browsing the arguments object while in a live debug quite clearly shows they are not there.

Note that verbosity does not appear to be available any more also.

These both worked in 0.38.5, but stopped working at some point during 1.0.0 release candidates.

Have tested in a cake script with 1.1.0, and this still works as it did in 0.38.5 - this seems to be specific to Frosting.

Original discussion at https://github.com/cake-build/cake/discussions/3290

### What is Expected?
The value for the target argument to be returned.

### What version of Cake are you using?
1.1.0

### Are you running on a 32 or 64 bit system?
64

### What environment are you running on?  Windows? Linux? Mac?
Windows

### Are you running on a CI Server?  If so, which one?
n/a - occurs locally or in a pipeline

### How Did You Get This To Happen? (Steps to Reproduce)
See my example repository at https://github.com/nathangiuliani/cake-targettest-frosting. The first commit has everything working with 0.38.5, and the last has it updated to 1.1.0 to demonstrate the issue.

Ensure to execute with a target argument, e.g. `--target=Another`, to illustrate this issue the code throws an exception if no target argument is found. This will occur with 1.1.0 no matter what arguments are provided, but will work fine on 0.38.5.

### Output Log
```
NuGet Config not specified. Will use NuGet default mechanism for resolving it.
Registering task: Another
Registering task: Default
Registering setup: Lifetime
Registering teardown: Lifetime

----------------------------------------
Setup
----------------------------------------
Executing custom setup action...
Setting things up...

----------------------------------------
Teardown
----------------------------------------
Executing custom teardown action...
Tearing things down...
Error: System.ArgumentNullException: Target argument is not available
 ---> Cake.Core.CakeException: Argument 'target' was not set.
   at Cake.Common.ArgumentAliases.Argument[T](ICakeContext context, String name)
   at Lifetime.Setup(Context context) in c:\code\cake-targettest-frosting\build\Lifetime.cs:line 15
   --- End of inner exception stack trace ---
   at Lifetime.Setup(Context context) in c:\code\cake-targettest-frosting\build\Lifetime.cs:line 19
[...]

Issue #4157, opened 2023-05-23
Upgrading to spectre.console 0.47.0 breaks the cake build


FrankRay78:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake .NET Tool

### Cake version

Most recent develop branch (b55239f748f921edc303d644839183e9727e00a4)

### Operating system

Linux, Windows, macOS

### Operating system architecture

64-Bit

### CI Server

GitHub action on the main cake-build/cake repo

### What are you seeing?

Something in spectre.console 0.47.0 breaks the Cake build, I've verified it conclusively on the following PR: https://github.com/cake-build/cake/pull/4156 

I took the latest upstream develop branch and pushed it - the build passes. I then upgrade the spectre.console NuGet packages to 0.47.0 and re-push - the build fails with timing issues

The failure is originating in `tests\integration\Cake\ScriptCache.cake`, namely:

```csharp
139            Assert.True(data.CompileResult.Elapsed > data.ExecuteResult.Elapsed, $"Compile time {data.CompileResult.Elapsed} should be greater than execute time  {data.ExecuteResult.Elapsed}.");
140            Assert.Equal(data.CompileResult.Hash, data.ExecuteResult.Hash);
```

And also

```csharp
152        Assert.True(data.ReCompileResult.Elapsed> data.ExecuteResult.Elapsed, $"ReCompileTime time {data.ReCompileResult.Elapsed} should be greater than execute time  {data.ExecuteResult.Elapsed}.");
153        Assert.NotEqual(data.CompileResult.Hash , data.ReCompileResult.Hash);
```

Even when I commented out the first Assert statement, the NotEqual Hash assert still fails (see commit: https://github.com/cake-build/cake/pull/4156/commits/bbb2324259f17b18d0a89d44db8f51f6594fe5a0)

### What is expected?

The build to pass

### Steps to Reproduce

1. Create a branch from the most recent cake `develop` branch
2. Open the solution and upgrade both spectre.console NuGet packages to 0.47.0
3. Build and run tests locally (should all pass)
4. Check-in and push, create PR
5. Build then fails

### Output log

https://github.com/cake-build/cake/actions/runs/5055159519
https://github.com/cake-build/cake/actions/runs/5055159519/jobs/9070991328#step:5:1035

[...]

PR #4236, opened 2023-10-24
Upgrade spectre.console to 0.47.0, modifying the Cake codebase to fix a breaking change this new version includes

FrankRay78:
This PR fully addresses https://github.com/cake-build/cake/issues/4157, which was preventing spectre.console from being upgraded to 0.47.0. The PR also includes the actual upgrade to spectre.console 0.47.0

The approach taken is very similar to a previous Cake.Frosting issue, which has already been fixed: https://github.com/cake-build/cake/issues/3291 

This PR follows the same pattern introduced for the fix, which was located in Cake.Frosting.Internal.DefaultCommand, here: https://github.com/cake-build/cake/blob/c3cd2af8d73b4f270d6abe585c41f049ea92a0d7/src/Cake.Frosting/Internal/Commands/DefaultCommand.cs#L82

Within Cake proper (rather than Cake.Frosting)... Cake.Commands.DefaultCommand.Execute(...) has been modified to explicitly create the CakeArguments from CommandContext.Remaining, adding in the DefaultCommandSettings.Recompile flag, if necessary. 

This PR has indirectly resulted in a move towards to using `CakeArguments` rather than the spectre.console `IRemainingArgument`, which seems a lot cleaner and self-contained.

----

**Full disclosure:** I'm a little uncomfortable submitting a PR with no unit tests, however my reasons for doing it are as follows:

1. `DefaultCommand`, both Cake or Cake.Frosting ones, have entirely zero unit tests (rather, they seem to represent the top level container, and as such aren't tested)
2. https://github.com/cake-build/cake/issues/3291, which this PR was modelled on, didn't see the need to introduce unit tests to cover the IRemainingArguments behaviour
3. It's not entirely clear whether this is even the right level for unit testing

However, and relevant to both Cake and Cake.Frosting DefaultCommand's, I would consider introducing tests for this argument handling behaviour, if deemed necessary. 

[It might be some kind of factory to create the `CakeArguments`, added into the IServiceCollection, and injected into the DefaultCommand, two versions of factory I guess, one for Cake and one for Cake.Frosting (given both commands have different DefaultCommandSettings classes). But that's just some thinking off the top of my head]

### DIFF
--- a/src/Cake.Cli/Cake.Cli.csproj
+++ b/src/Cake.Cli/Cake.Cli.csproj
@@ -19,7 +19,7 @@
 
     <ItemGroup>
       <PackageReference Include="Autofac" Version="7.1.0" />
-      <PackageReference Include="Spectre.Console" Version="0.46.0" />
-      <PackageReference Include="Spectre.Console.Cli" Version="0.46.0" />
+      <PackageReference Include="Spectre.Console" Version="0.47.0" />
+      <PackageReference Include="Spectre.Console.Cli" Version="0.47.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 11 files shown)


devlead (approved):
LGTM 👍

devlead:
@FrankRay78 your changes have been merged, thanks for your contribution 👍

PR #4237, opened 2023-10-27
Add defaultValue param to the GetArgument helper

harry-cpp:
This PR adds ability to specify the desired default value of an argument when retrieving it with the helper function with cake frosting, ie:
```cs
ArtifactsDir = context.Arguments.GetArgument("artifactsDir", "artifacts");
```

### DIFF
--- a/src/Cake.Core/Extensions/CakeArgumentsExtensions.cs
+++ b/src/Cake.Core/Extensions/CakeArgumentsExtensions.cs
@@ -17,6 +17,22 @@ namespace Cake.Core
         /// </remarks>
         /// <param name="arguments">The arguments.</param>
         /// <param name="name">The argument name.</param>
+        /// <param name="defaultValue">The default value to use if the argument is not found.</param>
+        /// <returns>The argument value.</returns>
+        public static string GetArgument(this ICakeArguments arguments, string name, string defaultValue)
+        {
+            return arguments.GetArgument(name) ?? defaultValue;
+        }
+
+        /// <summary>
+        /// Gets the value for an argument.
+        /// </summary>
+        /// <remarks>
+        /// If multiple arguments with the same name are
+        /// specified, the last argument value is returned.
+        /// </remarks>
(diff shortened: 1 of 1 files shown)


harry-cpp:

src/Cake.Core/Extensions/CakeArgumentsExtensions.cs
```diff
@@ -17,10 +17,11 @@ public static class CakeArgumentsExtensions
         /// </remarks>
         /// <param name="arguments">The arguments.</param>
         /// <param name="name">The argument name.</param>
+        /// <param name="defaultValue">The default value to use if the argument is not found.</param>
         /// <returns>The argument value.</returns>
-        public static string GetArgument(this ICakeArguments arguments, string name)
+        public static string GetArgument(this ICakeArguments arguments, string name, string defaultValue = null)
         {
-            return arguments.GetArguments(name).LastOrDefault();
+            return arguments.HasArgument(name) ? arguments.GetArguments(name).LastOrDefault() : defaultValue;
```
devlead, line 24: Adding an optional parameter will break library ABI, to not break ABI an overload should be added instead, something like 
```csharp
 public static string GetArgument(this ICakeArguments arguments, string name, string defaultValue)
   => arguments.GetArgument(name) ?? defaultValue;
```

devlead (changes requested):
The current proposed change will break ABI, is there an issue related to this change according to [CONTRIBUTING.md](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md)?

src/Cake.Core/Extensions/CakeArgumentsExtensions.cs
harry-cpp: I see, done!

harry-cpp:
I just noticed that `context.Arguments`  is both a property and a method, pardon the PR this change is redundant as the method version of it covers both reading the argument and getting the default value 😅

PR #4253, opened 2023-11-14
GH4251/4250/4249/4248/4247: Update dependencies

devlead:
* (GH-4251) Update System.Reflection.Metadata 8.0.0
* (GH-4250) Update System.Collections.Immutable to 8.0.0
* (GH-4249) Update Microsoft.Extensions.DependencyInjection to 8.0.0
* (GH-4248) Update Microsoft.NET.Test.Sdk to 17.8.0
* (GH-4247) Update xunit to 2.6.1

### DIFF
--- a/src/Cake.Common.Tests/Cake.Common.Tests.csproj
+++ b/src/Cake.Common.Tests/Cake.Common.Tests.csproj
@@ -16,7 +16,7 @@
   </ItemGroup>
   <!-- Global packages -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -25,6 +25,11 @@ jobs:
         with:
           dotnet-version: '6.0.x'
 
+      - name: Install .NET SDK 7.0.x
```
pascalberger, line 28: Why is this required?

pascalberger (changes requested):

.github/workflows/build.yml
devlead: It installs latest .NET 7 SDK as global json will install .NET 8.0.x SDK,

But it broke many things, so will revert and take it with .NET 8 support PR.

pascalberger (approved):

PR #4261, opened 2023-11-17
GH4260: Unzip alias should support overwrite files

### LINKED ISSUE #4260: Unzip alias should support overwrite files
I just hit this problem too.

@devlead An overload for overwriting was added to .NET 8 if that's helpful?
https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.zipfile.extracttodirectory?view=net-8.0#system-io-compression-zipfile-extracttodirectory(system-string-system-string-system-text-encoding-system-boolean)

_Originally posted by @molesmoke in https://github.com/cake-build/cake/issues/1236#issuecomment-1815673026_

devlead:
* fixes #4260


gep13 (approved):
LGTM!

PR #4279, opened 2023-11-24
GH4278: Update Spectre.Console to 0.49.1

### LINKED ISSUE #4278: Update Spectre.Console to  0.49.1

devlead:
* fixes #4278

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        os: [windows-latest, macos-13, ubuntu-latest]
     steps:
       - name: Get the sources
         uses: actions/checkout@v3
--- a/src/Cake.Cli/Cake.Cli.csproj
+++ b/src/Cake.Cli/Cake.Cli.csproj
@@ -19,7 +19,7 @@
 
     <ItemGroup>
       <PackageReference Include="Autofac" Version="8.0.0" />
-      <PackageReference Include="Spectre.Console" Version="0.47.0" />
-      <PackageReference Include="Spectre.Console.Cli" Version="0.47.0" />
+      <PackageReference Include="Spectre.Console" Version="0.49.1" />
+      <PackageReference Include="Spectre.Console.Cli" Version="0.49.1" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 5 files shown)


devlead:
Found the issue Spectre.Console.Cli no longer allows default commands to have a `--version` parameter, which Cake relied on to have it's own custom version command.

https://github.com/spectreconsole/spectre.console/commit/131b37fff858f8d5dd483ff369bb3d43a07c1e08#diff-d38c9fa48d9c39f8eec58957b2333d13943f2cbe1adee26f30325e527a56f600L26-R41

Will see if I can find away around that, can't use the default Spectre version command as it will produce a version string that will break Cake cli contract. Overriding command settings `ApplicationVersion` and removing cake version feature should probably achieve the same result for end user.

devlead:
As it's by design breaking change in Spectre.Console.Cli, I just added a workaround to opt-out of this behavior for Cake.Tool.

src/Cake/Program.cs
```diff
@@ -63,7 +64,14 @@ public async Task<int> Run(string[] args)
...
 
-            return await app.RunAsync(args);
+            return await app.RunAsync(
+                args.Select(
+                    arg => arg switch {
+                        // Opt-out Spectre.Console.Cli no version parameter for default command
+                        "-v" => "--ver",
+                        "--version" => "--ver",
+                        _ => arg
+                    }));
```
devlead, line 74: This is a terrible workaround, should've probably just addressed if first arg is `-v` / `--version`... but the problem goes deeper.

Spectre.Console.Cli switched from
```diff
-                if (firstArgument.Equals("--version", StringComparison.OrdinalIgnoreCase) ||
-                    firstArgument.Equals("-v", StringComparison.OrdinalIgnoreCase))
```
to
```diff
+        if (args.Contains("-v") || args.Contains("--version"))
+        {
```

default command behavior changed from first to any supplied argument is `--version` / `-v` will print the version and exit the application. even i.e.  `--` between command and remaining args used.

so will have to think about this a bit more. moving PR to WIP.

devlead:
The breaking `--version` change was revered in Spectre.Console 0.49.1

devlead:
Ah would seem remaining arguments has changed in Specre.Console.Cli will investigate, PR still WIP.

devlead:
@cake-build/cake-team this should be ready for review now.

patriksvensson (approved):
LGTM 👍

PR #4367, opened 2024-10-16
Task chains

ruslanmogilevskiy:
The discussion is started [here](https://github.com/orgs/cake-build/discussions/4368).

**Problem**:  
Managing build task dependencies in Cake Frosting can become challenging as projects grow, especially when maintaining and modifying task chains later. While existing solutions (IsDependentOn/IsDependeeOf attributes and Cake.Graph) help visualize dependencies, they don't provide a clear, maintainable way to manage the task execution chain.  

**Solution**:  
This PR introduces a Task Chain paradigm that allows developers to:  
- Define task execution order in a clear, sequential manner  
- Organize tasks into logical groups with support for nested grouping  
- Reference tasks by both type and name  
- Maintain compatibility with existing dependency systems

The task chain is enabled with a single line of the configuration code:
```
return new CakeHost()
    ...
	.UseChainedTaskConfigurator<AppTaskChainProvider>()
	.Run(args);
```

**Example usage**:  
```csharp  
class AppTaskChainProvider : ITaskChainProvider
{
    public TaskChainItem GetChain()
    {
        return Chain
            .Task<Task1>()
            .Group("Do something", _ =>
            {
                 // Reference the task type.
                _.Task<Task2>()
                 // Reference the task by name ([TaskName("Task3: do something")]).
                 .Task("Task3: do something");
            })
        // Task chain inline groups example.
        // .Group("Do something else", _ =>
        // {
        //     _.Task<SomeTask>()
        //      .Group("Internal group", _ =>
        //         {
        //             _.Task<OtherTask>();
        //         })
        //      .Task("Task name");
        // })
            // The last task in the chain which the Default task will be dependent on to execute all the chain.
            .Task<FinalTask>();
    }
}
```

This approach significantly improves build script maintainability and readability, making it easier to modify task chains as projects evolve. The implementation requires minimal configuration and can be gradually adopted alongside existing task dependencies

### DIFF
--- a/global.json
+++ b/global.json
@@ -3,7 +3,7 @@
         "src"
     ],
     "sdk": {
-        "version": "8.0.402",
+        "version": "8.0.400",
         "rollForward": "latestFeature"
     }
 }
--- a/src/Cake.Frosting.Example/TaskChains.cs
+++ b/src/Cake.Frosting.Example/TaskChains.cs
@@ -1,0 +1,85 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Core.Diagnostics;
+using Cake.Frosting;
+using Cake.Frosting.TaskChains;
+
+public static class ChainedTasksProgram
+{
+    public static int MainDemo(string[] args)
+    {
+        return new CakeHost()
+            .UseContext<BuildContext>()
+            .UseChainedTaskConfigurator<AppTaskChainProvider>()
+            .Run(args);
+    }
+}
+
+class AppTaskChainProvider : ITaskChainProvider
+{
+    public TaskChainItem GetChain()
+    {
+        return Chain
+            .Task<Task1>()
+            .Group("Do something", _ =>
+            {
+                 // Reference the task type.
+                _.Task<Task2>()
+                 // Reference the task by name ([TaskName("Task3: do something")]).
+                 .Task("Task3: do something");
+            })
+        // Task chain inline groups example.
+        // .Group("Do something else", _ =>
+        // {
+        //     _.Task<SomeTask>()
+        //      .Group("Internal group", _ =>
+        //         {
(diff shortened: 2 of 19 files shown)


ruslanmogilevskiy:

ruslanmogilevskiy:
@devlead , I've completed the PR description.

devlead:
Thanks for taking the time to write this up, for the discussion in https://github.com/orgs/cake-build/discussions/4368, and for the CLA. The pain you describe in Frosting is real: once a build is split across many task classes, the graph lives in `[IsDependentOn]` / `[IsDependeeOf]` attributes and is harder to read than a script Cake file.

We are going to decline merging this into Cake.

Cake’s execution model is a **DAG of named tasks**, not a linear pipeline. This PR does not change that: it generates extra `IsDependentOn` edges from a central list. That is a useful authoring style for some Frosting builds, but it is a niche overlay on the existing model, not something we want as a first-class API in Cake.Core or in the Frosting host.

In particular:

* A top-to-bottom chain encourages treating every task as sequential. Cake is built so you can run any target, fan-in/fan-out, skip with criteria, and use reverse dependencies. We do not want a second “chain” model beside that.
* Groups in this design are not Cake tasks, so they will not show up as stages in `--tree` / `--description`.
* Wiring `"Default"` by name is too magical for the host.
* Mixing chain edges with existing attributes is not safe today: Cake throws if the same dependency is registered twice (`The task already have a dependency on…`). The example also registers a second `[TaskName("Default")]` next to the existing example `DefaultTask`.

What we recommend instead:

* Keep declaring the graph with `[IsDependentOn]` / `[IsDependeeOf]` (or a small orchestrator task that depends on the leaves you care about).
* Use `--tree` / `--description` (and Cake.Graph if you want pictures) to inspect it.
* If you want the fluent chain DSL, that belongs in an **add-in**, not in Cake.Frosting. Cake.Host being sealed makes that awkward; we may consider a small registration hook later, but that would be a separate, much smaller change — not this PR.

Closing as not planned. Happy to keep the discussion thread open if others want to talk about Frosting graph authoring.

PR #4381, opened 2024-10-26
Add .NET 9, C# 13 and remove out of support .NET 6 & 7

### LINKED ISSUE #4345: Add .NET 9 (net9.0) TFM

### LINKED ISSUE #4346: Remove Unsupported TFMS .NET 6 & 7 (net6.0 & net7.0)

### LINKED ISSUE #4378: Update GitVersion next version to 5.0.0
https://github.com/cake-build/cake/blob/948bdc74e7d46c33e7ecb86725df88d99c57ca7c/GitVersion.yml#L1

### LINKED ISSUE #4379: C# 13 Scripting Support
https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-13

### LINKED ISSUE #4380: Update Verify.Xunit to 27.1.0

### LINKED ISSUE #4383: Update LatestPotentialBreakingChange to 5.0.0

### LINKED ISSUE #4384: Remove obsolete members / methods

### LINKED ISSUE #4387: Update NSubstitute to 5.3.0

devlead:
* [x] fixes #4379
* [x] fixes #4345
* [x] fixes #4346
* [x] fixes #4378
* [x] fixes #4380
* [x] fixes #4383 
* [x] fixes #4384
* [x] fixes #4387

Todo #4382 (depends on .NET 9 Release) will come in a separate PR on .NET launch, would be nice to be able to integration test against Azure Artifacts pre-release feed.

### DIFF
--- a/src/Cake.Core/Constants.cs
+++ b/src/Cake.Core/Constants.cs
@@ -11,7 +11,7 @@ namespace Cake.Core
         public const ConsoleColor DefaultConsoleColor = (ConsoleColor)(-1);
 
         public static readonly Version LatestBreakingChange = new Version(0, 26, 0);
-        public static readonly Version LatestPotentialBreakingChange = new Version(4, 0, 0);
+        public static readonly Version LatestPotentialBreakingChange = new Version(5, 0, 0);
 
         public static class Settings
         {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,20 +16,19 @@ jobs:
         os: [windows-latest, macos-13, ubuntu-latest]
     steps:
       - name: Get the sources
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - name: Install .NET SDK 6.0.x - 8.0.x
-        uses: actions/setup-dotnet@v3
+      - name: Install .NET SDK 8.0.x - 9.0.x
+        uses: actions/setup-dotnet@v4
         with:
           dotnet-version: |
-            6.0.x
-            7.0.x
             8.0.x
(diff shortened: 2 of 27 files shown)


src/Cake.Core/Constants.cs
```diff
@@ -11,7 +11,7 @@ internal static class Constants
         public const ConsoleColor DefaultConsoleColor = (ConsoleColor)(-1);
 
         public static readonly Version LatestBreakingChange = new Version(0, 26, 0);
```
patriksvensson, line 13: Not related, but is this used anymore?

patriksvensson (approved):
Looks good to me 👍 Left a comment, but more of a general question that doesn't need to be addressed in this PR.

src/Cake.Core/Constants.cs
devlead: Yes `LatestBreakingChange` will throw an error `LatestPotentialBreakingChange` will log an warning.

patriksvensson: Shouldn't `LatestBreakingChange` be something like 4.0.0 then?

devlead: No, most add-ins compiled for older versions of Cake will still work just fine. Cake NuGet will pick the highest available TFM. If we set 4.0.0, they won't even load, causing support hassle for us for something that might potentially break, hence the warning and recommendation were added a few years ago. We've removed nothing known to break from core interfaces since 0.26.0, but we've added stuff that could potentially break if someone provided their own implementation of core features.

patriksvensson: So couldn't we remove it all together then?

PR #3817, opened 2022-02-17
Add dotnet sln and list commands

megakid:
This adds some additional commands for inspection and manipulation of solution and csproj files using the `dotnet` SDK CLI.

PR #4388, opened 2024-11-03
Add DotNetSlnList alias for dotnet sln list command

Marusyk:
`dotnet sln list ` - Lists all projects in a solution file

The `dotnet sln list` command provides a convenient option to list projects for a given solution.

Docs: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-sln#list

alternative PR for #3817

Please review
Thank you in advance


devlead (approved):
LGTM 👍

PR #4392, opened 2024-11-08
Add DotNetSlnRemove alias for dotnet sln remove command

Marusyk:
`dotnet sln remove` - Removes a project or multiple projects from the solution file.

Docs: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-sln#remove

alternative PR for #3817

Please review
Thank you in advance

### DIFF
--- a/src/Cake.Common/Tools/DotNet/DotNetAliases.cs
+++ b/src/Cake.Common/Tools/DotNet/DotNetAliases.cs
@@ -2942,7 +2942,7 @@ namespace Cake.Common.Tools.DotNet
         [CakeMethodAlias]
         [CakeAliasCategory("Sln")]
         [CakeNamespaceImport("Cake.Common.Tools.DotNet.Sln.Remove")]
-        public static void Remove(this ICakeContext context, FilePath solution, IEnumerable<FilePath> projectPath, DotNetSlnRemoveSettings settings)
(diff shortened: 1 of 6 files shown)


src/Cake.Common/Tools/DotNet/DotNetAliases.cs
```diff
@@ -2883,5 +2884,78 @@ public static DotNetPackageList DotNetListPackage(this ICakeContext context, str
             var lister = new DotNetPackageLister(context.FileSystem, context.Environment, context.ProcessRunner, context.Tools);
             return lister.List(project, settings);
         }
+
+        /// <summary>
```
Marusyk, line 2888: Does it make sense to consider using partial classes for this DotNetAliases to better grouping and navigation?
e.g.  
DotNetAliases.cs
- DotNetAliases.Sln.cs
- DotNetAliases.Package.cs
- ...
What do you think?

devlead: Yes, not a bad idea.

Marusyk: in scope of this PR or create a new one? what do you think?

devlead: separate PR is probably more clear.

devlead (approved):
LGTM 👍

devlead:
@Marusyk your changes have been merged, thanks for your contribution 👍

Issue #4410, opened 2024-12-20
AzurePipelinesPullRequestInfo fails to detect if build is running in the context of a GitHub Pull Request


ap0llo:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake Frosting

### Cake version

5.0.0

### Operating system

Windows

### Operating system architecture

x64

### CI Server

Azure DevOps

### What are you seeing?

When running a build for a GitHub Pull Request on Azure Pipelines, `AzurePipelinesPullRequestInfo.Id` returns `0` and `AzurePipelinesPullRequestInfo.IsPullRequest` returns false even though the corresponding environment variables are set by Azure Pipelines

#### Analysis

I think I already found the root cause for the issue:
 - The Pull Request Id is stored as `int` but the Ids on github,com exceeded `int.MaxValue` some time ago (for example, in one recent PR I opened, the id was `2246475303`)
 - The fix would be to use `long`  instead, but I guess this would be a breaking change
 

#### Workaround

As a workaround for detecting if a build is a PR build, check if `AzurePipelinesPullRequestInfo.Number` is greater than `0`.
This will return the Pull Request number as shown in the GitHub UI, which should fir well within a `int`

### What is expected?

- `AzurePipelinesPullRequestInfo.IsPullRequest` should return true
- `AzurePipelinesPullRequestInfo.Id` should return the id of the pull request (this is different from the PR number)

### Steps to Reproduce

Example of a Cake.Frosting-based build that runs into this can be seen [here](https://dev.azure.com/ap0llo/OSS/_build/results?buildId=6955&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=cf4477f9-109f-5399-1612-bcbe5909c9f2&l=137)

- Set up a Cake.Frosting-based build in a GitHub-hosted repository
- Set up a pipeline on Azure Pipelines
- In the build, access `context.AzurePipelines().Environment.PullRequest.Id` or `context.AzurePipelines().Environment.PullRequest.IsPullRequest`

PR #4411, opened 2024-12-20
Fix detection of current GitHub pull request in Azure Pipelines (#4410)

ap0llo:
Fix detection of GitHub Pull Request builds in Azure Pipelines.  
This fixes the issue described in #4410.

To fix the issue, use a `long` instead of an `int` to store the current Pull Request id (read from the Azure Pipelines variable `SYSTEM_PULLREQUEST_PULLREQUESTID`).

I am aware that is a breaking change that might break all code that accesses `AzurePipelinesPullRequestInfo.Id` - however I am not aware of any other way to potentially fix this,

### DIFF
--- a/src/Cake.Common/Build/AzurePipelines/Data/AzurePipelinesPullRequestInfo.cs
+++ b/src/Cake.Common/Build/AzurePipelines/Data/AzurePipelinesPullRequestInfo.cs
@@ -41,7 +41,7 @@ namespace Cake.Common.Build.AzurePipelines.Data
         /// <value>
         ///   The ID of the pull request that caused this build.
         /// </value>
-        public int Id => GetEnvironmentInteger("SYSTEM_PULLREQUEST_PULLREQUESTID");
+        public long Id => GetEnvironmentLongInteger("SYSTEM_PULLREQUEST_PULLREQUESTID");
 
         /// <summary>
         /// Gets the number of the pull request that caused this build.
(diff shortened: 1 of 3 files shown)


src/Cake.Common/Build/AzurePipelines/Data/AzurePipelinesPullRequestInfo.cs
```diff
@@ -41,7 +41,7 @@ public AzurePipelinesPullRequestInfo(ICakeEnvironment environment)
         /// <value>
         ///   The ID of the pull request that caused this build.
         /// </value>
-        public int Id => GetEnvironmentInteger("SYSTEM_PULLREQUEST_PULLREQUESTID");
+        public long Id => GetEnvironmentLongInteger("SYSTEM_PULLREQUEST_PULLREQUESTID");
```
devlead, line 44: We're not keen on a breaking change and a major new version, so non-breaking suggestion  is to add a new property, obsolete the current with message and no error, and we'll take the breaking change next major version. Not the prettiest but the least disruptive.

```suggestion
        [Obsolete("Type will change in next major version to long, meanwhile use LongId to ensure proper value returned.")]
        public int Id => GetEnvironmentInteger("SYSTEM_PULLREQUEST_PULLREQUESTID");
        
        /// <summary>
        /// Gets the ID of the pull request that caused this build.
        /// This value is set only if the build ran because of a Git PR affected by a branch policy.
        /// </summary>
        /// <value>
        ///   The ID of the pull request that caused this build.
        /// </value>
        public long LongId => GetEnvironmentLongInteger("SYSTEM_PULLREQUEST_PULLREQUESTID");
```
and 
``` diff
-        public bool IsPullRequest => Id > 0;
+         public bool IsPullRequest => LongId > 0;
```

devlead (changes requested):

src/Cake.Common/Build/AzurePipelines/Data/AzurePipelinesPullRequestInfo.cs
ap0llo: Good point, I didn't think of that possibility.

I've adapted the code to match your suggestion

devlead (approved):
LGTM 👍

devlead:
@ap0llo your changes have been merged, thanks for your contribution 👍

PR #4430, opened 2025-01-05
GH4427: Update xunit to V3

### LINKED ISSUE #4427: Update xunit to V3

devlead:
* fixes #4427

Note: Assembly.GetEntryAssembly() now returns Test assembly which meant all Cake.Frosting.Tests tasks now autoregistered, therefor moved its tasks to own assembly Cake.Frosting.Tests.Tasks. Added new Cake.Testing.Xunit.v3 assembly to be used when using Xunit.v3 reusing same files as Cake.Testing.Xunit so we keep functionality intact, but at same time don't break existing users of Cake.Testing.Xunit. Xunit.v3 attributes no longer provide virtual overridable properties so PlatformRestrictedFactAttribute/PlatformRestrictedTheoryAttribute refactored a bit.


patriksvensson (approved):
LGTM 👍

PR #4452, opened 2025-01-17
Update build.ps1 to use TLS v1.3

kapsiR:
Since https://dot.net only supports TLS v1.2+, this drops older TLS/SSL versions too.


devlead (approved):
LGTM 👍

PR #4477, opened 2025-05-25
[WIP] Improve assembly loading and performance

GeertvanHorrik:
For some background info, see https://github.com/orgs/cake-build/discussions/4476

This adds timings and makes sure the correct version of addins are loaded.

Here are some added timings that could be used to find where cake is taking time to initialize:

```
Analyzing build script...
Analyzed build script in 00:00:01.3645522
Processing build script...
Installing tools...
Installed tools in 00:00:09.3698166
Installing addins...
Installed addins in 00:00:01.4150037
```

This is on a fast machine, so installing the tools is the hardest part. A very easy win is to set a shared tools path for the build scripts so it doesn't have to re-install all the tools on each build.

Pinging @gep13 since he was part of the original discussion, but no need for anything just yet.

### DIFF
--- a/GitReleaseManager.yaml
+++ b/GitReleaseManager.yaml
@@ -14,8 +14,10 @@ create:
   include-sha-section: true
   sha-section-heading: "SHA256 Hashes of the release artifacts"
   sha-section-line-format: "- `{1}\t{0}`"
+  include-contributors: true
 close:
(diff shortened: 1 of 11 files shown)


gep13:
Also.... what are the timing differences with this change implemented?  Is there a significant difference with this change in place?

PR #4486, opened 2025-05-26
GH4475: Always show task execution summary

gep13:
Currently, Cake prints out a task execution summary on each successful execution.  This summary includes which tasks were executed or skipped, and the duration of the task.  However, then an execution of Cake fails, the task summary is not printed, and it is necessary to scroll through the Cake output, to find out what task failed.  You also lose the information about which tasks, if any, were skipped.

Any changes to this workflow will have to be done carefully though, as we don't want to alter how tasks are executed, in what order, or what is done in the cases of ContinueOnError, etc.  There are a number of unit tests that check for this execution, and these will need to continue to pass.

This commit aims to address this problem by introducing a new CakeReportException, which has a property called Report of type CakeReport.  That way, when an exception needs to be thrown, the current execution status report can be passed along with it, and from there, the task execution summary can be printed.

The introduction of the CakeReportException has meant that a number of unit tests have had to be updated, since there is an expectation that an InvalidOperationException will be returned, however, I think that this is "ok", since it hasn't meant a change to how the workflow of task executions completes.

### DIFF
--- a/src/Cake.Cli/Hosts/BuildScriptHost.cs
+++ b/src/Cake.Cli/Hosts/BuildScriptHost.cs
@@ -84,14 +84,26 @@ namespace Cake.Cli
 
         private async Task<CakeReport> internalRunTargetAsync()
         {
-            var report = await Engine.RunTargetAsync(_context, _executionStrategy, Settings).ConfigureAwait(false);
-
-            if (report != null && !report.IsEmpty)
+            try
             {
-                _reportPrinter.Write(report);
-            }
+                var report = await Engine.RunTargetAsync(_context, _executionStrategy, Settings).ConfigureAwait(false);
 
-            return report;
+                if (report != null && !report.IsEmpty)
+                {
+                    _reportPrinter.Write(report);
+                }
+
+                return report;
+            }
+            catch (CakeReportException cre)
+            {
+                if (cre.Report != null && !cre.Report.IsEmpty)
+                {
+                    _reportPrinter.Write(cre.Report);
+                }
+
+                throw;
+            }
         }
     }
 }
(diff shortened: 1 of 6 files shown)


gep13:
The output of these changes can be seen here:

image

For both the CakeSpectreReportPrinter and the CakeReportPrinter.

gep13:
When attempting to solve this problem, I tried to:

* simply remove the `throw` [here](https://github.com/cake-build/cake/blob/f2a316dfde9a46df205ad554169d13c4f8cd3a39/src/Cake.Core/CakeEngine.cs#L208), but that introduced a number of failing tests, which led me to think it was a bad idea
* pass the `_reportPrinter` through in the call to `RunTargetAsync` [here](https://github.com/cake-build/cake/blob/f2a316dfde9a46df205ad554169d13c4f8cd3a39/src/Cake.Cli/Hosts/BuildScriptHost.cs#L87), but that was a fairly serious breaking change, that I didn't want to continue with

While this is technically a breaking change, since the exception type that is returned is different, the original exception type is still available to be found, so I personally think that this is "ok".

Happy to hear any other ideas about what can be done here.

devlead:
> @patriksvensson no objections to the possible breaking change here? Or do you not consider it a break?

Internal bits so think it's fine.

devlead (approved):
LGTM 👍

patriksvensson:
No, changing the exception type (if I understand correctly) might theoretically be a breaking change but I can't see how this will impact anyone.

devlead:
Missed last comment after enabled automerge.

Agree CakeReportException should inherit CakeException.
@gep13 Please follow up with a PR when possible.

gep13:
@devlead happy to follow up with another PR, but I am unsure about unsealing `CakeException`.

PR #4515, opened 2025-06-05
Expand short path for addin installations

GeertvanHorrik:
Extra required fix that should have been part of #4500


GeertvanHorrik:
@devlead while testing the fix for #4500 as prerelease, I noticed this one was missing.

PR #4516, opened 2025-06-05
Add ResolveShortPath for more tools / addins path resolving

GeertvanHorrik:
See #4500 and #4515 for previous work.


GeertvanHorrik:
Verified the fix on the machine that failed, it's working after these changes.

Issue #4517, opened 2025-06-06
Cake Frosting WithCriteria Description


tstewart65:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake Frosting

### Cake version

5.0

### Operating system

Windows

### Operating system architecture

x64

### What are you seeing?

The fix implemented in #4027 does not seem to work as implemented.  The SkippedMessage property was added to IFrostingTask and it was also implemented there to return an empty string.  IFrostingTask is implemented in FrostingTask but it has nothing in there to allow you to change the SkippedMessage property.  This is the only level where you can change the SkippedMessage property and have it picked up when trying to reference it using IFrostingTask.  I added a SkippedMessage property to my CleanTask I was creating and the message was not used when the following is called in FrostingEngine

```
// Is the criteria method overridden?
if (task.IsShouldRunOverridden(_context))
{
      cakeTask.WithCriteria(task.ShouldRun, task.SkippedMessage);
}

```

I made a small test project to verify this.  Unless I am missing something I think this needs changed.

### What is expected?

Want to be able to set the skipped message for the WithCriteria like you can when using Cake Tool.

### Steps to Reproduce

Just try to set the SkippedMessage so it can be used when using the ShouldRun override. 

### Output log

No error log to show the message does not show up.

devlead:
Confirmed: `SkippedMessage` as a default interface member is not visible through `IFrostingTask`, so #4027 never actually lets Frosting tasks set a skip message. That part of the report is correct.

Obsoleting `ShouldRun` is out of scope even for 7.0 — please keep it. The rewrite we want is on #4518: make `SkippedMessage` overridable on the task bases, add extra criteria *beside* `ShouldRun` (not instead of it), and avoid a required interface member.

Details: https://github.com/cake-build/cake/pull/4518#pullrequestreview-5263919121

PR #4518, opened 2025-06-17
GH4517: - Cake Frosting WithCriteria Description

tstewart65:
Marked the current ShouldRun method obsolete.
Added a new property to IFrostingTask, ShouldRunCriteria which is a List of CakeTaskCriteria.  This allows you to have multiple criteria just like what could be done in Cake.Tool.

The current method of the ShouldRun method that you could override and the SkippedMessage has shortcoming when compared to Cake.Tool.  First the SkippedMessage could not be set the way it is currently implemented and it only allowed for one message for the task being skipped.  It could not be changed if you had multiple criteria to check for skipping a task. 

The new ShouldRunCriteria property allows you to setup multiple criteria and have a different message for each one just like Cake.Tool.

This would be nice to have in 5.1 to allow me to continue converting a current cake recipe here at work over to Cake.Frosting.  The one downfall is since Cake.Frosting does not support the Spectre Console output the skipped messages are not shown in the summary at this time.

### DIFF
--- a/src/Cake.Frosting/AsyncFrostingTask.cs
+++ b/src/Cake.Frosting/AsyncFrostingTask.cs
@@ -26,7 +26,7 @@ namespace Cake.Frosting
         where T : ICakeContext
     {
         /// <inheritdoc/>
-        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new();
+        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new List<CakeTaskCriteria>();
 
         /// <summary>
         /// Runs the task using the specified context.
--- a/src/Cake.Frosting/AsyncFrostingTask.cs
+++ b/src/Cake.Frosting/AsyncFrostingTask.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Cake.Core;
 
@@ -24,6 +25,9 @@ namespace Cake.Frosting
     public abstract class AsyncFrostingTask<T> : IFrostingTask
         where T : ICakeContext
     {
+        /// <inheritdoc/>
+        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new();
+
         /// <summary>
         /// Runs the task using the specified context.
         /// </summary>
@@ -41,6 +45,7 @@ namespace Cake.Frosting
         /// <returns>
         ///   <c>true</c> if the task should run; otherwise <c>false</c>.
         /// </returns>
+        [Obsolete("ShouldRun method will be removed in a future version")]
         public virtual bool ShouldRun(T context)
         {
             return true;
@@ -79,6 +84,7 @@ namespace Cake.Frosting
         }
 
         /// <inheritdoc/>
+        [Obsolete("ShouldRun method will be removed in a future version")]
         bool IFrostingTask.ShouldRun(ICakeContext context)
         {
             if (context is null)
--- a/src/Cake.Frosting/FrostingTask.cs
+++ b/src/Cake.Frosting/FrostingTask.cs
@@ -26,7 +26,7 @@ namespace Cake.Frosting
         where T : ICakeContext
     {
         /// <inheritdoc/>
-        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new();
+        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new List<CakeTaskCriteria>();
 
         /// <summary>
         /// Runs the task using the specified context.
--- a/src/Cake.Frosting/FrostingTask.cs
+++ b/src/Cake.Frosting/FrostingTask.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Cake.Core;
 
@@ -24,6 +25,9 @@ namespace Cake.Frosting
     public abstract class FrostingTask<T> : IFrostingTask
         where T : ICakeContext
     {
+        /// <inheritdoc/>
+        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new();
+
         /// <summary>
         /// Runs the task using the specified context.
         /// </summary>
--- a/src/Cake.Frosting/IFrostingTask.cs
+++ b/src/Cake.Frosting/IFrostingTask.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Cake.Core;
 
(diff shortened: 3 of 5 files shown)


gep13:
As discussed offline via email, we will not be able to proceed with this PR (as it is currently written), until the next major release of Cake, since it contains breaking changes.

gep13:
I am adding the `Breaking Change` label here (normally this would be applied to the issue, but putting it on the PR for now, as a reminder about what is going on here.

src/Cake.Frosting/IFrostingTask.cs
```diff
@@ -13,6 +14,11 @@ namespace Cake.Frosting
     /// </summary>
     public interface IFrostingTask
     {
+        /// <summary>
+        /// Gets ShouldRunCriteria for this task.
+        /// </summary>
+        List<CakeTaskCriteria> ShouldRunCriteria { get; }
```
devlead, line 20: Please do not add a **required** `ShouldRunCriteria` member (no default implementation). Existing custom `IFrostingTask` implementers would fail to compile, and existing binaries would miss the member.

Prefer keeping extra criteria off the interface (store them on the task bases) or a DIM that returns an empty enumerable so this stays additive.

src/Cake.Frosting/IFrostingTask.cs
```diff
@@ -27,6 +33,7 @@ public interface IFrostingTask
         /// <returns>
         ///   <c>true</c> if the task should run; otherwise <c>false</c>.
         /// </returns>
+        [Obsolete("ShouldRun method will be removed in a future version")]
```
devlead, line 36: Please drop `[Obsolete]` on `ShouldRun`. This is the documented Frosting criteria API and we are not removing it in 7.0 (or as part of this work). Obsoleting it is too disruptive — CS0618 everywhere, and warnings-as-errors CI would fail — for a change that is not required to fix #4517 or to add per-criteria messages.

src/Cake.Frosting/FrostingTask.cs
```diff
@@ -24,6 +25,9 @@ public abstract class FrostingTask : FrostingTask<ICakeContext>
     public abstract class FrostingTask<T> : IFrostingTask
         where T : ICakeContext
     {
+        /// <inheritdoc/>
+        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new List<CakeTaskCriteria>();
```
devlead, line 29: An expression-bodied getter that returns `new List<CakeTaskCriteria>()` allocates a **new list on every access**. Callers cannot `ShouldRunCriteria.Add(...)` in a constructor; those items disappear on the next get.

Prefer `protected void WithCriteria(Func<T, bool> predicate, string message = null)` into a private field, or a field initialized once. Also please add `virtual string SkippedMessage` here (and explicit `IFrostingTask` forwarding) — that is the actual #4517 fix, and this PR never wires it, so existing `ShouldRun` users still cannot set a message through the interface.

src/Cake.Frosting/AsyncFrostingTask.cs
```diff
@@ -24,6 +25,9 @@ public abstract class AsyncFrostingTask : AsyncFrostingTask<ICakeContext>
     public abstract class AsyncFrostingTask<T> : IFrostingTask
         where T : ICakeContext
     {
+        /// <inheritdoc/>
+        public virtual List<CakeTaskCriteria> ShouldRunCriteria => new List<CakeTaskCriteria>();
```
devlead, line 29: Same as `FrostingTask<T>`: `=> new List<...>()` is a new list every get, so ctor `.Add` is lost. Prefer a private field / `protected WithCriteria(...)`, plus `virtual SkippedMessage` forwarded on the explicit interface implementation.

src/Cake.Frosting/AsyncFrostingTask.cs
```diff
@@ -41,6 +45,7 @@ public virtual Task RunAsync(T context)
         /// <returns>
         ///   <c>true</c> if the task should run; otherwise <c>false</c>.
         /// </returns>
+        [Obsolete("ShouldRun method will be removed in a future version")]
```
devlead, line 48: Please remove `[Obsolete]` from `ShouldRun` (and from the explicit `IFrostingTask.ShouldRun` implementation below). Keep the method; it is not going away in 7.0.

src/Cake.Frosting/Internal/FrostingEngine.cs
```diff
@@ -125,11 +126,21 @@ private void ConfigureTasks()
...
+                            cakeTask.WithCriteria(criteria.Predicate, criteria.Message);
+                        }
+                    }
+                    // Check to see if obsoleted ShouldRun is overridden?  Done to not break old projects.
+#pragma warning disable CS0618 // Type or member is obsolete
+                    else if (task.IsShouldRunOverridden(_context))
                     {
                         cakeTask.WithCriteria(task.ShouldRun, task.SkippedMessage);
                     }
+#pragma warning restore CS0618 // Type or member is obsolete
```
devlead, line 143: Please register **both**, not if/else. If `ShouldRun` is overridden, always `WithCriteria(task.ShouldRun, task.SkippedMessage)`, then register each extra criterion. Combining an override with per-criteria messages should work; extra criteria must not hide `ShouldRun`.

src/Cake.Frosting/Internal/Extensions/FrostingTaskExtensions.cs
```diff
@@ -61,6 +61,7 @@ public static bool IsRunOverridden(this IFrostingTask task, IFrostingContext con
             throw new InvalidOperationException($"This method expects all {nameof(IFrostingTask)} to be instances of {nameof(FrostingTask)} or {nameof(AsyncFrostingTask)}.");
         }
 
+        [Obsolete("ShouldRun method will be removed in a future version")]
```
devlead, line 64: Please remove `[Obsolete]` here. `IsShouldRunOverridden` should stay as a normal helper while `ShouldRun` remains the public API.

devlead (changes requested):
Thanks for tracking this down — #4517 is real. `SkippedMessage` lives as a default interface member, `FrostingEngine` reads it via `IFrostingTask`, and a property on a task class does not implement that member, so the message is always empty. Per-criteria messages (Cake.Tool `WithCriteria` parity) are also a reasonable addition.

We still cannot take this PR as written for 7.0.

**Please do not obsolete `ShouldRun`.** That is the documented Frosting criteria API. `[Obsolete]` would warn (and fail warnings-as-errors) across the ecosystem for something we are not removing. Keep `ShouldRun` and `SkippedMessage`; add extra criteria **beside** them, not instead of them.

A 7.0-friendly rewrite:
1. Add `virtual string SkippedMessage` on `FrostingTask<T>` / `AsyncFrostingTask<T>` and forward the explicit `IFrostingTask` implementation (same pattern as `ShouldRun`). That fixes #4517 for existing tasks.
2. Add optional extra criteria (preferred: `protected void WithCriteria(Func<T, bool> predicate, string message = null)` on the bases, or a DIM returning an empty enumerable). Register those **and** the overridden `ShouldRun` — not if/else.
3. Do not add a **required** `ShouldRunCriteria` member on `IFrostingTask` (no DIM). That is a source/binary break for custom implementers and is not needed.
4. If a list property remains, do not use `=> new List<...>()` on the getter (new list every access; ctor `.Add` is lost).
5. Tests for skip message, multiple criteria with distinct messages, and `ShouldRun` combined with extra criteria. Frosting tests currently have no coverage here.

Frosting summaries still use `CakeReportPrinter`, which does not print skip reasons (Spectre does on Cake.Tool). That is optional follow-up, not a reason to drop messages — `CakeEngine` already uses them in “could not reach target … skipped due to a criteria”.

PR #4519, opened 2025-06-20
(maint) Fixed formatting of several files

gep13:
Corrected the usage of the following blocks:

* if
* using
* foreach

Once seen, these could not be _unseen_!


devlead (approved):
LGTM

Issue #4474, opened 2025-05-22
Task Dependees don't show up using the `tree` command


CDTR-MattConroy:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake Frosting

### Cake version

5.0.0

### Operating system

N/A

### Operating system architecture

N/A

### What are you seeing?

When running the `tree` command, only Tasks registered as Dependent on others show up. Tasks registered as Dependees are missing.

### What is expected?

The `tree` command shows the entire tree of tasks regardless of how they were registered.

### Steps to Reproduce

1. Create a Frosting project
2. Create 3 Tasks, A,B,C
3. Register Task A as dependent on Task B
4. Register Task B as dependee of Task C
5. Run `--tree`

Mohammad1384a:
Hey guys, I had two questions about the issue and then i will get started working on it
1. what should be the expected output? like in A,B,C case should it be like
A |-B |-C
or A |- B
B|-c
2. can a task have multiple dependent tasks? like A has B and C directly depending on it and D depends on B. in case that's possible i think the second option for printing the tasks would make more sense. like we start from the top and first print the directly dependent ones and on the next line we go for the next level

Mohammad1384a:
@devlead  @CDTR-MattConroy

CDTR-MattConroy:
In the example I gave originally, A is dependent on B, and B is a dependee of C, which is to say C is dependent on B. So if I want to run A or C, I must run B first.
So if I understand correctly how `--tree` currently works I would expect the following:
```
A
└─B
C
└─B
```

I don't think the logic for how the `--tree` command works needs to change, it just seems to be not discovering dependents correctly either when:

- using `DependeeOf`
- tasks are from a different assembly
- tasks are from a package

Or a combination of the above.

Mohammad1384a:
@CDTR-MattConroy created a pr please take a look
https://github.com/cake-build/cake/pull/4663

PR #4663, opened 2025-11-21
Fix Frosting --tree output to include tasks connected via IsDependeeOf

Mohammad1384a:
Fix Frosting --tree output so it includes tasks related via IsDependeeOf.
TreeScriptHost previously traversed only task.Dependencies, ignoring relationships encoded via Dependees in the graph.
Updated to use the full graph connections from CakeGraphBuilder, ensuring the printed tree matches the actual dependency graph.

### DIFF
--- a/src/Cake.Cli/Hosts/TreeScriptHost.cs
+++ b/src/Cake.Cli/Hosts/TreeScriptHost.cs
@@ -9,14 +9,13 @@ using System.Threading.Tasks;
 using Cake.Core;
 using Cake.Core.Graph;
 using Cake.Core.Scripting;
+using ThreadingTask = System.Threading.Tasks.Task;
 
-namespace Cake.Cli
-{
+namespace Cake.Cli {
     /// <summary>
     /// The script host used for showing task descriptions.
     /// </summary>
-    public sealed class TreeScriptHost : ScriptHost
-    {
+    public sealed class TreeScriptHost : ScriptHost {
         private const int _maxDepth = 0;
         private const string _cross = "├─";
         private const string _corner = "└─";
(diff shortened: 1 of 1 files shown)


CDTR-MattConroy:
Thanks for looking into this issue! 

I don't know what the conventions for this project are, but it seems like there are a lot of formatting and style changes here that are otherwise unnecessary making it difficult to see what has actually changed logic wise. Could the styling be kept consistent with what was already there?

gep13:
@Mohammad1384a thank you for your interest in this project, and also for taking the time to provide this PR.

Before we will be able to review this properly, we would require you to make some changes.

1. Get the changes to the point that the build succeeds
2. Provide some testing steps with regard to how this work can be verified
3. As @CDTR-MattConroy has pointed out, update the changes to adhere to the existing formatting of this project

Let us know if you have any problems.

src/Cake.Cli/Hosts/TreeScriptHost.cs
```diff
@@ -31,102 +30,110 @@ public sealed class TreeScriptHost : ScriptHost
         /// <param name="console">The console.</param>
         public TreeScriptHost(ICakeEngine engine, ICakeContext context, IConsole console)
             : base(engine, context)
-        {
+            {
```
devlead, line 33: Please restore Allman braces and indentation to match `DescriptionScriptHost` / `DryRunScriptHost` (`namespace Cake.Cli` on its own line, constructor body aligned with `base(...)`, braces on their own lines). Formatting-only commits still leave a noisy, non-conforming diff that hides the real logic change.

src/Cake.Cli/Hosts/TreeScriptHost.cs
```diff
@@ -31,102 +30,110 @@ public sealed class TreeScriptHost : ScriptHost
...
         }
 
         /// <inheritdoc/>
-        public override Task<CakeReport> RunTargetAsync(string target)
-        {
+        public override Task<CakeReport> RunTargetAsync(string target) {
             PrintTaskTree();
-
-            return System.Threading.Tasks.Task.FromResult<CakeReport>(null);
+            return ThreadingTask.FromResult<CakeReport>(null);
```
devlead, line 40: This `ThreadingTask` alias is unnecessary. This file already imports `System.Threading.Tasks` and the other script hosts use `System.Threading.Tasks.Task.FromResult<CakeReport>(null)` directly. Please drop the alias as part of restoring the original style.

src/Cake.Cli/Hosts/TreeScriptHost.cs
```diff
@@ -31,102 +30,110 @@ public sealed class TreeScriptHost : ScriptHost
...
         }
 
-        private List<ICakeTaskInfo> GetTopLevelTasks()
+        private List<ICakeTaskInfo> GetTopLevelTasks(CakeGraph graph)
         {
             // Display "Default" first, then alphabetical
-            var graph = CakeGraphBuilder.Build(Tasks);
-            return Tasks.Where(task => !graph.Edges.Any(
-                edge => edge.Start.Equals(task.Name, StringComparison.OrdinalIgnoreCase)))
+            // Top-level = tasks that never appear as Start (no outgoing edges)
```
devlead, line 68: "No outgoing edges" is easy to misread. In `CakeGraph`, `Start` is the prerequisite and `End` is the dependent (`Connect(dependency, dependent)`). Top-level tasks are those that never appear as `Start` (they are never a prerequisite of another task). A short comment along those lines would match `CakeGraph.Traverse`.

src/Cake.Cli/Hosts/TreeScriptHost.cs
```diff
@@ -31,102 +30,110 @@ public sealed class TreeScriptHost : ScriptHost
...
+            // Children = all tasks that have an edge Start -> End = current task.
+            // This respects both IsDependentOn and IsDependeeOf,
+            // because CakeGraphBuilder already encoded both into the graph.
+            var childTasks = graph.Edges
+                .Where(edge => edge.End.Equals(task.Name, StringComparison.OrdinalIgnoreCase))
+                .Select(edge => Tasks.First(t =>
+                    t.Name.Equals(edge.Start, StringComparison.OrdinalIgnoreCase)))
+                .Distinct()
+                .OrderBy(t => t.Name, StringComparer.OrdinalIgnoreCase)
+                .ToList();
```
devlead, line 109: This is the correct orientation: `Start` is the prerequisite, so children of the current task are edges with `End == task.Name`. Please keep this — it is what includes `IsDependeeOf` — and add a test so it cannot regress (issue #4474 A/B/C, and/or Frosting `CleanTask` + `DependeeTask` with `--tree`).

Optional: children are now ordered by name rather than `Dependencies` declaration order. That is acceptable for `--tree` if you mention it.

devlead (changes requested):
Thanks for taking this on — using `CakeGraphBuilder` for children is the right fix. `IsDependeeOf` is stored on the other task and only shows up as a graph edge, so walking `task.Dependencies` was why those nodes disappeared from `--tree` (Cake and Frosting share `TreeScriptHost`). That child lookup should stay.

This is not merge-ready yet. The remaining asks from earlier review still apply:

1. **Formatting** — Please restore Allman braces and indentation to match the other hosts in `Cake.Cli` (`DescriptionScriptHost`, `DryRunScriptHost`). The style-only follow-up commits still leave a noisy, non-conforming diff (K&R braces, mis-indented constructor / `PrintTask` / `PrintName` bodies, unnecessary `ThreadingTask` alias). Keep the logic change and drop unrelated formatting so the real diff is reviewable.
2. **Tests and verification** — Please add coverage and a short note on how to verify:
   - Cake.Cli unit test for the issue A/B/C case (`A` depends on `B`, `B` is dependee of `C`) expecting:
     ```
     A
     └─B
     C
     └─B
     ```
   - Frosting `--tree` with the existing `CleanTask` / `DependeeTask` pair (or `dotnet run -- --tree` in `src/Cake.Frosting.Example`, which should list `Dependee` under `World`).
3. **CI / CLA / rebase** — Builds are currently red. Please sign the CLA (the bot is still pending), rebase onto current `develop`, and get CI green.

Minor: children are now sorted by name instead of `Dependencies` declaration order. That is fine for `--tree` if you call it out. I am not asking for cycle detection; that is pre-existing (`Traverse` already handles it at run time).

PR #4666, opened 2025-11-28
GH4520: add slnx support to SolutionParser

### LINKED ISSUE #4520: add slnx support
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake .NET Tool

### Cake version

latest

### Operating system

N/A

N/A

### CI Server

_No response_

### What are you seeing?

slnx is an XML SLN format.
.net9 added slnx support `dotnet new sln -f slnx` and latest VS opens and builds such solution.

`src/Cake.Common/Tools/DotNet/Sln` is missing slnx.

### What is expected?

slnx parsing and serialization using the same library dotnet-sdk is using https://github.com/microsoft/vs-solutionpersistence.

### Steps to Reproduce

N/A

### Output log

_No response_

wgnf:
Closes #4520

As discussed in the above-mentioned issue I have added support for `.slnx` Solution files without any third-party library.

I have added the exact same tests as for the `.sln` files, using `dotnet sln migrate` to migrate the used `.sln` files to `.slnx` and keeping the assertions. That way, I could make sure that there aren't any regressions.  
  
There are few things to point out:
- The new `.slnx` format does not provide any IDs anymore, as cross-references between things in the Solution are not necessary anymore. The `SolutionProject` still has a mandatory ID however (and I didnt't want to make breaking changes to that). So I just submitted `string.Empty` to the ID
- `.sln` and `.slnx` files support nested folders. In `.slnx` files these are modeled using the name of a solution-folder (i.e. `Root` is the root folder, and `Root/Sub` is the folder "Sub" underneath that `Root`-Folder). To not make things too complicated for now, I did not adhere to that consideration just yet (for the above-mentioned example, the "Root" `SolutionFolder` would not have an instance of the "Sub" `SolutionFolder`). If that is something that you want to see, I can make sure to respect that
- I have unfortunately no idea if there are any documentation hints that I would need to extend to make sure that users know that `.slnx` files are now also supported

If something is unclear here in the PR or in the code, please feel free to reach out :)


wgnf:

wgnf:
Seems like the pipelines are failing however. Is that something I have to worry about?

gep13:
> @wgnf said...
Seems like the pipelines are failing however. Is that something I have to worry about?

Yes, this is absolutely something that we need to worry about 😄 

Looks like your changes are failing one of the StyleCop rules that we have in place.

It could be that some newer syntax that you are using is causing a rule to be tripped.

wgnf:
Ahh, I guess the `[]` is making an issue. I can go ahead and change that.

kasperk81:
@wgnf, you can run it through this test suite https://github.com/microsoft/vs-solutionpersistence/tree/main/test/Microsoft.VisualStudio.SolutionPersistence.Tests/SlnAssets to iron out corner cases. i.e. `serializeToXml(deserialize(<name>.sln.txt)) == <name>.slnx.xml`

As someone who added slnx support in `dotnet new sln -f slnx`, on github linguist for ` ```slnx ` and `.slnx` syntax highlighting, then participated in whole msbuild and sdk integration process, I can tell you with confidence that there are many corner cases handled in their parser; which is why i was rooting for using official parser that was rejected by the maintainer

wgnf:
@kasperk81 

Maybe the full-fledged support using the third-party library you mentioned, supporting all the edge-cases there might be, is something that can be done in a separate AddIn, as already mentioned in the Issue.

I guess, but I might be alone on this, simple `.sln` (that is already there) and `.slnx`, that does not support _all_ the edge-cases, is ok for the core-library, as it adds the functionality for most use-cases.

wgnf:
The current pipeline error is the following:

```
/home/runner/work/cake/cake/src/Cake.Frosting.Template/Cake.Frosting.Template.csproj(25,4): error MSB3733: Input file "templates/cakefrosting/build/Build.csproj" cannot be opened. The process cannot access the file '/home/runner/work/cake/cake/src/Cake.Frosting.Template/templates/cakefrosting/build/Build.csproj' because it is being used by another process. [TargetFramework=net8.0]
```

Seems like some race-condition, but not something caused by my changes, to me?

devlead:
> The current pipeline error is the following:
> 
> ```
> /home/runner/work/cake/cake/src/Cake.Frosting.Template/Cake.Frosting.Template.csproj(25,4): error MSB3733: Input file "templates/cakefrosting/build/Build.csproj" cannot be opened. The process cannot access the file '/home/runner/work/cake/cake/src/Cake.Frosting.Template/templates/cakefrosting/build/Build.csproj' because it is being used by another process. [TargetFramework=net8.0]
> ```
> 
> Seems like some race-condition, but not something caused by my changes, to me?

Yes, likely a file in use; I've seen the .NET 10 SDK holding on to files a bit more at times.

kasperk81:
> that does not support _all_ the edge-cases, is ok for the core-library, as it adds the functionality for most use-cases.

I'm not sure what you meant by "all the edge cases". Did you find concrete results from that test suite like x/N cases are not handled such that those x case are unlikely to happen in reality and hence considered "edge cases"?

wgnf:
No, I have not used that suite yet.

devlead:
> No, I have not used that suite yet.

They're likely be edge cases, but add a couple integration tests here
https://github.com/cake-build/cake/blob/develop/tests/integration/Cake.Common/Solution/SolutionAliases.cake

Test solutions are added and copied from here
https://github.com/cake-build/cake/tree/develop/tests/integration/resources/Cake.Common/Solution

And we then we can always add more tests if a new edge case or actual issue should arise.
But if `dotnet new`,  Cake slnx and the unit tests ones work, then we've got a good starting point.

kasperk81:
Something not in [Slnx.xsd schema](https://github.com/microsoft/vs-solutionpersistence/blob/main/src/Microsoft.VisualStudio.SolutionPersistence/Serializer/Xml/Slnx.xsd) is considered edge case? or what is the definition of edge case here?
 
We should at least gather the data what is not going to be supported and write it down in unit test in terms of: what is the expected behavior of cake when such content is provided that is understood by msbuild/vscode but rejected by cake.

devlead:
> Something not in [Slnx.xsd schema](https://github.com/microsoft/vs-solutionpersistence/blob/main/src/Microsoft.VisualStudio.SolutionPersistence/Serializer/Xml/Slnx.xsd) is considered edge case? or what is the definition of edge case here?
>  
> We should at least gather the data what is not going to be supported and write it down in unit test in terms of: what is the expected behavior of cake when such content is provided that is understood by msbuild/vscode but rejected by cake.

To be clear, with edge cases, we mean things not needed, known or tested at this point. 

We've not reviewed PR yet, but if it handles similar scenarios as existing sln parser, it'll likely be a good starting point. To fully handle everything string possible in format today or in the future is out of scope. For that we recommend creating an addin with official parser(or just reference it). 100% spec compliance is out of scope, if other common needs come up we can add more test scenarios and sort those.

wgnf:
I just added two simple integration tests for `.slnx` files. I wanted to do something more complex too, but then I realized:

- The `SolutionProjectEqualityComparer` that is used for the solution-file integration-tests also checks for the parent of a project (see [here](https://github.com/cake-build/cake/blob/develop/tests/integration/utilities/comparers/SolutionProjectEqualityComparer.cake#L19)], which I would expect and want
- but the `Parent` property of the `SolutionProject` is `internal set;` (see [here](https://github.com/cake-build/cake/blob/develop/src/Cake.Common/Solution/SolutionProject.cs#L37)), which makes it impossible to set up an "expected" `SolutionProject` in the integration test, when dealing with solution folders

And this is something I don't feel comfortable changing.

Or am I missing something that I could to, to be able to set up an integration-test considering these restrictions?

devlead:
@wgnf you can add completely new tests that you see appropriate, important part is we test the alias works as expected with real files and cake script context.

wgnf:
Yeah, I've already gone ahead and added two integration tests

src/Cake.Common/Solution/SolutionParser.cs
```diff
@@ -182,5 +195,104 @@ private static void ParseNestedProjectLine(List<SolutionProject> projects, strin
             parent.Items.Add(child);
             child.Parent = parent;
         }
+
+        private static SolutionParserResult ParseSlnxSolution(IFile file)
+        {
+            var xmlDocument = new XmlDocument();
+            using var xmlContentStream = file.OpenRead();
+            xmlDocument.Load(xmlContentStream);
```
devlead: To reduce Malicious XML potential XXE attacks, we can disable external schema/file processing, something like.
```suggestion
            var xmlReaderSettings = new XmlReaderSettings
            {
                DtdProcessing = DtdProcessing.Prohibit, // Prevent XXE
                XmlResolver = null // Disable external entity resolution
            };

            using var xmlContentStream = file.OpenRead();
            var xmlDocument = new XmlDocument();
            xmlDocument.Load(xmlContentStream);
            using var xmlReader = XmlReader.Create(xmlContentStream, xmlReaderSettings);
            xmlDocument.Load(xmlReader);
```

src/Cake.Common/Solution/SolutionParser.cs
```diff
@@ -182,5 +195,104 @@ private static void ParseNestedProjectLine(List<SolutionProject> projects, strin
...
+
+            var projects = new List<SolutionProject>();
+            foreach (var xmlElement in root.ChildNodes.OfType<XmlElement>())
+            {
+                var projectsFromElement = ParseSlnxElement(xmlElement, file);
+                projects.AddRange(projectsFromElement);
+            }
+
+            // the new SLNX has no information about the file format version, VS version and minimal VS version, so it's omitted.
+            return new SolutionParserResult(string.Empty, string.Empty, string.Empty, projects);
```
devlead: ParseSlnSolution uses projects.AsReadOnly() (line 125), but ParseSlnxSolution passes a mutable list. This breaks immutability and can potentially cause runtime issues.
```suggestion
            return new SolutionParserResult(string.Empty, string.Empty, string.Empty, projects.AsReadOnly());
```

src/Cake.Common/Solution/SolutionParser.cs
```diff
@@ -182,5 +195,104 @@ private static void ParseNestedProjectLine(List<SolutionProject> projects, strin
...
+                // the new project type id notation does not quite fit the known format, so it is adjusted.
+                : $"{{{projectTypeId.ToUpper()}}}";
+
+            var projectFilePath = new FilePath(projectPath);
+            var projectFileFullPath = projectFilePath.IsRelative ?
+                solutionFile.Path.GetDirectory().CombineWithFilePath(projectFilePath) :
+                projectFilePath;
+
+            // the project name is part of its path.
+            var actualProjectName = projectPath.Split('/', StringSplitOptions.RemoveEmptyEntries).Last().Replace(".csproj", string.Empty);
```
devlead: We could already created file path here.
```suggestion
            var actualProjectName = projectFilePath.GetFilenameWithoutExtension().FullPath;
```

src/Cake.Common/Solution/SolutionParser.cs
```diff
@@ -182,5 +195,104 @@ private static void ParseNestedProjectLine(List<SolutionProject> projects, strin
...
+
+            // elements in the SLNX file don't have any IDs anymore, so it's omitted.
+            var solutionProject = new SolutionProject(string.Empty, actualProjectName, projectFileFullPath, projectTypeId);
+            return new List<SolutionProject> { solutionProject };
```
devlead: Just a note, new collection syntax can be used, current stylecop rules not aware of new feature, but formatted acording to current rules would be
```suggestion
            return
                [
                    solutionProject
                ];
```

src/Cake.Common.Tests/Unit/Solution/SolutionParserTests.cs
```diff
@@ -150,5 +151,145 @@ public void Should_Properly_Parse_Projects_With_Absolute_Path()
...
+
+                // When
+                var result = solutionParser.Parse(slnFilePath);
+
+                // Then
+                Assert.NotNull(result);
+                Assert.NotNull(result.Projects);
+                Assert.Equal(5, result.Projects.Count);
+                var onlyProjects = result.Projects.Where(x => !(x is SolutionFolder)).ToList();
+                Assert.Equal(3, onlyProjects.Count);
```
devlead: We've started to use Verify for simpler and more complete asserts, currently first in Cake.Core.Tests, but feel this would be a good addition to add it to Cake.Common.Tests, this is optional, I could do the leg work to sort this.

```suggestion
                await Verify(result);
```

devlead (changes requested):
Quick initial look, and this looks good to me, some minor comments. Will take a deeper look as soon as I can.

wgnf:
Thanks for the fast feedback! I think I'll be able to adress the requests later this week or on the weekend.

devlead:
> Thanks for the fast feedback! I think I'll be able to adress the requests later this week or on the weekend.

Excellent, let us know if any questions arise.

src/Cake.Common/Solution/SolutionParser.cs
wgnf: Thank you. I have addressed this with my latest changes. I have omitted the additional `xmlDocument.Load(xmlContentStream);` though, as it is redundant.

src/Cake.Common/Solution/SolutionParser.cs
wgnf: Good point, thanks. This slipped through. Addressed this with my latest changes.

src/Cake.Common/Solution/SolutionParser.cs
wgnf: Addressed this with my latest changes.

src/Cake.Common/Solution/SolutionParser.cs
wgnf: Ah okay. I used this layout with the new changes.

src/Cake.Common.Tests/Unit/Solution/SolutionParserTests.cs
wgnf: I just started to have a look at this. I have not used this before, so I am not quite sure what steps I need to take to enable this. But I'll give it a try :)

devlead: You can see example packages added here
https://github.com/cake-build/cake/blob/develop/src/Cake.Core.Tests/Cake.Core.Tests.csproj#L18-L20

Cake specific settings you set here
https://github.com/cake-build/cake/blob/develop/src/Cake.Core.Tests/VerifyConfig.cs

Believe VerifyDiffPlex.Initialize can only be called  once per solution, so that can be skipped. 

Cake solution doesn't use implicit using, so you need an

using static VerifyXunit.Verifier;



First time you execute tests it'll fail, launch a VSCode diff with snapshot, if correct just copy correct to other side, close and rerun tests.

wgnf: I have migrated the tests for `SlnParser` to use the `Verify` logic you mentioned. Was pretty straight-forward actually. Though I have no idea if there are some details I should look out for. Looks good to me though and tests fail if I change the parsing logic in any way.

wgnf: I just had to rearrange the initialization logic, because:

- When running all tests it would fail with "already initialized"
- But, when running a single test, tests in a test class, in a single test project, it would fail with "needs to be initialized"

So I basically made sure in each `[ModuleInitializer]` to only call `VerifyDiffPlex.Initialize()` when it has not been initialized already (`VerifyDiffPlex.Initialized`).

Let me know if this aligns with the way you would want to see things done.

wgnf:
I integrated the requested changes. The Azure Pipeline has problems with a hanging process again however :/

devlead (approved):
LGMT 👍

devlead:
@wgnf your changes have been merged, thanks for your contribution 👍

Issue #4660, opened 2025-11-18
Consider adding IFileSystemInfo returning APIs in addition to IFile and IDirectory


paulomorgado:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake Frosting

### Cake version

6.0.0

### Operating system

N/A

### Operating system architecture

x64

### What are you seeing?

APIs like `IFileSystem`, `IDirectory` and `GlobberExtensions` have APIs that return either `IFile` or `IDirectory`, but, some times, we just want a file system entry and decide what to do with it based in whether it's a file or directory.

### What is expected?

`context.FileSystem.GetFileSystemInfo(path)`
`context.Globber.GetFileSystemInfos(pattern)`
`directory.GetFileSystemInfos()`

### Steps to Reproduce

Try getting a list of file system infos that contains both files and directories,

Abde1rahman1:
I'm interested in working on this feature. Is there anyone already assigned to it ?

PR #4704, opened 2026-01-11
GH-4660: Implement GetFileSystemInfo in FileSystem, Directory and Glo…

Abde1rahman1:
## Description
This PR addresses issue #4660 by providing a way to retrieve file system entries (files and directories) without needing to know the type beforehand.

## Changes
- Added `GetFileSystemInfo` to `IFileSystem`.
- Added `GetFileSystemInfos` to `IDirectory` and `IGlobber`.
- Implemented changes in `FileSystem`, `Directory`, `Globber` and their `Fake` counterparts.


Abde1rahman1:

Abde1rahman1:
Any Updates here ?!

devlead:
> Any Updates here ?!

Thank you for your contribution and effort here! I see that this PR introduces GetFileSystemInfo on IFileSystem and related methods on IDirectory/IGlobber.

However, adding members directly to these core interfaces without a default implementation is a breaking change, which we'll need to consider carefully, as it may require a major release, as it potentially breaks ABI and any existing external dependency i.e., Cake addins.

There are also no tests verifying these changes (Cake.Core Unit tests are located https://github.com/cake-build/cake/tree/develop/src/Cake.Core.Tests, and integration tests are here). Which means we'll need to allocate time and resources to carefully review this PR; we will, but can't guarantee when.

src/Cake.Core/IO/IFileSystem.cs
```diff
@@ -22,5 +22,12 @@ public interface IFileSystem
         /// <param name="path">The path.</param>
         /// <returns>A <see cref="IDirectory"/> instance representing the specified path.</returns>
         IDirectory GetDirectory(DirectoryPath path);
+
+        /// <summary>
+        /// Gets a <see cref="IFileSystemInfo"/> for the specified path.
+        /// </summary>
+        /// <param name="path">The path.</param>
+        /// <returns>A <see cref="IFileSystemInfo"/> for the specified path.</returns>
+        IFileSystemInfo GetFileSystemInfo(DirectoryPath path);
```
devlead, line 31: `DirectoryPath` cannot represent a file, and there is no `FilePath` → `DirectoryPath` conversion, so `GetFileSystemInfo(someFilePath)` will not compile. Issue #4660 needs a path that might be either type.

Please take `Path` (with `FilePath` / `DirectoryPath` overloads if implicit `string` conversion is ambiguous). Also give this member a **default interface implementation** (probe `GetDirectory` then `GetFile`) so existing `IFileSystem` implementers / addins are not a breaking change.

src/Cake.Core/IO/IDirectory.cs
```diff
@@ -100,5 +100,13 @@ public interface IDirectory : IFileSystemInfo
...
         /// <returns>The <see cref="IDirectory"/> instance representing the specified path.</returns>
         IDirectory SetUnixFileMode(UnixFileMode unixFileMode) => this;
+
+        /// <summary>
+        /// Gets all file system entries in the directory.
+        /// </summary>
+        /// <param name="filter">The filter.</param>
+        /// <param name="scope">The search scope.</param>
+        /// <returns>The file system entries.</returns>
+        IEnumerable<IFileSystemInfo> GetFileSystemInfos(string filter, SearchScope scope);
```
devlead, line 110: Adding this without a default implementation is a breaking change for every `IDirectory` implementer.

Please add a DIM, e.g. concatenate `GetDirectories(filter, scope)` and `GetFiles(filter, scope)`, matching how `SetCreationTime` / `SetUnixFileMode` were added. Place it next to `GetFiles` / `GetDirectories` rather than after the timestamp setters. With a DIM, `FakeDirectory` does not need its own copy unless we want different behavior.

src/Cake.Core/IO/IGlobber.cs
```diff
@@ -20,5 +20,12 @@ public interface IGlobber
         ///   <see cref="Path" /> instances matching the specified pattern.
         /// </returns>
         IEnumerable<Path> Match(GlobPattern pattern, GlobberSettings settings);
+
+        /// <summary>
+        /// Returns a list of <see cref="IFileSystemInfo"/> matching the specified pattern.
+        /// </summary>
+        /// <param name="pattern">The pattern to match.</param>
+        /// <returns>A list of <see cref="IFileSystemInfo"/>.</returns>
+        IEnumerable<IFileSystemInfo> GetFileSystemInfos(string pattern);
```
devlead, line 29: Please do not add this to `IGlobber`. `GetFiles` / `GetDirectories` live on `GlobberExtensions` and Cake.Common `GlobbingAliases` so we do not break globber implementers.

A better fit is an extension (and optionally `GetFileSystemInfos` on `ICakeContext`) that reuses `Match(GlobPattern, GlobberSettings)` and maps `FilePath` / `DirectoryPath` through `IFileSystem`. That also gives settings/predicates and avoids a parallel globber implementation.

src/Cake.Core/IO/Globber.cs
```diff
@@ -52,5 +54,21 @@ public IEnumerable<Path> Match(GlobPattern pattern, GlobberSettings settings)
...
+                throw new ArgumentNullException(nameof(pattern));
+            }
+
+            var glob = _parser.Parse(pattern, new GlobberSettings
+            {
+                IsCaseSensitive = !_comparer.IsCaseSensitive
+            });
+            var settings = new GlobberSettings { IsCaseSensitive = !_comparer.IsCaseSensitive };
+            return _visitor.Walk(glob, settings);
+        }
```
devlead, line 72: This diverges from `Match` in ways that will surprise callers:

- `IsCaseSensitive = !_comparer.IsCaseSensitive` **inverts** the platform default (`Match` leaves it unset so the parser uses `environment.Platform.IsUnix()`). On Windows this becomes case-sensitive; on Unix, case-insensitive.
- Empty/whitespace patterns: `Match` returns an empty sequence; here `Parse` can `NullReferenceException` on `CurrentToken`.
- No `.Distinct(_comparer)` after `Walk`, so duplicate paths are possible.
- `string` instead of `GlobPattern`, and no `GlobberSettings`.
- `_fileSystem` is stored but unused.

Please reuse `Match` (or the same parse/walk/distinct/empty handling) instead of a second globber path. Cover this in unit tests and in `tests/integration/Cake.Common/IO/GlobbingAliases.cake` (existing `GetFiles`/`GetDirectories`/`GetPaths` plus new mixed-entry tasks).

src/Cake.Core/IO/FileSystem.cs
```diff
@@ -20,5 +22,20 @@ public IDirectory GetDirectory(DirectoryPath path)
...
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+            var directory = GetDirectory(path);
+            if (directory.Exists)
+            {
+                return directory;
+            }
+            return GetFile(path.FullPath);
+        }
```
devlead, line 39: Same API issue as `IFileSystem`: the caller must pass a `DirectoryPath`, so this cannot look up a known `FilePath` without an awkward `new DirectoryPath(file.FullPath)`.

If the interface uses a default implementation, this override may not be needed. If we keep an explicit implementation, probe by existence for both files and directories, and document the non-existent-path convention (return `IFile` with `Exists == false`).

src/Cake.Testing/FakeDirectory.cs
```diff
@@ -276,5 +277,15 @@ private static Regex CreateRegex(string pattern)
             pattern = pattern.Replace(".", "\\.").Replace("*", ".*").Replace("?", ".{1}");
             return new Regex(pattern, RegexOptions.Singleline | RegexOptions.Compiled);
         }
+
+        /// <inheritdoc/>
+        public IEnumerable<IFileSystemInfo> GetFileSystemInfos(string filter, SearchScope scope)
+        {
+            // بنجمع المجلدات والملفات الوهمية
```
devlead, line 284: Please drop this comment (repo style is English XML docs / no narrative comments here). If `IDirectory.GetFileSystemInfos` has a default implementation, this override can be removed entirely.

If it stays, add Verify tests next to the existing `GetFiles` / `GetDirectories` cases in `src/Cake.Testing.Tests`.

devlead (changes requested):
Thank you for picking up [#4660](https://github.com/cake-build/cake/issues/4660) — the direction is right (mixed file/directory entries without the caller knowing the type up front).

This still needs another pass before we can merge. Adding members directly to `IFileSystem`, `IDirectory`, and `IGlobber` without default implementations is a **breaking ABI change** for addins and fakes (already labeled). Cake already uses default interface methods on `IDirectory` for newer members; please follow that, and keep globbing helpers on extensions/aliases like `GetFiles` / `GetDirectories`.

There are also correctness bugs in `Globber.GetFileSystemInfos` (inverted case sensitivity vs `Match`, empty pattern `NullReferenceException`, missing `Distinct`), `GetFileSystemInfo` takes `DirectoryPath` so a file path cannot be passed, and there are **no tests**.

### Requested changes

1. **Avoid a breaking change**
   - `IDirectory.GetFileSystemInfos`: default implementation concatenating existing `GetDirectories` + `GetFiles` (place it next to those methods).
   - `IFileSystem.GetFileSystemInfo`: default implementation that probes directory then file (or overloads for `FilePath` / `DirectoryPath` / `Path`).
   - Do **not** add a member to `IGlobber`. Put `GetFileSystemInfos` on `GlobberExtensions` (and a Cake.Common alias if we want `GetFileSystemInfos(pattern)` in scripts), mapping `Match` results through `IFileSystem`.

2. **`GetFileSystemInfo` must accept a file or a directory**
   - Issue #4660 is specifically “we don’t know the type.” `DirectoryPath` only cannot express a `FilePath` (no conversion). Prefer `Path` plus overloads if string conversion is ambiguous. If the path exists as a directory, return `IDirectory`; otherwise return `IFile` (including non-existent — please document that).

3. **Globber must match `Match` / `GetFiles`**
   - Reuse `Match` (or the same empty-pattern, parse, walk, `Distinct(_comparer)`, and default case-sensitivity handling). Do not invert `IsCaseSensitive`. Take `GlobPattern` + `GlobberSettings` like the rest of the globber API. Drop the unused `_fileSystem` field if it is not needed.

4. **Tests** (currently none)
   - **Unit:** `src/Cake.Core.Tests` — `GetFileSystemInfo` (file exists, directory exists, neither, null); globber beside `GlobberTests` (empty pattern, case sensitivity vs `Match`, mixed files+dirs, settings/predicates).
   - **Fake:** `src/Cake.Testing.Tests` Verify tests next to existing `GetFiles` / `GetDirectories` cases.
   - **Integration (required):** `tests/integration` is how we prove `.cake` scripts still compile and run. Existing `GetFiles` / `GetDirectories` / `GetPaths` tasks in `tests/integration/Cake.Common/IO/GlobbingAliases.cake` must still pass. Also add script tests for `Context.FileSystem.GetFileSystemInfo`, `directory.GetFileSystemInfos`, and `Context.Globber.GetFileSystemInfos` / a `GetFileSystemInfos` alias, wired into the same aggregator tasks in `tests/integration/build.cake`.

Nit: please remove the non-English comment in `FakeDirectory.GetFileSystemInfos`.

Abde1rahman1:
Thanks @devlead  for the detailed review — I've pushed a new commit addressing all of it.
  Avoided the breaking change
  - IDirectory.GetFileSystemInfos is now a default interface implementation next to GetDirectories/GetFiles, concatenating the two.
  - IFileSystem.GetFileSystemInfo is now a default implementation that probes directory-then-file.
  - Removed GetFileSystemInfos from IGlobber entirely. It now lives on GlobberExtensions (mapping Match results through
    IFileSystem.GetDirectory/GetFile), plus a GetFileSystemInfos alias on ICakeContext in Cake.Common so it's usable from scripts.

  GetFileSystemInfo now accepts file or directory
  - Signature changed to take Path (so both FilePath and DirectoryPath work). It returns IDirectory if the path exists as a directory,
    otherwise IFile — including for non-existent paths, which is documented on the member.

  Globber correctness
  - GetFileSystemInfos (on GlobberExtensions now) reuses Match directly instead of re-implementing parse/walk, so it inherits the same
    empty-pattern handling, Distinct(_comparer), and default case sensitivity — the inverted IsCaseSensitive bug is gone.
  - Dropped the unused _fileSystem field from Globber.

  Tests
  - Unit: added FileSystemTests covering GetFileSystemInfo (file, directory, missing, null path), and a TheGetFileSystemInfosMethod block
    in GlobberTests covering empty pattern, case sensitivity vs Match, mixed files+directories, and settings/predicates.
  - Fake: updated FakeFileSystemTests.FakeDirectory and removed the bespoke FakeDirectory/FakeFileSystem overrides in favor of the default    interface implementations (so the fakes now exercise the same code path as real usage).
  - Integration: added GetFileSystemInfo/GetFileSystemInfos tasks to tests/integration/Cake.Common/IO/GlobbingAliases.cake, wired into the    Cake.Common.IO.GlobbingAliases aggregator, covering Context.FileSystem.GetFileSystemInfo, directory.GetFileSystemInfos, and the
    Context.Globber/alias path. Existing GetFiles/GetDirectories/GetPaths tasks are untouched and still pass.

  Nit
  - Removed the non-English comment in FakeDirectory (the method itself is gone now, replaced by the default interface implementation).

src/Cake.Core/IO/IFileSystem.cs
```diff
@@ -22,5 +24,26 @@ public interface IFileSystem
...
+
+        /// <summary>
+        /// Gets a <see cref="IFileSystemInfo"/> for the specified path.
+        /// </summary>
+        /// <param name="path">The path.</param>
+        /// <returns>
+        /// A <see cref="IDirectory"/> when the path exists as a directory; otherwise an <see cref="IFile"/>,
+        /// including when the path does not exist.
+        /// </returns>
+        IFileSystemInfo GetFileSystemInfo(Path path)
```
devlead: This is the remaining blocker. FilePath and DirectoryPath both convert from string, so GetFileSystemInfo("./foo") does not compile (ambiguous conversion to Path). Issue #4660 is specifically “we don’t know the type,” so a string/script call has to work.

Please add a string DIM overload that does the same directory-then-file probe (and a test that passes a string). The Path overload can stay. FilePath/DirectoryPath overloads alone will not fix this.

Nit: prefer ArgumentNullException.ThrowIfNull(path) to match the rest of Cake.Core.

src/Cake.Core/IO/Directory.cs
```diff
@@ -121,5 +121,12 @@ public IDirectory SetUnixFileMode(UnixFileMode unixFileMode)
             _directory.Refresh();
             return this;
         }
+
+        public IEnumerable<IFileSystemInfo> GetFileSystemInfos(string filter, SearchScope scope)
```
devlead: This duplicates IDirectory’s default implementation. Please remove it so FakeDirectory and the physical Directory both use the DIM (single behavior).

src/Cake.Core/IO/GlobberExtensions.cs
```diff
@@ -37,6 +37,41 @@ public static IEnumerable<DirectoryPath> GetDirectories(this IGlobber globber, G
...
+
+        /// <summary>
+        /// Gets all files and directories matching the specified pattern.
+        /// </summary>
+        /// <param name="globber">The globber.</param>
+        /// <param name="fileSystem">The file system.</param>
+        /// <param name="pattern">The pattern.</param>
+        /// <param name="settings">The globber settings.</param>
+        /// <returns>The file system entries matching the specified pattern.</returns>
+        public static IEnumerable<IFileSystemInfo> GetFileSystemInfos(this IGlobber globber, IFileSystem fileSystem, GlobPattern pattern, GlobberSettings settings)
```
devlead: Reusing Match here is correct — please keep that.

The extra IFileSystem parameter is fine on Cake.Core (IGlobber has no file system). It does mean context.Globber.GetFileSystemInfos(pattern) from #4660 still does not exist. The Cake.Common alias GetFileSystemInfos(pattern) is the script/Frosting entry point; please say that in the XML docs (and an <example> like GetFiles) so callers are not sent to the two-arg extension.

src/Cake.Core.Tests/Unit/IO/FileSystemTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    public sealed class FileSystemTests
+    {
+        [Fact]
+        public void Should_Return_File_When_Path_Is_A_File()
+        {
+            var fakeFileSystem = new FakeFileSystem(FakeEnvironment.CreateUnixEnvironment());
+            IFileSystem fileSystem = fakeFileSystem;
+            var expected = fakeFileSystem.CreateFile("/test/file.txt");
+
+            var result = fileSystem.GetFileSystemInfo(new FilePath("/test/file.txt"));
```
devlead: Please add a case that calls GetFileSystemInfo with a string (file exists, directory exists, missing). Also add the license header used by sibling tests in this folder.

src/Cake.Core/IO/IGlobber.cs
```diff
@@ -20,5 +20,6 @@ public interface IGlobber
         ///   <see cref="Path" /> instances matching the specified pattern.
         /// </returns>
         IEnumerable<Path> Match(GlobPattern pattern, GlobberSettings settings);
+
```
devlead: Whitespace-only change — please revert this file (and the same extra blank / missing EOF newline on FileSystem.cs, FakeDirectory.cs, FakeFileSystem.cs).

devlead (changes requested):
Thank you for the follow-up — the DIM / extensions / tests pass addresses most of the previous review.

`IDirectory.GetFileSystemInfos` and `IFileSystem.GetFileSystemInfo` are now default interface methods, `IGlobber` is unchanged, globbing goes through `GlobberExtensions` + the Cake.Common alias, `Match` is reused, and there are unit, Fake Verify, and integration tests. That is the right shape for a non-breaking add.

This still is not merge-ready because `GetFileSystemInfo(Path)` cannot be called with a string. `FilePath` and `DirectoryPath` both have implicit `string` conversions, so `context.FileSystem.GetFileSystemInfo("./foo")` is ambiguous — which is exactly the [#4660](https://github.com/cake-build/cake/issues/4660) case (the caller does not know the type). Typed `FilePath`/`DirectoryPath` tests do not cover that.

### Requested changes

1. Add a `GetFileSystemInfo(string path)` default implementation (same directory-then-file probe). Keep the `Path` overload. Optional `FilePath`/`DirectoryPath` overloads do not fix strings by themselves. Add a unit test and an integration assertion that pass a string.
2. Delete the copy of `GetFileSystemInfos` on internal `Directory` so the interface default is the only implementation (same as `FakeDirectory`).
3. Revert whitespace-only changes (`IGlobber.cs`, `FileSystem.cs`, `FakeDirectory.cs`, `FakeFileSystem.cs`, extra blanks / missing EOF newlines).
4. Update the PR description (it still says members were added to `IGlobber` and the fakes).

Nits: license header + Given/When/Then on `FileSystemTests`; `ArgumentNullException.ThrowIfNull` on the DIM; an `<example>` on the alias like `GetFiles`. Script callers should use `GetFileSystemInfos(pattern)`; Core `IGlobber` should not grow a member.

The **Breaking change** label can come off once the string overload exists. Please do not add anything back onto `IGlobber`.

Abde1rahman1:
Thanks for the detailed review, @devlead  All requested changes are addressed:

  - **String overload:** added `GetFileSystemInfo(string path)` as a default interface method with the same directory-then-file probe. The `Path` overload is kept and  delegates to it, so `context.FileSystem.GetFileSystemInfo("./foo")` now compiles.
  - **Tests:** added unit tests that pass a string (file exists, directory exists, missing, null), plus integration assertions in `GlobbingAliases.cake` that use
  string paths.
  - **Single implementation:** removed the `GetFileSystemInfos` copy from the internal `Directory`, so `Directory` and `FakeDirectory` both use the `IDirectory`
  default.
  - **Whitespace:** reverted `IGlobber.cs`, `FileSystem.cs`, `FakeDirectory.cs`, and `FakeFileSystem.cs`. They're no longer in the diff.
  - **`IGlobber`:** unchanged. Nothing was added back.
  - **Nits:** license header and Given/When/Then on `FileSystemTests`, `ArgumentNullException.ThrowIfNull` in the default methods, and `<example>` blocks on the
  `GetFileSystemInfos` aliases. The Core `GlobberExtensions` docs now point script and Frosting callers to the `GetFileSystemInfos(pattern)` alias.
  - **PR description:** updated to match the current implementation.

  Since the string overload is in, could the "Breaking change" label be removed? Let me know if anything else needs adjusting.

devlead (approved):
LGTM👍 Will merge once CI passes.

devlead:
@Abde1rahman1 your changes have been merged, thanks for your contribution 👍

Abde1rahman1:
Thanks a lot, @devlead Really enjoyed contributing to Cake and learned a lot from your review. Looking forward to making more contributions!

Issue #4292, opened 2024-02-21
Bug with InspectCodeSettings parsing in InspectCodeRunner on Linux (and probably MacOS)


lmb-djaquier:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/cake-build/cake/issues) to ensure it has not already been reported

### Cake runner

Cake Frosting

### Cake version

4.0.0

### Operating system

Linux, macOS

### Operating system architecture

64-Bit

### CI Server

Teamcity

### What are you seeing?

Parameters are starting with a / (see output logs) . With a / they are not considered as parameters but as paths, so the output of the InspectCode.sh script is the error "Specify only one solution file".

Regarding the [sourcecode of the runner](https://github.com/cake-build/cake/blob/911a2f4ed8ba36a6a63758564956883f884bcc33/src/Cake.Common/Tools/InspectCode/InspectCodeRunner.cs#L213), that bug impacts all parameters, except --no-build

### What is expected?

Parameters must start with a - or -- (like already implemented for the --no-build parameter but not the others) on Linux and MacOS operating systems

### Steps to Reproduce

Configure your inspectcode in cake with some parameters that way:

```
context.InspectCode(context.PathToSolution, new InspectCodeSettings
{
    OutputFile = "output.json",
    Debug = true,
    Build = false,
    SolutionWideAnalysis = false,
    Severity = InspectCodeSeverity.Warning,
    Verbosity = InspectCodeVerbosity.Trace
}
.WithToolPath("<toolpath>/JetBrains.ReSharper.CommandLineTools.2023.3.3/tools/inspectcode.sh")
);
```

To make it work, currently i used that workaround:

```
context.InspectCode(context.PathToSolution, new InspectCodeSettings
{
    Build = false,
    ArgumentCustomization = args => args.Append("--swea --severity=WARNING --output=output.json"),
}
.WithToolPath("<toolpath>/JetBrains.ReSharper.CommandLineTools.2023.3.3/tools/inspectcode.sh")
);
```

### Output log

##teamcity[blockOpened name='Code Inspection']
Executing: /source/solution/build/tools/JetBrains.ReSharper.CommandLineTools.2023.3.3/tools/inspectcode.sh "/output:/source/solution/output/reports/inspectCode.xml"  /debug "/verbosity=TRACE" "/severity=WARNING" --no-build "/source/solution/Jsnow.sln"
JetBrains Inspect Code 2023.3.3
[...]

PR #4760, opened 2026-03-12
Use inspectcode.sh on non-Windows platforms (#4313)

### LINKED ISSUE #4313: `InspectCode` calls the exe file even on Linux, where it doesn't work
### Prerequisites

### Cake runner

Cake .NET Tool, Cake Frosting

### Cake version

4.0.0

### Operating system

Linux

64-Bit

### CI Server

_No response_

### What are you seeing?

I would like to run `InspectCode` on my solution. I have installed the [ReSharper Command Line Tools](https://www.jetbrains.com/resharper/features/command-line.html), as they contain the same code as the NuGet package (but can be installed globally once). On Linux, they need a custom runtime, therefore should be started from `inspectcode.sh` and not `inspectcode.exe`.
Therefore, I get the error
```
Failed to fill the environment container with handler ?{myArgs=?(2pcs)[/output:/output.xml, solution.sln]?}?. Could not find method '.ctor' due to a type load error: Could not load file or assembly 'PresentationCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. assembly:/opt/resharper/JetBrains.Platform.Util.dll type:ListEvents`1 member:(null)
```

Unfortunately, `inspectcode.exe` is hardcoded in Cake. I would like to see a switch based on the current OS.

### What is expected?

If I run `InspectCode` on a Linux machine, `inspectcode.sh` is used instead of `inspectcode.exe`.

### Steps to Reproduce

Call `InspectCode` on a Linux machine

### Output log

See above for the error message

1PetarVujakovic:
Fixes #4313

On Linux/macOS installations of ReSharper Command Line Tools the entry point is
"inspectcode.sh", but Cake currently attempts to execute "inspectcode.exe".

This change prefers `inspectcode.sh` on non-Windows platforms while keeping
".exe" as fallback to preserve compatibility with existing setups and tests.

### DIFF
--- a/src/Cake.Common/Tools/InspectCode/InspectCodeRunner.cs
+++ b/src/Cake.Common/Tools/InspectCode/InspectCodeRunner.cs
@@ -225,7 +225,17 @@ namespace Cake.Common.Tools.InspectCode
         /// <returns>The tool executable name.</returns>
         protected override IEnumerable<string> GetToolExecutableNames(InspectCodeSettings settings)
         {
-            return new[] { settings != null && settings.UseX86Tool ? "inspectcode.x86.exe" : "inspectcode.exe" };
+            if (_environment.Platform.Family == PlatformFamily.Windows)
+            {
+                return new[] { settings != null && settings.UseX86Tool ? "inspectcode.x86.exe" : "inspectcode.exe" };
+            }
+
+            if (settings != null && settings.UseX86Tool)
+            {
+                return new[] { "inspectcode.x86.exe", "inspectcode.sh", "inspectcode.exe" };
+            }
+
+            return new[] { "inspectcode.sh", "inspectcode.exe" };
         }
 
         /// <summary>


1PetarVujakovic:

devlead:
I'm wondering would the "correct" way be to implement support for the [JetBrains.ReSharper.GlobalTools](https://www.jetbrains.com/help/resharper/ReSharper_Command_Line_Tools.html#install-and-use-resharper-command-line-tools-as-net-core-toolsl) and support the `jb inspectcode` command mentioned in their docs.

Did a quick test, and it looks like it should work.
```cake
// Cake.Tool
#tool dotnet:?package=JetBrains.ReSharper.GlobalTools&version=2025.3.3

// Cake.Sdk
InstallTool("dotnet:?package=JetBrains.ReSharper.GlobalTools&version=2025.3.3");


var projectPath = GetFiles("./src/**/*.csproj")
                .FirstOrDefault() ?? throw new FileNotFoundException("Failed to find csproj");

Command(
    ["jb", "jb.exe"], // jb non-Windows, jb.exe Windows
    new ProcessArgumentBuilder()
        .Append("inspectcode")
        .AppendQuoted(projectPath.FullPath)
        .Append("--stdout")
);
```

Arguments might need to be prefixed `inspectcode` as `jb` can also launch `cleanupcode`.

src/Cake.Common/Tools/InspectCode/InspectCodeRunner.cs
```diff
@@ -225,7 +225,17 @@ protected override string GetToolName()
...
         {
-            return new[] { settings != null && settings.UseX86Tool ? "inspectcode.x86.exe" : "inspectcode.exe" };
+            if (_environment.Platform.Family == PlatformFamily.Windows)
+            {
+                return new[] { settings != null && settings.UseX86Tool ? "inspectcode.x86.exe" : "inspectcode.exe" };
+            }
+
+            if (settings != null && settings.UseX86Tool)
+            {
+                return new[] { "inspectcode.x86.exe", "inspectcode.sh", "inspectcode.exe" };
```
devlead, line 235: `UseX86Tool` putting `inspectcode.x86.exe` first on non-Windows is a footgun: that PE will not run on Linux/macOS. Ignore x86 off Windows (same list as the default Unix path: `inspectcode.sh` then `inspectcode.exe`).

Please also add Unix fixture tests: create both `inspectcode.sh` and `inspectcode.exe` and assert the resolved path is `.sh`. Today `Should_Find_Inspect_Code_Runner` still asserts `/Working/tools/inspectcode.exe` because `ToolFixture` only seeds the fallback.

devlead (changes requested):
Thanks for taking #4313 — preferring `inspectcode.sh` on non-Windows matches JetBrains Command Line Tools and is the right direction for the executable name.

This is not ready to merge as-is:

1. **Linux still fails after this change** — `GetArguments` still emits `/output:`, `/swea`, `/verbosity=`, etc. On Unix InspectCode treats those as extra solution paths (`Specify only one solution file`). That is [#4292](https://github.com/cake-build/cake/issues/4292) and was also called out on #4313. Switching the launcher without `--` flags (or an explicit deferral linking #4292) is not a complete Linux fix.
2. **Design choice is still open** — please respond to the [`jb inspectcode` / `JetBrains.ReSharper.GlobalTools` question](https://github.com/cake-build/cake/pull/4760#issuecomment-2920000000). Aliases still document `#tool "nuget:?package=JetBrains.ReSharper.CommandLineTools"`. We need an explicit product choice: extend this runner for CLT `.sh`, add `jb`/`jb.exe` resolution, or both.
3. **The new resolution path is untested** — existing tests still assert `inspectcode.exe` because the Unix fixture only creates the `.exe` fallback. Please add coverage that Unix prefers `.sh` when both files exist.

Also rebase/merge `develop`; this branch is quite stale. `DupFinderRunner` still only looks for `dupfinder.exe` — fine to leave out of this PR, but worth a follow-up if we stay on CLT scripts.

Please either emit `--` flags on non-Windows here, or state in the PR that #4292 is required follow-up before claiming Linux support.

Issue #3245, opened 2021-03-08
Debug/Diagnostic color is far too dark


gitfool:
While investigating issues with https://github.com/cake-build/cake/pull/3240, I temporarily enabled `Debug/Diagnostic` logging for the integration tests, during which the log rendering is far too dark and becomes unreadable in some environments.

GitHub Actions run [2052071902](https://github.com/cake-build/cake/runs/2052071902?check_suite_focus=true#step:8:1383):
![image](https://user-images.githubusercontent.com/750121/110259924-eda76480-800e-11eb-9b2d-56597c9f01f7.png)

Azure Pipelines build [17082](https://dev.azure.com/cake-build/Cake/_build/results?buildId=17082&view=logs&j=00af5505-930d-50dd-3844-b167e8b4bb13&t=0fadbdea-3d1d-5635-586e-134b382e6b33&l=1380):
![image](https://user-images.githubusercontent.com/750121/110260114-cef59d80-800f-11eb-88b6-36a125702451.png)

Here's another screenshot of a normal GitHub Actions run showing the color spectrum for different `LogLevel/Verbosity`:
![image](https://user-images.githubusercontent.com/750121/110260406-f731cc00-8010-11eb-9091-61c82e95b7cb.png)

And the same for a normal Azure Pipelines build:
![image](https://user-images.githubusercontent.com/750121/110260531-8c34c500-8011-11eb-9fc5-9365f0772fbc.png)

Finally, I've seen similar problems when building local in WSL.

augustoproiete:
Agreed. I think we need [custom console color palettes](Support custom ConsolePalette/ANSI colors) for different CI systems.

TeamCity has the opposite problem (not dark enough) having a white background, so parts of the log is not easily readable.

![image](https://user-images.githubusercontent.com/177608/110260792-0fa7d500-7f84-11eb-8e96-ac8690d9df2e.png)

(selected text)
![image](https://user-images.githubusercontent.com/177608/110260799-1cc4c400-7f84-11eb-8ac1-c6f37f893f0d.png)

gitfool:
One more for good measure... reading the log output with [Visual Studio Code](https://github.com/Microsoft/vscode), in dark mode, with the [ANSI Colors](https://github.com/iliazeus/vscode-ansi) extension:

![image](https://user-images.githubusercontent.com/750121/110261448-26e2d300-8015-11eb-8110-e0164ae23818.png)

PR #4799, opened 2026-05-17
fix(logging): use light-background palette on TeamCity

wyf027:
## Summary

- When `TEAMCITY_VERSION` is set, use a console palette tuned for TeamCity's light build log background (darker foreground colors, especially for Information-level output).

## Motivation

Relates to #3245 / #2976 — GitHub Actions/Azure Pipelines need readable output on dark logs; TeamCity needs the opposite on its white log background.

## Test plan

- [x] Added `CakeBuildLogTests` coverage for TeamCity ANSI and system-console colorization
- [ ] `dotnet test src/Cake.Core.Tests/Cake.Core.Tests.csproj --filter FullyQualifiedName~TeamCity`


devlead:

```
```

devlead:

PR #4805, opened 2026-05-17
fix: preserve folder structure when glob matches only subdirectories (#2102)

### LINKED ISSUE #2102: CopyFiles does not respect folder structure when pattern only matches files in a subdirectory
### What You Are Seeing?
CopyFiles does not correctly respect the folder structure when a pattern is used which only matches files in a subdirectory. 
### What is Expected?
If I have the files:
./src/file1.txt
./src/sub/file2.txt
./src/sub/file3.dat

The command CopyFiles("./src/**/*.txt", "./dest", true) will result in:
./dest/file1.txt
./dest/sub/file2.txt

However the command CopyFiles("./src/**/*.dat", "./dest", true) will result in:
./dest/file3.dat

I would expect it to result in:
./dest/sub/file3.dat
0.26.1
64bit
Windows
No
This repo demonstrates the issue:
https://github.com/dandezille/cake-copyfiles
### Output Log
https://gist.github.com/dandezille/f34f8f0c0fc8429be8fe85397c588856

wyf027:
## Summary

- When `CopyFiles` is called with `preserveFolderStructure: true` and a glob pattern, derive the structure base directory from the pattern prefix (text before the first `*`, `?`, or `{`).
- Use that base (e.g. `src` from `src/**/*.dat`) instead of inferring a common path only from matched file paths, which failed when every match lived in a subdirectory.

## Test plan

- [x] `Should_Preserve_Subdirectory_Structure_When_Only_Subdirectory_Files_Match_Glob` — single `src/sub/file3.dat` via `./src/**/*.dat` lands at `dst/sub/file3.dat`
- [ ] Existing `FileCopierTests` (CI)

CI not run locally (no .NET SDK in contributor environment).

Fixes #2102

### DIFF
--- a/src/Cake.Common.Tests/Unit/IO/FileCopierTests.cs
+++ b/src/Cake.Common.Tests/Unit/IO/FileCopierTests.cs
@@ -147,6 +147,23 @@ namespace Cake.Common.Tests.Unit.IO
             }
 
             [Fact]
+            public void Should_Preserve_Subdirectory_Structure_When_Only_Subdirectory_Files_Match_Glob()
+            {
+                const string dstPath = "./dst";
+
+                // Given
+                var fixture = new FileCopierFixture();
+                fixture.EnsureFileExists("./src/sub/file3.dat");
+                fixture.EnsureDirectoryExists(dstPath);
+
+                // When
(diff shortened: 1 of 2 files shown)


devlead:

```
```

devlead:

PR #4806, opened 2026-05-17
feat: expose stdout and stderr when a tool process fails (#2460)

### LINKED ISSUE #2460: Tool.Run should make stdout and stderr available on failure
Tool.Run and related should make stdout and stderr available so that some analysis is possible on failure. Currently it calls [ProcessExitCode](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Tooling/Tool.cs#L119) which [throws](https://github.com/cake-build/cake/blob/develop/src/Cake.Core/Tooling/Tool.cs#L139) a simple `CakeException`.

Ideally, it would capture stdout and stderr and throw a `CakeProcessException`, or similar, that has additional properties for exit code, stdout and stderr.

wyf027:
## Summary

- Add `CakeProcessException` (extends `CakeException`) with `StandardOutput` and `StandardError` line collections.
- `Tool.Run` now calls `ProcessExitCode(IProcess)` and throws `CakeProcessException` on non-zero exit codes when output redirection captured lines.
- NUnit runners keep custom exit-code mapping via the existing `ProcessExitCode(int)` override.

## Test plan

- [x] `Should_Throw_On_NonZero_ExitCode_Without_Custom_Validation` — expects `CakeProcessException`
- [x] `Should_Include_Standard_Output_And_Error_On_NonZero_ExitCode`
- [ ] Full Cake.Core / Cake.Common test suites (CI)

CI not run locally (no .NET SDK in contributor environment).

Fixes #2460

### DIFF
--- a/src/Cake.Common/Tools/NUnit/NUnit3Runner.cs
+++ b/src/Cake.Common/Tools/NUnit/NUnit3Runner.cs
@@ -248,6 +248,12 @@ namespace Cake.Common.Tools.NUnit
             return new[] { "nunit3-console.exe" };
         }
 
+        /// <inheritdoc />
+        protected override void ProcessExitCode(IProcess process)
+        {
+            ProcessExitCode(process.GetExitCode());
+        }
+
         /// <summary>
         /// Customized NUnit 3 exit code handling.
         /// Throws <see cref="CakeException"/> on non-zero exit code.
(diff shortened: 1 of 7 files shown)


devlead:

```
```

devlead:

PR #4836, opened 2026-05-20
Add RWX build provider

### LINKED ISSUE #4840: Add RWX as a build provider
### Discussed in https://github.com/orgs/cake-build/discussions/4796

<div type='discussions-op-text'>

<sup>Originally posted by **robinaugh** April 30, 2026</sup>
Hey team - Per the contributing guide, I'm seeking buyoff before opening a PR.

[RWX](https://www.rwx.com/) is a cloud CI/CD platform with a growing .NET user base. I'd like to add it as a build provider under src/Cake.Common/Build/Rwx/, alongside the 14 already supported (GitHub Actions, GitLab CI, Azure Pipelines, etc.) so that Cake scripts running on RWX get the same `IsLocalBuild` / `IsRunningOnRwx` / `BuildSystem.Rwx.Environment` behavior they get on every other supported CI.

## Detection contract

RWX exposes a documented env var contract: https://www.rwx.com/docs/environment-variables#meta-environment-variables

## Proposed shape

Modeled on the existing TravisCI / GitLabCI providers — env/info only, no Commands surface for now:

```
src/Cake.Common/Build/Rwx/
  IRwxProvider.cs
  RwxProvider.cs
  RwxInfo.cs
  Data/RwxEnvironmentInfo.cs
```

Plus the usual additions to BuildProvider.cs, BuildSystem.cs, BuildSystemAliases.cs, DI registration, and unit tests modeled on the existing provider tests.

Worth flagging up front:
- **`IsPullRequest` aggregation.** RWX's current envvar contract doesn't expose a PR identifier (many runs are kicked off from the CLI before PRing), so the initial provider would not contribute to `BuildSystem.IsPullRequest`.
- **Casing**. Existing folders are inconsistent (GoCD, GitLabCI vs. MyGet). I'd go Rwx to match .NET conventions, but happy to defer.

If approved, I'll open the accompanying issue linking back to this discussion and start the PR against develop. Once that's in, we'd also document the integration in our docs at rwx.com/docs.

I appreciate the time, and I'm happy to adjust scope or shape based on what you'd prefer to see!</div>

robinaugh:
Following https://github.com/orgs/cake-build/discussions/4796, this PR adds [RWX](https://www.rwx.com/) as a build provider.

I have already been in contact with @devlead via email and we've set up a Cake organization on RWX. A sample run can be seen here: https://cloud.rwx.com/mint/cake/runs/a8eb7be0436248508544a2c0fda9df08
- The sample run requires a login (Mattias and the Cake team have access), but runs triggered from this public repository (on `main`/`develop`/`hotfix*` branches) will be publicly accessible and linked from the badge in the README. 
- More documentation on how RWX permissions work with Open Source can be found [here](https://www.rwx.com/docs/open-source).


robinaugh:
@microsoft-github-policy-service agree company="RWX"

.rwx/cake.yml
```diff
@@ -0,0 +1,48 @@
...
+    push:
+      - if: ${{ event.git.branch == 'develop' || event.git.branch == 'main' || starts-with(event.git.branch, 'hotfix/') }}
+        init:
+          commit-sha: ${{ event.git.sha }}
+  cli:
+    init:
+      commit-sha: ${{ event.git.sha }}
+
+base:
+  image: ubuntu:24.04
```
devlead: does this have a latest or will yaml need to be updated every time a new image is releases / retired?

.rwx/cake.yml
```diff
@@ -0,0 +1,48 @@
...
+      git branch -f develop HEAD
+      git symbolic-ref HEAD refs/heads/develop
+
+      # Pre-install the latest GA SDK in the 10.0 channel (currently 10.0.3xx) so global.json's
+      # rollForward: latestFeature has something to roll forward to. build.sh otherwise installs
+      # exactly 10.0.201, whose NuGet client escalates NU1901 against NuGet.Packaging 7.3.0;
+      # 10.0.3xx does not.
+      mkdir -p .dotnet
+      curl -Lsfo .dotnet/dotnet-install.sh https://dot.net/v1/dotnet-install.sh
+      bash .dotnet/dotnet-install.sh --channel 10.0 --install-dir .dotnet --no-path
```
devlead: global.json in repo is now a `10.0.300`, so is this needed?
https://github.com/cake-build/cake/blob/4ad5f9bd018e23aea5a7f7c3a7f3b309e21931b3/global.json#L6

src/Cake.Common/Build/Rwx/Data/RwxEnvironmentInfo.cs
```diff
@@ -0,0 +1,188 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Cake.Core;
+
+namespace Cake.Common.Build.Rwx.Data
+{
+    /// <summary>
+    /// Provides RWX environment information for the current build.
+    /// </summary>
+    public sealed class RwxEnvironmentInfo : RwxInfo
```
devlead, line 12: I'm new to RWX so maybe stupid question but is there any VCS like branch, repo, commit ref/tag environment variables available?

Had a quick look at https://www.rwx.com/docs/environment-variables#meta-environment-variables
but didn't seem like it, but just wanted to double check.

Is there like GitHub actions / Azure Pipelines any per run runtime token that can be uses with cli i.e. something like RWX_ACCESS_TOKEN?

Is there any runtime api / files / folder conventions to do any actions like set environment / output variable  or upload artifacts?

devlead (commented):
In general LGTM 👍 , just a few questions.

.rwx/cake.yml
robinaugh: We can use `ubuntu:latest` here - we generally recommend using a named version in order to get a guaranteed cache miss when it updates, but it's up to you - I can also just bump this to `ubuntu:26.04` to extend its shelf life (using 24.04 here was old muscle memory on my part, so I'll definitely do one or the other based on your preference).

.rwx/cake.yml
robinaugh: Just tested and I think we're good without it. Per the contribution guidelines, it looks like you're OK with me rebasing on `develop` to address this feedback - do you prefer the feedback changes in a new commit, or OK to squash into the existing commits?

.rwx/cake.yml
devlead: `ubuntu:26.04` sounds reasonable.

.rwx/cake.yml
devlead: Please rebase into one commit

src/Cake.Common/Build/Rwx/Data/RwxEnvironmentInfo.cs
robinaugh: #### VCS

One thing that's different between RWX and other providers is that a lot of folks kick off runs via CLI _before_ committing/pushing (the CLI sends up a git patch through our API), so there may not be VCS metadata available. 

However, we just pushed an update to our docs detailing how to get those values when using the `git/clone` package (which, realistically, users will have). That's here:

https://www.rwx.com/docs/environment-variables#git-environment-variables

I can thread those through.

#### Access tokens

We do have an access token available as a secret, so users can make it available as an envvar like this:

```yml
tasks:
  - key: cake
    run: dotnet cake
    env:
      RWX_ACCESS_TOKEN: ${{ secrets.RWX_ACCESS_TOKEN }}
```

However, that is not _guaranteed_ to be set, and can be wired up using any arbitrary envvar name. Is that something that should be documented somewhere? FWIW I also intend to add a Cake entry to https://www.rwx.com/docs/works-with once this ships, so if there are nuances that don't make sense to live in this repo's docs, it could link to that page.

#### Output values and artifacts

We do support setting [output values](https://www.rwx.com/docs/output-values) and [artifacts](https://www.rwx.com/docs/artifacts) - both can be done like this:

```
echo "value" > $RWX_VALUES/key
echo "file content" > $RWX_ARTIFACTS/output.txt
```

I think it should be straightforward to do those in commands (e.g. `BuildSystem.Rwx.Commands.SetValue` etc). Is that something you'd want to see supported in this same PR?

devlead: > https://www.rwx.com/docs/environment-variables#git-environment-variables
>
>I can thread those through.

Sounds good 👍 


> Access tokens

If not guaranteed  / standard, agree can just be passed as an env var at users discresion.


> Output values and artifacts

That would be great, would be very similar to how variables are set with i.e. GitHubActions, except the file name is key and value content.

https://github.com/cake-build/cake/blob/4ad5f9bd018e23aea5a7f7c3a7f3b309e21931b3/src/Cake.Common/Build/GitHubActions/Commands/GitHubActionsCommands.cs#L169-L190

And `UploadArtifact(FilePath path)`

Could be something like
```cake
if (ArtifactsPath == null) 
{ 
    throw new CakeException("RWX Artifact path missing."); 
} 

var absoluteFilePath = path.MakeAbsolute(_environment);
if (!_fileSystem.Exist(absoluteFilePath))
{
    const string format = "The artifact file '{0}' does not exist.";
    var message = string.Format(CultureInfo.InvariantCulture, format, absoluteFilePath.FullPath);
    throw new FileNotFoundException(message, absoluteFilePath.FullPath);
}

var absoluteTargetPath = targetFilePath.MakeAbsolute(context.Environment);
var file = _fileSystem.GetFile(absoluteFilePath);

var newAbsoluteTargetPath = ArtifactsPath.CombineWithFilePath(path.GetFilename());

file.Copy(newAbsoluteTargetPath, true);
```

tests/integration/Cake.Common/Build/Rwx/RwxProvider.cake
```diff
@@ -0,0 +1,65 @@
...
+
+        Information("  Runtime:");
+        Information("    ValuesPath: {0}", BuildSystem.Rwx.Environment.Runtime.ValuesPath);
+        Information("    ArtifactsPath: {0}", BuildSystem.Rwx.Environment.Runtime.ArtifactsPath);
+        Information("    IsRuntimeAvailable: {0}", BuildSystem.Rwx.Environment.Runtime.IsRuntimeAvailable);
+});
+
+Task("Cake.Common.Build.RwxProvider.Commands")
+    .Does(() => {
+        BuildSystem.Rwx.Commands.SetValue("cake_integration_test", "ok");
```
devlead, line 50: UploadArtifact works great

Image


But how do we validate SetValue? Could we have an extra task in `.rwx/cake.yml` that echoes / asserts `cake_integration_test` ?

src/Cake.Common/Build/Rwx/Commands/RwxCommands.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        /// <summary>
+        /// Sets an RWX output value for the current task. The value is written to
+        /// <c>{RWX_VALUES}/{key}</c>; each key is a separate file, so repeated
+        /// calls with the same key overwrite the previous value. Multiline values
+        /// are written verbatim.
+        /// </summary>
+        /// <param name="key">The value key. May not contain path separators or <c>..</c>.</param>
+        /// <param name="value">The value contents.</param>
```
devlead, line 50: Looking at docs, we could have similar methods for env vars
https://www.rwx.com/docs/use#environment-variables

```yaml
tasks:
  - key: set-env-one
    run: echo hello >> $RWX_ENV/ONE

  - key: set-env-two
    run: echo world >> $RWX_ENV/TWO

  - key: print-env
    use: [set-env-one, set-env-two]
    run: |
      echo $ONE $TWO
      env | grep -E 'ONE|TWO'
```

devlead (changes requested):

tests/integration/Cake.Common/Build/Rwx/RwxProvider.cake
robinaugh: Yep, I'm pushing a change to add tasks to assert the values of the artifact, output value, and envvar set by the integration test.

src/Cake.Common/Build/Rwx/Commands/RwxCommands.cs
robinaugh: Good call, I'm adding this too.

devlead (approved):
LGTM 👍

devlead:
@robinaugh your changes have been merged. Thanks for your contribution and for addressing all feedback 👍

PR #4842, opened 2026-05-25
Add FindFilesInDirectory alias to FileAliases

### LINKED ISSUE #2056: Make easier to search files in a certain directory
I often find myself in the situation of looking for files matching a certain pattern within a folder or below.

Here is a typical use case

As of today, executing `dotnet test` on a solution will try to find and execute tests on every project filling the log with errors because no testable project was found. This is why I prefer to find every project file within the `tests` directory hive.

```csharp
Task("Test")
    .IsDependentOn("Build")
    .Does(() => 
{
    var testProjects = FindFilesInDirectory(testDirectory, "*.csproj", SearchScope.Recursive);

    foreach (var project in testProjects)
    {
        DotNetCoreTest(project.FullPath);
    }
});
```

I am aware that the same functionality is offered by `IFileSystem` but the result of `IFileSystem.GetFiles` is a list of `IFile` instead of `FilePath`. This is not convenient since most of the aliases work with `FilePath`.

codenano077:
## Summary
Adds a new `FindFilesInDirectory` alias to `FileAliases` that allows users 
to easily search for files matching a pattern within a directory.

## Changes
- Added `FindFilesInDirectory` method to `FileAliases.cs`
- Added unit tests for the new method

## Related Issue
Closes #2056

## Example Usage
```csharp
var testProjects = FindFilesInDirectory(
    "./tests", 
    "*.csproj", 
    SearchScope.Recursive);

foreach (var project in testProjects)
{
    DotNetTest(project.FullPath);
}
```

### DIFF
--- a/src/Cake.Common.Tests/Unit/IO/FileAliasesTests.cs
+++ b/src/Cake.Common.Tests/Unit/IO/FileAliasesTests.cs
@@ -1279,5 +1279,75 @@ namespace Cake.Common.Tests.Unit.IO
                 Assert.Equal("/bar/baz.qux", result.FullPath);
             }
         }
+
+        public sealed class TheFindFilesInDirectory
+        {
+            [Fact]
+            public void Should_Throw_If_Context_Is_Null()
+            {
+                // Given
+                var directory = new DirectoryPath("./tests");
+
+                // when
+                var result = Record.Exception(() =>
+                    FileAliases.FindFilesInDirectory(null, directory, "*.csproj"));
+
+                // Then
+                AssertEx.IsArgumentNullException(result, "context");
+            }
+
+            [Fact]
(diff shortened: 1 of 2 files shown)


devlead:
Thanks for taking the time to contribute, and for adding tests along with the alias.

Cake already covers this through globbing. `GetFiles` returns `FilePath` results, so the example in this PR can be written as:

```csharp
var testProjects = GetFiles("./tests/**/*.csproj");
```

or, with a directory path:

```csharp
var testProjects = GetFiles($"{testDirectory}/**/*.csproj");
```

Because of that, a separate `FindFilesInDirectory` alias wrapping `IDirectory.GetFiles` would add a second, less powerful search API without a clear benefit over the existing globber. I'm going to close this PR, but we appreciate the effort.

PR #4986, opened 2026-09-12
GH4977: Fix CS8632 for aliases from nullable enabled addins

### LINKED ISSUE #4977: CS8632 when Cake.Tool generates aliases from nullable-enabled addins
CS8632 when Cake.Tool generates aliases from nullable-enabled addins

Cake.Tool emits CS8632 for addins compiled with NRT (Cake.Docker 1.5.0-beta.1, current Cake.FileHelpers). `#nullable disable` in the `.cake` file does nothing: aliases are generated *above* user script in a compilation with annotations disabled.

```cake
#addin nuget:?package=Cake.Docker&version=1.5.0-beta.1&prerelease
```

Do not suppress CS8632 globally. Wrap only nullable-aware generated aliases in `#nullable enable` / `#nullable restore`, and emit `?` only for `NullableAttribute` flag `2` (including inner generics/arrays).

https://github.com/orgs/cake-build/discussions/4719
https://github.com/cake-build/cake/issues/4197

### LINKED ISSUE #4985: CakeExecuteScript/CakeExecuteExpression silently ignore several CakeSettings when running via Cake.Tool
`CakeRunner.ExecuteScript` has two execution paths, and they don't honour the same set of
`CakeSettings` (i.e. `ToolSettings`) members:

* when the resolved tool path is **not** `Cake.dll` (`cake`, `Cake.exe`, homebrew), it calls
  `Tool<CakeSettings>.Run(settings, args)`, which honours every `ToolSettings` member;
* when the resolved tool path **is** `Cake.dll` — which is the default whenever the outer build
  runs on `Cake.Tool`, since `GetAlternativeToolPaths` points at the executing assembly's
  `Cake.dll` — it hand-copies a subset of the settings onto a `DotNetExecuteSettings` and runs
  them through `DotNetExecutor`.

The hand-copied subset is missing four members, so they are silently dropped:

* `HandleExitCode`
* `NoWorkingDirectory`
* `PostAction`
* `SetupProcessSettings`

`src/Cake.Common/Tools/Cake/CakeRunner.cs`:

```csharp
_coreExecutor.Execute(
    toolPath,
    GetArguments(scriptPath, settings),
    new DotNetExecuteSettings
    {
        ArgumentCustomization = settings.ArgumentCustomization,
        EnvironmentVariables = settings.EnvironmentVariables,
        ToolTimeout = settings.ToolTimeout,
        WorkingDirectory = settings.WorkingDirectory
    });
```

`CakeExecuteExpression` delegates to `ExecuteScript`, so it is affected identically.

## Repro

```csharp
var output = new List<string>();

CakeExecuteExpression(
    "Information(\"hello from the nested script\");",
    new CakeSettings
    {
        HandleExitCode = exitCode => true,
        SetupProcessSettings = process =>
        {
            process.RedirectStandardOutput = true;
            process.RedirectedStandardOutputHandler = line => { output.Add(line); return line; };
        }
    });

Information("Captured {0} lines", output.Count);
```

**Expected:** the nested script's stdout is redirected and captured, and a non-zero exit code is
handled by the delegate.

**Actual:** when the outer build runs on `Cake.Tool`, `output` is empty (the nested output goes
straight to the console) and a non-zero exit code throws, because neither delegate reaches the
process. Running the same script with `ToolPath` pointing at `cake`/`Cake.exe` behaves as
documented, which makes this look like an environment-dependent bug rather than a missing feature.

This blocks any scenario that needs to inspect a nested Cake run — capturing its output, asserting
on its warnings, or tolerating a deliberate non-zero exit — which is exactly what integration tests
around nested script execution need.

devlead:
Cake.Tool emits CS8632 for addins compiled with nullable reference types enabled, and
`#nullable disable` in the `.cake` file doesn't help, because aliases are generated above the user
script in a compilation where annotations are disabled.

### Nullable aware alias generation (#4977)

* Only aliases that actually carry a nullable annotation are wrapped in `#nullable enable` /
  `#nullable restore`. CS8632 is not suppressed anywhere, and scripts keep the oblivious context
  they have today.
* `?` is emitted from `NullableAttribute` / `NullableContextAttribute` flag `2`, including inner
  generic arguments and array elements, so `IList<string?>`, `string?[]?` and `Task<string?>` all
  round trip.
* Unconstrained type parameters stay unannotated. `NullabilityInfoContext` reports every usage of
  such a parameter as nullable, so it can't tell `T` from `T?`; for those positions the nullable
  flag of the parameter or return position decides instead.
* Generic parameter constraints keep their nullability: `class?` is emitted as `class?`, and
  `notnull` is emitted whenever no other constraint implies it, including alongside `new()`.

### Dropped CakeSettings on the Cake.dll path (#4985)

`CakeRunner.ExecuteScript` hand-copies settings onto `DotNetExecuteSettings` when the resolved tool
path is `Cake.dll`, which is the default whenever the outer build runs on Cake.Tool.
`HandleExitCode`, `NoWorkingDirectory`, `PostAction` and `SetupProcessSettings` were missing from
that copy and silently dropped. They're now propagated, which is also what lets the new integration
test capture the nested script's compiler output.

- fixes #4977
- fixes #4985

### DIFF
--- a/src/Cake.Common.Tests/Fixtures/Tools/CakeRunnerCoreFixture.cs
+++ b/src/Cake.Common.Tests/Fixtures/Tools/CakeRunnerCoreFixture.cs
@@ -1,0 +1,36 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Cake.Common.Tools.Cake;
+using Cake.Core.IO;
+using Cake.Testing;
+using Cake.Testing.Fixtures;
+
+namespace Cake.Common.Tests.Fixtures.Tools
+{
+    /// <summary>
+    /// Runs the Cake tool through the .NET host, which is what happens when the
+    /// resolved tool path is Cake.dll instead of a Cake executable.
+    /// </summary>
+    internal sealed class CakeRunnerCoreFixture : ToolFixture<CakeSettings>
+    {
+        public FilePath ScriptPath { get; set; }
+
+        public CakeRunnerCoreFixture()
+            : base("dotnet.exe")
+        {
+            ScriptPath = new FilePath("./build.cake");
+            FileSystem.CreateFile(ScriptPath.MakeAbsolute(Environment));
+
+            Settings.ToolPath = new FilePath("./tools/Cake.dll");
+            this.GivenSettingsToolPathExist();
+        }
+
+        protected override void RunTool()
+        {
+            var runner = new CakeRunner(FileSystem, Environment, Globber, ProcessRunner, Tools);
+            runner.ExecuteScript(ScriptPath, Settings);
+        }
+    }
+}
--- a/src/Cake.Common.Tests/Unit/Tools/Cake/CakeRunnerTests.cs
+++ b/src/Cake.Common.Tests/Unit/Tools/Cake/CakeRunnerTests.cs
@@ -139,5 +139,105 @@ namespace Cake.Common.Tests.Unit.Tools.Cake
                              "--configuration=\"Debug\"", result.Args);
             }
         }
+
+        public sealed class TheExecuteScriptMethodViaTheDotNetHost
+        {
+            [Fact]
+            public void Should_Execute_Cake_Dll_Through_The_DotNet_Host()
+            {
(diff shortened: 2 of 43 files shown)


patriksvensson (approved):
LGTM!

PR #5010, opened 2026-09-18
GH2835: Replace Autofac with Microsoft.Extensions.DependencyInjection.

### LINKED ISSUE #2835: Migrate Cake.Tool from Autofac to Microsoft.Extensions.DependencyInjection
We're currently not using any advanced features of Autofac (to my knowledge), so we should investigate whether or not we can move to something more lightweight (perhaps the homebrewed DI container we use in `Cake.Frosting`?). I suspect we could reduce the load time for implicit bootstrapping by doing this.

Any investigation of this needs some kind of benchmarking to make sure we're not making it worse (or replace something without an improvement).

devlead:
- Replace Autofac with `IServiceCollection` / `IServiceProvider` for Cake.Tool CLI composition, bootstrap/build scopes, and module loading.
- Add `ContainerRegistrar` in Cake.Cli so Cake modules map onto MS.DI, including one singleton instance registered as multiple service types.
- Share `AddCakeDiagnostics` and `UseCakeDefaultModules` from Cake.Cli for both `IServiceCollection` and `ICakeContainerRegistrar`, so Cake.Tool and Cake.Frosting compose diagnostics and Core/Common/NuGet/DotNetTool the same way.
- Reuse that registrar from Cake.Frosting, drop the Frosting-only Autofac-style adapter, and keep Cake.Tool able to swap the Spectre report printer for the plain printer.
- Expose the execution container on `ICakeContext.ServiceProvider` and the `ServiceProvider` script alias so built-in and `#module` services can be resolved from Cake.Tool scripts.
- Reference Microsoft.Extensions.DependencyInjection from script conventions, and load `Cake.*.Module.dll` assemblies into the script so module types resolve after `#module`.
- fixes #2835

## Test plan
- [ ] Run unit tests for `ContainerRegistrar`, `CakeContext.ServiceProvider`, and `ServiceProvider` aliases.
- [ ] Run Cake.Common/Cake.Core integration tests for `ServiceProvider` resolution and `#module` packing of a custom module.
- [ ] Confirm a Cake.Frosting host still builds and that `IFrostingContext` is the frosting wrapper while `ICakeContext` remains the core context from `CoreModule`.
- [ ] Confirm Cake.Tool still uses the Spectre report printer by default, and the plain printer when Spectre console output is disabled.


patriksvensson (approved):
LGTM 👍

PR #5011, opened 2026-09-19
(GH-3829) Defer Windows registry key creation until used.

### LINKED ISSUE #3829: Error: Error getting value from 'CurrentUser' on 'Cake.Core.IO.WindowsRegistry'.
### Prerequisites

### Cake runner

Cake Frosting

### Cake version

2.1.0.0

### Operating system

Linux

N/A

### CI Server

_No response_

### What are you seeing?

I am logging the Build Context in the Setup, it works on the Windows and fails on the Ubuntu.
```
context.Information(context.ToJson(Formatting.Indented))
```

### What is expected?

Please find the expected Output from Windows, I should get similar in Ubuntu
```
----------------------------------------
Setup
----------------------------------------
Cake Context Start
{
  "SharedSettings": {},
  "FileSystem": {},
  "Environment": {
    "WorkingDirectory": "D:/Repos/FleetHub/Build/cake-frosting/CakeBuild/bin/Debug/net5.0",
    "ApplicationRoot": "D:/Repos/FleetHub/Build/cake-frosting/CakeBuild/bin/Debug/net5.0",
    "Platform": {
      "Family": "windows",
      "Is64Bit": true
    },
    "Runtime": {
      "BuiltFramework": {
        "Identifier": ".NETCoreApp",
        "Version": "5.0",
        "Profile": "",
        "FullName": ".NETCoreApp,Version=v5.0"
      },
      "Runtime": "coreClr",
      "CakeVersion": "2.1.0.0",
      "IsCoreClr": true
    }
  },
  "Globber": {},
  "Log": {
    "Verbosity": "normal"
  },
  "Arguments": {},
  "ProcessRunner": {},
  "Registry": {
    "CurrentUser": {},
    "LocalMachine": {},
    "ClassesRoot": {},
    "Users": {},
    "PerformanceData": {},
    "CurrentConfig": {}
  },
  "Tools": {},
  "Data": {},
  "Configuration": {}
}
```

### Steps to Reproduce

Log the Build Context in the Setup
```
context.Information(context.ToJson(Formatting.Indented))
```

### Output log

_No response_

Samyra312007:
Fixes #3829
 
## Pre-submit checklist
 
- [x] Submitted from a feature branch (`feature/gh-3829`), not `master`
- [x] No reformatting — only the 3 files specific to this fix are touched
- [x] New code has unit test coverage
- [x] All tests passing; compiles without errors or warnings
 
## What are you seeing today? / What does this fix?
 
Serializing the build context on Linux/macOS fails with:
 

Error: Error getting value from 'CurrentUser' on 'Cake.Core.IO.WindowsRegistry'.

 
Repro (from the issue, Cake Frosting 2.1.0.0 on Ubuntu):
 
```csharp
context.Information(context.ToJson(Formatting.Indented));

**Root cause:**

 WindowsRegistry  eagerly materialized the underlying  Microsoft.Win32.Registry 
base keys in its property getters:

// csharp
public IRegistryKey CurrentUser => new WindowsRegistryKey(Microsoft.Win32.Registry.CurrentUser);

On non-Windows platforms, dereferencing  Microsoft.Win32.Registry.*  throws
 PlatformNotSupportedException  (or returns  null  on newer .NET versions), so
any reflective consumer of  ICakeContext e.g. a JSON serializer walking the
context throws on the first registry property, even when the script never
touches the registry.

**What this PR changes:**

Defers creation of the underlying  RegistryKey  until an actual registry
operation is performed:
-  src/Cake.Core/IO/WindowsRegistry.cs  each  IRegistry  property now
returns a key wrapping a lazy  Func<RegistryKey>  factory instead of
dereferencing the hive immediately.
-  src/Cake.Core/IO/WindowsRegistryKey.cs  accepts a  Func<RegistryKey> 
factory; the key is created exactly once (double-checked locking) on first
 GetSubKeyNames()  /  OpenKey()  /  GetValue() .  Dispose()  never forces
creation. A  null  hive (newer .NET on unsupported platforms) is normalized
to  PlatformNotSupportedException  so real registry use fails predictably.
-  src/Cake.Core.Tests/Unit/IO/WindowsRegistryTests.cs, new tests.

**Why this approach:**

- No public API changes IRegistry  /  IRegistryKey ,  CoreModule  DI
registration, and existing consumers ( SignToolResolver ,  InnoSetupRunner )
are untouched.
- Behavior preserved actual registry usage on Linux/macOS still throws
 PlatformNotSupportedException , just at the point of use rather than at
property access.
- Reported scenario now works serializing the context on Ubuntu produces
the same shape as Windows:  "Registry": { "CurrentUser": {}, ... } .

**Test evidence:**

New tests (17) in  src/Cake.Core.Tests/Unit/IO/WindowsRegistryTests.cs :

┌──────────────┬─────┬─────────────────────────────────────────────────────────────────────────────────────────┐
│ Test group   │ Cou │ Verifies                                                                                │
│              │ nt  │                                                                                         │
├──────────────┼─────┼─────────────────────────────────────────────────────────────────────────────────────────┤
│ Each of the  │ 12  │ Property access does not throw on any platform; returns a non-null key                  │
│ 6 IRegistry  │     │                                                                                         │
│ properties   │     │                                                                                         │
│ DeferredKeyC │ 5   │ Factory not invoked on Dispose(); invoked exactly once across repeated use; factory     │
│ reation      │     │ exceptions propagate; PlatformNotSupportedException on actual use on non-Windows        │
│              │     │ (NonWindowsFact); real sub-key enumeration on Windows (WindowsFact)                     │
└──────────────┴─────┴─────────────────────────────────────────────────────────────────────────────────────────┘

Full suite -  Cake.Core.Tests , net10.0, Linux:

Total: 1850
Passed: 1774 
Skipped (platform-restricted): 76
Failed: 0      

Build: 0 warnings, 0 errors.

### CODE REVIEW

File: src/Cake.Core/IO/WindowsRegistryKey.cs [hunk 1]
```diff
@@ -1,46 +1,82 @@
...
-// See the LICENSE file in the project root for more information.
 
+using System;
 using Microsoft.Win32;
 
 namespace Cake.Core.IO
 {
-    internal sealed class WindowsRegistryKey : IRegistryKey
-    {
+internal sealed class WindowsRegistryKey : IRegistryKey
```
devlead, line 9: Please restore the usual 4-space indent for the class (and keep members indented one more level), and put back the third license header line used elsewhere in Cake.Core:

`
// See the LICENSE file in the project root for more information.
`

`EnforceCodeStyleInBuild` + Release `TreatWarningsAsErrors` will treat the current formatting as IDE0055.

src/Cake.Core/IO/WindowsRegistryKey.cs
```diff
@@ -1,46 +1,82 @@
...
+                        }
+                    }
+                }
+
+                // Depending on the .NET version, the static Microsoft.Win32.Registry hive
+                // properties either throw PlatformNotSupportedException on their own or return
+                // null on platforms without registry support. Normalize the latter to the same
+                // exception so that actual registry use always fails predictably.
+                return _key ?? throw new PlatformNotSupportedException(RegistryNotSupportedMessage);
+            }
```
devlead, line 52: Normalizing a null hive to `PlatformNotSupportedException` is the right behavior across .NET versions.

Optional simplification: `Lazy<RegistryKey>` with `LazyThreadSafetyMode.ExecutionAndPublication` would cover once-only creation, null caching, and factory exceptions without a custom `_keyCreated` flag. `Dispose()` should still not force creation (only dispose `lazy.IsValueCreated` / the cached key).

src/Cake.Core.Tests/Unit/IO/WindowsRegistryTests.cs
```diff
@@ -0,0 +1,271 @@
...
+using Xunit;
+
+namespace Cake.Core.Tests.Unit.IO
+{
+    public sealed class WindowsRegistryTests
+    {
+        public sealed class TheCurrentUserProperty
+        {
+            [Fact]
+            public void Should_Not_Throw_When_Accessed()
```
devlead, line 16: These per-hive tests are useful, but they only read a single property. #3829 is a reflective walk of `ICakeContext.Registry` (Newtonsoft `ToJson`). Cake.Core.Tests does not reference Newtonsoft, so a test that loops `typeof(IRegistry).GetProperties()`, asserts each getter does not throw, and asserts a non-null `IRegistryKey` would cover the actual failure mode without adding a package.

Worth adding the same idea for `OpenKey` / `GetValue` sharing the deferred `Key` path, not only `GetSubKeyNames`.

devlead (changes requested):
Thanks for taking this on — the approach is the right fix for #3829.

`WindowsRegistry` currently materializes `Microsoft.Win32.Registry.*` in the hive getters, so any reflective walk of `ICakeContext` (including `context.ToJson(...)` on Linux/macOS) throws `PlatformNotSupportedException` wrapped as `Error getting value from 'CurrentUser' on 'Cake.Core.IO.WindowsRegistry'`, even when the script never uses the registry.

Deferring native `RegistryKey` creation until `GetSubKeyNames` / `OpenKey` / `GetValue` keeps `IRegistry` / `IRegistryKey` and the `CoreModule` registration unchanged, leaves SignTool/InnoSetup on the same `OpenKey` path, and still fails at the point of actual registry use on non-Windows. Serializer output of empty hive objects (`"CurrentUser": {}`) matches the Windows shape from the issue because `IRegistryKey` only has methods, not properties.

A few things to address before merge:

1. **Formatting in `WindowsRegistryKey`** — the class is flush-left and the standard third license line was dropped. Cake builds with `EnforceCodeStyleInBuild` and Release `TreatWarningsAsErrors`, so this is likely to fail CI as IDE0055.
2. **Regression test for the reported scenario** — the new tests cover property access and deferred use, but none walk all `IRegistry` properties the way a JSON serializer does. A single reflection test over `typeof(IRegistry).GetProperties()` (Cake.Core.Tests has no Newtonsoft reference) would lock in #3829.
3. **CI on this PR is red** across AppVeyor / Azure / Bitrise / TeamCity while `develop` is green. Please get a green run (or confirm it is fork-infra only) before we merge.

Nit: `Lazy<RegistryKey>` (`ExecutionAndPublication`) would replace the hand-rolled double-checked locking; Cake already uses `Lazy<T>` elsewhere. Not a blocker if you prefer to keep the explicit factory.

Happy to re-review once the formatting and test are in.

Samyra312007:
@devlead I would like to confirm are these CI failures are related to my PR or are these the existing failures.

devlead:
> @devlead I would like to confirm are these CI failures are related to my PR or are these the existing failures.

All CI on target branch is green and confirmed by checking out this PR and executing
```bash
dotnet tool restore 
dotnet cake
```
*(`build.sh` / `build.ps1` in repo root will do similar)*
That errors occur locally too, so it's issues introduced in this PR.
```diff
- src\Cake.Core\IO\WindowsRegistryKey.cs(9,1): error IDE0055: Fix formatting (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0055)
- src\Cake.Core\IO\WindowsRegistryKey.cs(10,1): error IDE0055: Fix formatting (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0055)
```

More guidance on contributing can be read in our [Contribution Guidelines](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md)

devlead (approved):
LGTM 👍

devlead:
@Samyra312007 your changes have been merged, thanks for your contribution 👍

